PR #50, opened 2016-10-07
Add dockerization for quick deploying and testing

strazzere:
I'm not extremely familiar with vagrant, though I have worked with Docker a bit. Unsure if there is a shared interest in having this easily set up by as a Docker instance or not, however I quickly put this together so I can get an instance locally and remotely running for testing purposes.

I can also continue to maintain this with deploying instructions as well.


docker/nginx/config/project.conf
```diff
@@ -0,0 +1,38 @@
...
+        #
+        listen 443 ssl http2;
+        listen [::]:443 ssl http2;
+	# If there is a letsencrypt cert, let us use this instead
+	# include snippets/ssl-params.conf;
+	# Otherwise, use hardcoded ssl cert for debugging purposes
+	# Can easily generate new ones following;
+	# https://www.digitalocean.com/community/tutorials/how-to-create-a-ssl-certificate-on-nginx-for-ubuntu-12-04
+	ssl on;
+	ssl_certificate /etc/nginx/conf.d/server.crt;
```
freddymartinez9: is this path correct? It sticks out to me as not being wrong. `/etc/nginx/conf.d/server.crt` but I see this file path as `docker/nginx/config/server.crt`. unless this file `docker/nginx/Dockerfile` does the remapping of relative paths.

freddymartinez9:
So @redshiftzero did a lot of refactoring to the PG database and now we have test data. I don't know the best way forward with merging the test data. Overall this commit looks good, just a question I asked inline and we need to resolve the conflict with app/forms.py as well

docker/nginx/config/project.conf
strazzere: The file `docker/nginx/Dockerfile` does the mapping, specifically this line;
`ADD config/ /etc/nginx/conf.d/ `

strazzere:
I believe I should have some time later today to rebase - I'd like to get the test data added into the PR prior to a merge.

redshiftzero:
Hey @strazzere, if you do further work on this, please feel free to reopen! Until then I'm going to close for now - we want to be at PR 0 for the holidays 😃 🎁 🎄 .

PR #82, opened 2016-10-17
remove ampersand which throws errors in systemd

### LINKED ISSUE #75: Fix systemd unit file in DEPLOY.md
Currently its failing to start but if I run `gunicorn app:app` manually it functions properly.

freddymartinez9:
fixes #75 but the line about env variables in `CONTRIB.md` is still what is borked on our deployment. @r4v5 and I can inspect the server but this PR should fix other deployments.

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -75,7 +75,8 @@ SECRET_KEY = 'changemeplzorelsehax'
 
 # Systemd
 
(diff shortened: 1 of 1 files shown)


r4v5:
My understanding is that it's failing to start because $PGPASS isn't defined in the systemd context, not because of the backgrounding at all. Is that incorrect?

(I have an idea of a fix for the env vars once I get a few actual free minutes)

freddymartinez9:
Yes, these are separate but related issues

PR #84, opened 2016-10-17
Add PGPASS to systemd environment variable

freddymartinez9:
Similar to #75, this fix will pick up the correct variables based on `CONTRIB.md`

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -75,7 +75,7 @@ SECRET_KEY = 'changemeplzorelsehax'
 
 # Systemd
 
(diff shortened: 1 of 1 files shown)


DEPLOY.md
```diff
@@ -87,6 +87,7 @@ User=nginx
 Group=nginx
 WorkingDirectory=/home/nginx/oovirtenv/OpenOversight/OpenOversight
 Environment="PATH=/home/nginx/oovirtenv/bin"
+Environment="PGPASS=~/.pgpass"
```
r4v5: good job Doing the Right Thing, systemd

PR #87, opened 2016-10-18
Switch to using python-dotenv for deployment-specific SECRETS and un-break the test env

### LINKED ISSUE #36: Dev Environment
If we want to develop with any amount of rigor, we should have a dev environment. This is probably (because I'm doing it, and I stick with the devils I know) going to be a vagrant box provisioned via puppet, with a helluva lot of rake magic. I'm also going to try to make this match the production schema with dummy values, but an actual migrations management system is also work we should do (in a separate issue).

r4v5:
Ideological discussions about 12-factor apps aside, we currently have
two deployment-specific secrets (DB URI and CSRF token) and we're
assuredly going to add more later. We need a way of managing those that
isn't "creating and managing a new file with secrets" or "stuffing them
in the global environment".

Enter dotenv. This python module lets you specify things in a file
called .env that then get added to the os.environ namespace. This is
handy because instead of deployment requiring hand-hacking of the
process environment, it just lives with the app. It's also a good
starting point if future users of OO don't want to janitor their own
servers: heroku, for instance, mostly works through .env files.

This change did some reworking to the app itself that we should test out
before sending to prod; it also changes the provisioning of the vagrant
VM to properly plop the right thing into .env. Now the app starts and
runs with sample data automatically in a vagrant. I also made the
create_db.py script idempotent-ish, so it doesn't do anything if it
receives a DatabaseAlreadyControlledError. This resolves #36 hopefully.

I've also updated the dev and deploy documentation to reflect our wonderful
New Reality.

I suggest we merge this _after_ tonight, and let people who aren't me play with it and see how it goes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,9 +6,6 @@
 # Credentials
 dbcred*
 
-# flask db migrations
-OpenOversight/db_repository/
-
 # Local environment files
 **/.env
 
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,15 @@
 # Credentials
 dbcred*
 
+# flask db migrations
+OpenOversight/db_repository/
+
+# Local environment files
+**/.env
+
+# Swapfiles from our text editors
+*.swp
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -8,44 +8,56 @@ When you come to implement your new feature, you should branch off `develop` and
 
 ## Development Environment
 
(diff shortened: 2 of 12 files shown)


freddymartinez9:
+1 will get more feedback from everyone else

r4v5:
Rebased this to resolve conflicts in DEPLOY.md and un-gitignored the migration repo. I'd love to get this in so folks can dev more easily.

redshiftzero:
Just been playing with this and testing it out: really, really awesome @r4v5. Thanks so much for doing this!!! Merging in...

Also, I really love our test data :trollface:

screen shot 2016-10-19 at 10 45 20 pm

PR #93, opened 2016-10-18
Notes about running in production

freddymartinez9:
Update with notes about how we had to fix prod deployment on the fly for release.


DEPLOY.md
```diff
@@ -47,6 +47,8 @@ server {
                 proxy_redirect     off;
                 proxy_set_header Host $host;
                 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+		proxy_connect_timeout 300s;
```
r4v5: hard tabs have revealed that the rest of this was annoyingly misindented. if you feel like fixing that, sure, if not, I'll hit the button.

PR #96, opened 2016-10-18
dev env tweaks

scuerda:
Minor tweaks. Vagrant can be crabby if running a dev server that isn't exposing 0.0.0.0. Also, updated the contrib doc to reflect changes to `test_data.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,6 @@ vagrant/puppet/modules/
 vagrant/puppet/.tmp
 .librarian
 .vagrant
+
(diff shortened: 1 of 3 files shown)


r4v5:
Interesting report on the binding. What were you seeing when not bound to 0.0.0.0? Were you accessing it at localhost:3000 or at the vm's ip?

scuerda:
The vagrantfile just has port forwarding, so I was just using localhost:3000. My browser was just reporting that the site couldn't be reached.

Issue #16, opened 2016-08-22
Pagination


redshiftzero:
There are a lot of officers in CPD - which can make for slow queries as the app tries to get a photo for thousands of officers. Let's have the app show 10-50 at a time.

scuerda:
Do you have feelings on if you want to hard code the pagination intervals or expose on front-end for user to chose?

redshiftzero:
I think it would be great to expose them to the user, but at first if we can get e.g. 8 officers shown per page that would be wonderful

redshiftzero:
I will work on this next when I have a free moment but will leave it unassigned until then so if someone else gets to it, awesome :metal:

scuerda:
@redshiftzero I'm working on pagination and I have a basic version working. However there is a conflict between pagination and sorting as it has been implemented. 

Since I'm only working w/ the test data I don't have a great handle on what the sorting is handling. Is it fair to assume that it is intended to push officers for which there is no tagged photo down to the bottom of the list? 

If so, there will need to be some changes in the primary officer query to allow for this alongside pagination since the current flow will only sort the subset of officers active within the current pagination item.

Any thoughts?

redshiftzero:
Done, thanks @scuerda!

PR #101, opened 2016-10-19
Pagination implementation

scuerda:
Ok, I believe this implements pagination as discussed in #16 and may address performance issues outlined in #95 as well.

I certainly mucked around quite a bit in here so feel free to let me know if I'm overstepping. Pagination required some small tweaks in how the query was being executed and passed around. Specifically, instead of using `db.Session.query`, I used `Model.query` as the original approach did not expose the pagination api. I also added some explicit relationships in the model declarations which allows the face/image lookups to be handled at rendering vs. within the query. I think this should speed things up a bit, hopefully addressing the worker time out issues in #95.

The styling of the pagination controls are very basic and will need to be updated. I'm happy to tackle that separately at a different point in time.

Apologies for the messy commit log. My original implementation conflicted with changes in the develop branch and so I decided to rebase and redo to hopefully make this easier to incorporate. I also pulled back my pull request from yesterday re: changes to run.py and included that here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,6 @@ vagrant/puppet/.tmp
 
 #Dirs created in dev
 OpenOversight/db_repository/
+
+# Editor files
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -78,8 +78,5 @@ vagrant/puppet/.tmp
 .librarian
 .vagrant
 
-#Dirs created in dev
-OpenOversight/db_repository/
-
 # Editor files
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,9 @@ vagrant/puppet/.tmp
 
 #Dirs created in dev
 OpenOversight/db_repository/
+<<<<<<< HEAD
 
 # Editor files
+=======
+>>>>>>> 98df3c8cefe7af4d1f558dcb130c33cdbeeae674
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,6 @@ vagrant/puppet/modules/
 vagrant/puppet/.tmp
 .librarian
 .vagrant
+
+#Dirs created in dev
+OpenOversight/db_repository/
--- a/.gitignore
+++ b/.gitignore
@@ -80,9 +80,6 @@ vagrant/puppet/.tmp
 
 #Dirs created in dev
 OpenOversight/db_repository/
-<<<<<<< HEAD
 
 # Editor files
-=======
->>>>>>> 98df3c8cefe7af4d1f558dcb130c33cdbeeae674
 .idea/
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -33,11 +33,6 @@ def grab_officers(form):
                                                 Officer.birth_year == None))
     return officer_query
 
-def grab_officers(form):
-    officer_query = db.session.query(Assignment, Officer).join(Officer)
-    officer_query = filter_by_form(form, officer_query)
-    return officer_query.all()
-
 
 def grab_officer_faces(form):
     officer_query = db.session.query(Assignment, Officer, Face, Image) \
@@ -64,7 +59,6 @@ def sort_officers_by_photos(all_officers, officers_w_images):
             officers.append(officer)
 
     return officers, all_officer_images
-
 
 
 def allowed_file(filename):
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -9,7 +9,9 @@ import pdb
 db = SQLAlchemy(app)
 
 
-def filter_by_form(form, officer_query):
+def grab_officers(form):
+    # officer_query = db.session.query(Officer, Assignment).join(Assignment)
+    officer_query = Officer.query.join(Assignment)
     if form['race'] in ('BLACK', 'WHITE', 'ASIAN', 'HISPANIC', 'PACIFIC ISLANDER'):
         officer_query = officer_query.filter(Officer.race.like('%%{}%%'.format(form['race'])))
     if form['gender'] in ('M', 'F'):
@@ -30,7 +32,6 @@ def filter_by_form(form, officer_query):
                                                         Officer.birth_year >= max_birth_year),
                                                 Officer.birth_year == None))
     return officer_query
-
 
 def grab_officers(form):
     officer_query = db.session.query(Assignment, Officer).join(Officer)
@@ -61,8 +62,9 @@ def sort_officers_by_photos(all_officers, officers_w_images):
         else:
             all_officer_images.update({officer.Officer.id: 'static/images/placeholder.png'})
             officers.append(officer)
-    
+
     return officers, all_officer_images
+
 
 
 def allowed_file(filename):
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -2,6 +2,7 @@ import config
 import datetime
 
 from flask_sqlalchemy import SQLAlchemy
+from sqlalchemy import desc, asc
 from app import app
 from app.models import Officer, Assignment, Image, Face
 import pdb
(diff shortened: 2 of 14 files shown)


OpenOversight/app/utils.py
```diff
@@ -29,40 +29,41 @@ def filter_by_form(form, officer_query):
...
+    return filter_by_form(form, Officer.query)
 
 
-def grab_officer_faces(form):
-    officer_query = db.session.query(Assignment, Officer, Face, Image) \
-                                    .join(Officer).join(Face).join(Image)
-    officer_query = filter_by_form(form, officer_query)
-    officer_images = officer_query.all()
-    return officer_images
+# def grab_officer_faces(form):
```
r4v5: you can just remove the commented out code instead of commenting it out

OpenOversight/app/utils.py
```diff
@@ -29,40 +29,41 @@ def filter_by_form(form, officer_query):
...
-
-    for officer in all_officers:
-        if officer.Officer.id in all_officer_ids_w_photos:
-            continue
-        else:
-            all_officer_images.update({officer.Officer.id: 'static/images/placeholder.png'})
-            officers.append(officer)
-    
-    return officers, all_officer_images
+#
```
r4v5: same here

OpenOversight/app/views.py
```diff
@@ -1,14 +1,17 @@
 import os
 from flask import (render_template, request, redirect, url_for,
-                  send_from_directory, flash, session)
+                   send_from_directory, flash, session)
 from werkzeug import secure_filename
 from app import app
 
-from utils import (allowed_file, grab_officers, grab_officer_faces,
-                  sort_officers_by_photos)
+# from utils import (allowed_file, grab_officers, grab_officer_faces,
```
r4v5: and here

OpenOversight/test_data.py
```diff
@@ -21,22 +21,32 @@ def populate():
     po4 = models.Officer(last_name='Cuddleme', first_name='Haywood', middle_initial='U',
                          race='HISPANIC', gender='F', birth_year=1950, pd_id=1,
                          employment_date=datetime(2014, 4, 4, 1, 1, 1))
+    po5 = models.Officer(last_name='Klozoff', first_name='Oliver', middle_initial='U',
+                         race='WHITE', gender='M', birth_year=1950, pd_id=1,
+                         employment_date=datetime(2004, 4, 4, 1, 1, 1))
+    po6 = models.Officer(last_name='O\'Problem', first_name='Bea', middle_initial='U',
```
r4v5: these changes are making it increasingly clear we need to just iterate over something to create test data, which I'll try to grab this week.

OpenOversight/app/utils.py
```diff
@@ -29,40 +29,41 @@ def filter_by_form(form, officer_query):
     officer_query = officer_query.filter(db.or_(db.and_(Officer.birth_year <= min_birth_year,
                                                         Officer.birth_year >= max_birth_year),
                                                 Officer.birth_year == None))
+    # This handles the sorting upstream of pagination and pushes officers w/o tagged faces to the end of list
+    officer_query = officer_query.outerjoin(Face).order_by(Face.id)
```
r4v5: This is a bit on the bounds of my depth here, so if I'm wrong, feel free to correct me:
I don't know whether postgres enforces strict ordering of returned values, but an issue I've seen elsewhere in pagination code that called out to other RDBMSes is that the data can be returned in a way that causes duplicates and missing entries if the order_by clause is non-unique.
I know that `face` has a FK to `officer`, but i see nothing in the app right now that would make that mapping 1-to-1: many `officer`s (including ones without images but I don't know if any do right now) don't have a `face` entry.
So in a scenario where we have, say, 21 un-`face`d officers, since we're making the queries at different times there's nothing in the `order_by` clause to make that ordering guaranteed.
I see some model changes above to possibly address this, but I'm not convinced they're sufficient and it'd be an insidious and difficult-to-catch bug.

.gitignore
```diff
@@ -77,3 +77,9 @@ vagrant/puppet/modules/
 vagrant/puppet/.tmp
 .librarian
 .vagrant
+
+#Dirs created in dev
+OpenOversight/db_repository/
```
r4v5: So, I did this too, and on [reading the docs](https://flask-migrate.readthedocs.io/en/latest/) I'm not sure that we want to actually gitignore this. You've made changes to the declaration of the models, but I don't see any actual migrations generated, which leads me to believe that a) we are using flask-migrate wrong and b) the model changes below won't actually show up.

OpenOversight/config.py
```diff
@@ -15,6 +15,9 @@
 UNLABELLED_UPLOADS = 'uploads/'
 ALLOWED_EXTENSIONS = set(['png', 'jpg', 'jpeg', 'mpeg', 'mp4'])
 
+# pagination
+OFFICERS_PER_PAGE = 2
```
r4v5: how about an OFFICERS_PER_PAGE of os.environ.get('OFFICERS_PER_PAGE') or 20, so that we can override it for testing and still have a large set of results in prod?

.gitignore
scuerda: So the model changes that made are not actually db-level changes, which is why no migration was required (I think). As far as I know, the sqlalchemy orm lets you declare relationships in this way without requiring schema changes. Both the models I touched already had FK declarations, but the addition of a relationship declaration just lets you traverse the relationships as if they are properties of the returned object. Which is to say, I don't think anything about how flask-migrate is being used is incorrect.

OpenOversight/config.py
scuerda: Yep, makes sense to me. I'll modify.

OpenOversight/app/utils.py
r4v5: The only thing guaranteed to be unique is Officer.id, so whatever else we sort by, that should be the outermost sort to break ties.

freddymartinez9:
I did a quick review and it certainly looks good to me, or at least a very good start. I'll let @redshiftzero sign off on it but I'm not seeing major issues. We may also need to rope in someone to test the cases after we commit it.

OpenOversight/app/utils.py
redshiftzero: I'm definitely happy with `py.test`!

OpenOversight/test_data.py
```diff
@@ -21,22 +21,32 @@ def populate():
     po4 = models.Officer(last_name='Cuddleme', first_name='Haywood', middle_initial='U',
                          race='HISPANIC', gender='F', birth_year=1950, pd_id=1,
                          employment_date=datetime(2014, 4, 4, 1, 1, 1))
+    po5 = models.Officer(last_name='Klozoff', first_name='Oliver', middle_initial='U',
```
redshiftzero: Thanks for conforming with our Springfield Police Department test data

OpenOversight/test_data.py
```diff
@@ -21,22 +21,32 @@ def populate():
...
-    po1 = models.Officer.query.get(1)
-    po2 = models.Officer.query.get(2)
-    po3 = models.Officer.query.get(3)
-    po4 = models.Officer.query.get(4)
+    # po1 = models.Officer.query.filter(models.Officer.last_name == 'Tinkle').first()
+    # po2 = models.Officer.query.filter(models.Officer.last_name == 'Jass').first()
+    # po3 = models.Officer.query.filter(models.Officer.last_name == 'Tinkle').first()
+    # po4 = models.Officer.query.get(4)
+    # po5 = models.Officer.query.get(5)
+    # po6 = models.Officer.query.get(6)
```
redshiftzero: Feel free to delete this stuff now

.gitignore
redshiftzero: Yep it looks like these changes won't require a migration [for the reason](http://docs.sqlalchemy.org/en/latest/orm/tutorial.html#orm-tutorial-relationship) @scuerda describes. (Btw we're actually using `sqlalchemy-migrate` for the database migrations here, we _can_ switch to something else like `flask-migrate` (which uses `alembic`) if we encounter issues)

redshiftzero:
Ah, I see @r4v5 still has changes requested, @r4v5 were they addressed? If so we can go ahead and merge in (one conflict we will need to resolve with `CONTRIB.md` but should be aight).

.gitignore
r4v5: I love how many packages Python has but sometimes I really hate how many packages Python has :)

r4v5: That said, this gitignore entry does need to go

OpenOversight/test_data.py
scuerda: ok, done.

PR #110, opened 2016-10-23
Fix for case sensitive query

redshiftzero:
Fix for bug found by @r4v5 using our swanky new staging server!

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -14,7 +14,7 @@ db = SQLAlchemy(app)
 def filter_by_form(form, officer_query):
     if form['name']:
         officer_query = officer_query.filter(
-            Officer.last_name.ilike(func.upper('%%{}%%'.format(form['name'])))
+            Officer.last_name.ilike('%%{}%%'.format(form['name']))
             )
     if form['race'] in ('BLACK', 'WHITE', 'ASIAN', 'HISPANIC',
                         'PACIFIC ISLANDER'):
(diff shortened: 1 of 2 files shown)


OpenOversight/app/utils.py
```diff
@@ -14,7 +14,7 @@
 def filter_by_form(form, officer_query):
     if form['name']:
         officer_query = officer_query.filter(
-            Officer.last_name.like('%%{}%%'.format(form['name']))
+            Officer.last_name.like('%%{}%%'.format(form['name'].upper))
```
r4v5: We should make sure the data import scripts we use for ETL actually do a .upper on the strings they throw into the DB if we're going to say "everything gets case-insensitive compares to .upper" in the query.

OpenOversight/test_data.py
```diff
@@ -9,22 +9,22 @@
 
 def populate():
     """ Populate database with test data"""
-    po1 = models.Officer(last_name='Tinkle', first_name='Ivana', race='BLACK',
+    po1 = models.Officer(last_name='TINKLE', first_name='IVANA', race='BLACK',
```
r4v5: I kinda want to leave at least one of these mixed-case so that we can have a test case around actually returning it for queries

freddymartinez9: You can image there are officers who have mixed cased last names (i.e. `McAuliffe`)

redshiftzero: Well.. in the database everything is uppercase. But let me fix the query to make it explicit (may not get to it tonight but will do at some point)

freddymartinez9: in that case, i think this approach is fine

redshiftzero:
Added mixed case test row and modified query to use ILIKE, in the future this filter should work no matter what the case is in the db

r4v5:
lgtm

PR #112, opened 2016-11-03
Facedetect

JoshuaOpolko:
Fixed typos in the readme and added a script for removing non-face images using Google Vision.


redshiftzero:
Josh - thanks for this great work! I'm going to merge this in as is so people can check it out. Once we get the uploading implemented in the app code, we should get this face detection integrated into the workflow.

Issue #38, opened 2016-09-24
One-command deploys


r4v5:
Once we have migrations and a local dev environment, we'll want the ability to deploy this thing in a far more straightforward manner than we currently have. I'm open to suggestions; I'm thinking either a rake task that rsyncs things or a capistrano setup, but that's because most of my deployment orchestration experience is in Ruby. For all I know, there's a better solution here, and if so, we should find it and use it.

scuerda:
I've done a fair amount with `fabric` and it's fairly easy to work with. I've used it to do essentially what it sounds like is already being done (`git pull` followed by other commands).

One thing to consider is that now that travis close to being up and running, it might be nice to use those tools to test and deploy automatically. Not sure how that would work w/ `fabric`, but it would be easy enough with a shell script.

freddymartinez9:
@scuerda if you can point me to some resources where this is set up, I can modify them and set them up. I've never used `fabric` but could figure it out with an existing example.

scuerda:
I'll throw up a gist and share the link.

scuerda:
@freddymartinez9 [Here](https://gist.github.com/scuerda/3d4aa909a8be00d27c2618d29d046d9b) is a variant of one for a CKAN project I manage.

This lets me deploy with `fab prod deploy` or dump and download the database with `fab prod backup`

[Here](https://gist.github.com/onyxfish/299803) is another one that is MUCH more robust, but maybe a bit dated at this point.

This has some nice things like rollback functions, etc. In general, it isn't too different from working with something like gulp/grunt.

freddymartinez9:
thanks Sasha! assigning to myself and Mason.

PR #121, opened 2016-11-16
Working fabric file

freddymartinez9:
initial one line deploy for #38 . A couple of caveats, the staging environment was throwing `out: /bin/bash: /root/.bash_profile: Permission denied` which I don't quite understand. but it works!  If @r4v5 or @scuerda have ideas about making it better, let me know. Also, it has not been tested in prod, only because I assume we only want to deploy dot releases. Is that correct @redshiftzero ? 

```
fab staging deploy
Executing task 'deploy'
run: source /home/nginx/.bashrc
sudo: git pull
out: /bin/bash: /root/.bash_profile: Permission denied
out: From https://github.com/lucyparsons/OpenOversight
out:  + bcf9292...818c0d5 onelinefabric -> origin/onelinefabric  (forced update)
out: Already up-to-date.
out: 

[162.243.156.49] run: sudo systemctl restart openoversight

Done.
```

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -1,0 +1,29 @@
+from __future__ import with_statement
+from fabric.api import env, local, run, sudo, cd, hosts, get
+from fabric.context_managers import prefix
+import datetime
+
+env.use_ssh_config = False
+
+# Hosts
+# env.hosts list references aliases in ~/.ssh/config or IP address. When using .ssh/config,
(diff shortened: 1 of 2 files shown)


redshiftzero:
Yeah we should only deploy releases (that we've tested) to production

freddymartinez9:
@scuerda based on your feedback I re-did the fabric. I rebased and will merge in now.

PR #122, opened 2016-11-17
Sadly include all our puppet modules in the repo

r4v5:
I don't like doing this for a few reasons, but it's so much easier for
new folks to start out if they don't have to also install puppet on
their local machines in order to provision their VMs.

I am erring on the side of making this easier for folks to get up and
running with. The trade-off is that when I make provisioning changes,
the diffs will be larger, louder, and more annoying. After helping
several folks get this provisioning stuff working, I'm leaning more and
more on the side of 'just commit it.


r4v5:
I'm going to use this PR to also make our dev infrastructure Xenial. Wish me luck, as 14.04 has been the most dev-and-ops-hostile Ubuntu release I've seen thus far.

r4v5:
Okay, I'm not going to do the xenial update, because xenial and gunicorn aren't playing super well with the puppet module for managing python. I made a bit of a refactor of the manifest so that it'll "just work" if/when puppet-python gets updated. If we run into any "prod is different from dev" issues, I'll start down this rabbit hole myself instead of waiting for others to get there first, but until that point, i'm punting.

PR #136, opened 2016-12-19
Fix fallout from #126's refactoring

r4v5:
The move to flask app factories had some fallout.
First and most obnoxiously, the app wouldn't start in a wsgi context
because we were doing some weird cross-importing and double-defining
of the db instance and messed up some relative imports. `db` as an
object now belongs to `models.py`, and everything else asking for a
db has to import and grab it from there.
 (most notably `app/__init__.py` and `utils.py`, which was for some reason
 instantiating its own flask app instance and grabbing _its_ db
 instance instead?).

In addition, our non-wsgi instantiations were doing some naughty things,
so I've refactored them to grab the db object from models.py as well, so
things like 'provisioning a vm' now work happily, with generation of
test data happening.

The relative path of the db_repository changed when we changed to using
relative imports, so a simple `git mv` put it where the `create_db.py`
script expected it to be.

Something (I'm not going to dig into exactly what) needed python-cffi,
which failed to install due to libffi not being installed on the vagrant.
Things mostly-worked but I'm not a fan of big scary messages in provisioning.
Add it to puppet's package list to get it installed.

Testing still fails on my vagrant:
```
(oovirtenv)vagrant@vagrant-ubuntu-trusty-64:/vagrant/OpenOversight$ py.test -v --cov=app
============================================================================================== test session starts ==============================================================================================
platform linux2 -- Python 2.7.6, pytest-3.0.3, py-1.4.32, pluggy-0.4.0 -- /home/vagrant/oovirtenv/bin/python
cachedir: ../.cache
rootdir: /vagrant, inifile:
plugins: pep8-1.0.6, cov-2.4.0
collected 26 items

tests/test_models.py::test_officer_repr Killed
(oovirtenv)vagrant@vagrant-ubuntu-trusty-64:/vagrant/OpenOversight$
```
But this gets the app back into a state where it's not actively broken.

Nota bene: i also did not fix the etl scripts, which still use the old-old way of grabbing a db connection.


freddymartinez9:
this actually looks pretty good. I think I'll hold off on merging it since I'm not as competent to review as @redshiftzero (or even @scuerda) but I am happy-ish

Issue #98, opened 2016-10-18
Faces are squished 


redshiftzero:
I just noticed that a bunch of our faces get kinda squished in the gallery view and it's hard for humans to recognize faces when distorted... If a frontend person wants to rock up in here and fix that I would be very happy

freddymartinez9:
I believe this is controlled by this line in `app/static/css/bootstrap.css`, although someone should with CSS knowledge would know more
```
.img-thumbnail {
  padding: 4px;
  line-height: 1.42857143;
  background-color: #2b3e50;
  border: 1px solid #dddddd;
  border-radius: 0;
  -webkit-transition: all 0.2s ease-in-out;
  -o-transition: all 0.2s ease-in-out;
  transition: all 0.2s ease-in-out;
  display: inline-block;
  max-width: 100%;
  height: auto;
}
```

freddymartinez9:
<img width="404" alt="screen shot 2016-12-19 at 14 26 41" src="https://cloud.githubusercontent.com/assets/3100220/21327961/34e136cc-c5f7-11e6-9de2-8c8b069a5c63.png">
So I had tried the change you committed and it resulted in this on the front-end. You can check it out on OO staging. I think the real fix might be to use static width and height (i.e. 200x200) images for images.

redshiftzero:
Waaah. Okay assigning myself, will figure out why this isn't working in Chrome.

redshiftzero:
Faces are no longer squished, but there is some extra whitespace. I'm gonna leave the frontend alone for now because fixing some dead space is low priority.

PR #137, opened 2016-12-20
Rewrite gallery template and modify CSS of thumbnail images

redshiftzero:
OK another attempt at fixing #98! Try this one @freddymartinez9. I ended up rewriting this template because it was a mess, and in the process learned that [bootlint](https://github.com/twbs/bootlint) - an HTML linter for Bootstrap projects - was a thing! (Probably I should do this for the other templates but not a super high priority atm). I tested this in Safari and Chrome and on mobile using [Chrome's Device Mode](https://developers.google.com/web/tools/chrome-devtools/device-mode/).

### DIFF
--- a/OpenOversight/app/static/theme.css
+++ b/OpenOversight/app/static/theme.css
@@ -24,7 +24,7 @@ body {
 .thumbnail_container {
      float:left;
      max-height:300px;
-     max-width:auto;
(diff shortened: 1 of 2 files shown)


freddymartinez9:
I can consistently reproduce it! thanks for your PR.

PR #138, opened 2016-12-21
Fix run.py to run dev server

redshiftzero:
After #136, run.py was in the wrong location, so this PR moves it and updates the developer documentation (in case people are using `run.py` and not `gunicorn` in the dev environment).


r4v5:
Good catch, I knew I'd miss one or two of these lingering references. :+1:

PR #139, opened 2016-12-21
Serve nicely formatted error pages

redshiftzero:
Make error pages nicer for UX wins:

* Adds GET requests to the /*gallery views since a not-very-user-friendly “Method Not Allowed” error appears when a user refreshes a page (we are correctly POSTing the form values, and after validation we are serving a redirect such that the last request the browser sees is a GET).
* Adds nicely formatted error pages for 404, 403 (needed in advance of the upcoming user accounts branch), and 500 errors.

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -25,13 +25,12 @@ def get_officer():
 def label_data():
     form = FindOfficerIDForm()
     if form.validate_on_submit():
-        #  flash('[DEBUG] Forms validate correctly')
         return redirect(url_for('main.get_tagger_gallery'), code=307)
     return render_template('label_data.html', form=form)
 
 
-@main.route('/tagger_gallery/<int:page>', methods=['POST'])
-@main.route('/tagger_gallery', methods=['POST'])
+@main.route('/tagger_gallery/<int:page>', methods=['GET', 'POST'])
+@main.route('/tagger_gallery', methods=['GET', 'POST'])
 def get_tagger_gallery(page=1):
     form = FindOfficerIDForm()
     if form.validate_on_submit():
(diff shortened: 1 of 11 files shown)


redshiftzero:
Before: 
before

After: 
after

OpenOversight/app/templates/403.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
+{% block title %}Forbidden{% endblock %}
+
+{% block content %}
+
+<div class="container theme-showcase" role="main">
+
+<h1>Forbidden</h1>
+<p>You do not have permissions to view this page. <a href="{{ url_for('main.get_tagger_gallery') }}">Return to homepage</a>.</p>
```
r4v5: why are these all linking to `get_tagger_gallery()`? that's a weird place to end up compared to `index()`, no?

OpenOversight/app/main/views.py
```diff
@@ -46,8 +45,8 @@ def get_tagger_gallery(page=1):
         return redirect(url_for('main.label_data'), code=307)
 
 
-@main.route('/gallery/<int:page>', methods=['POST'])
-@main.route('/gallery', methods=['POST'])
+@main.route('/gallery/<int:page>', methods=['GET', 'POST'])
+@main.route('/gallery', methods=['GET','POST'])
```
r4v5: what is the actual effect of refreshing the page the visitor ends up on (the one that's been loaded as a GET)? Do the params come along such that folks can still use stuff, or is it an error?

OpenOversight/app/templates/403.html
redshiftzero: Lol yepppp that is weird. Fixed in 22bcf9e

redshiftzero:
Aight so there is an outstanding issue explained in #140 regarding the bare `/gallery` and `/tagger_gallery` views but otherwise this can go in

freddymartinez9:
LGTM

Issue #79, opened 2016-10-15
Security Audit


freddymartinez9:
We should do a basic penetration test and internal code audit so that we don't get hacked and can harden out application. This will also become more important once we fix #5 and if we have users logging in. I can lead this project.

redshiftzero:
Moving this off the 0.2 milestone. Someone should really do this. 😑

PR #145, opened 2016-12-27
Add Volunteer User Accounts

### LINKED ISSUE #114: User accounts for volunteers
We should make have volunteers log in before they add tags to the database. This way we can:

-  restrict this action to logged in users
-  monitor if volunteers are putting fraudulent tags in the database and block those that do
-  keep a leaderboard so that we can exploit that strange quirk of human psychology to motivate users to tag moar

We can also use these accounts for image submission in order to keep a leaderboard of "images submitted" there too. When submitting images of officers, people should have the option of using a pseudonym or the option of not logging in and being anonymous (including using Tor, which we currently allow and will continue to).

redshiftzero:
This PR closes #114, and implements user accounts for OpenOversight volunteers. **Important:** Don't deploy to staging without me since this will require a database migration so we should be a wee bit careful :cowboy_hat_face: .  

Adds some new dependencies:

* Flask-Bootstrap to automate some of the template generation for forms
* Flask-Login for managing user sessions
* Flask-Mail for sending emails, which is needed for resetting passwords and confirming user email addresses

This PR adds a new blueprint for authentication tasks under `/auth`. In more detail this:

* Adds login and register links to top of the base template:

screen shot 2016-12-24 at 1 47 07 pm

If a user is logged in, then the link at the top right of the navigation bar of the base template will point to a dropdown menu with several account settings a user can modify:

screen shot 2016-12-27 at 2 48 54 am

* User login page: 

screen shot 2016-12-27 at 3 03 44 am

* User registration page:

screen shot 2016-12-27 at 3 04 31 am

* During registration, an email with a token is sent, a user must confirm their email before they can access protected views. Adds dependencies Flask-Mail (for the emailing) and itsdangerous (for the secure tokens). Emails are sent using another thread, eventually we should use Celery workers for this task (and maybe other tasks as we scale up). Email confirmation looks like: 

screen shot 2016-12-27 at 2 46 35 am

If a user tries to sign in before they confirm their email address, they see:

screen shot 2016-12-27 at 2 47 02 am

* Password reset is also implemented (why we needed to have a working email address). 

screen shot 2016-12-27 at 3 05 20 am

* Password change: 

screen shot 2016-12-27 at 3 04 50 am

* Email address change: 

screen shot 2016-12-27 at 3 05 04 am

* I also refactored the `/label` route a bit (much more to come) to only show the signup links if a user is not signed in:

screen shot 2016-12-27 at 2 59 23 am

And to show the OO primary key search tool if they are signed in:

screen shot 2016-12-27 at 2 56 25 am

* Adds some unit tests for password salting and hashing in `test_models.py`. Will try to write more tests soon

I pretty much followed this [really nice book](http://shop.oreilly.com/product/0636920031116.do) I just finished reading about Flask to implement most of this: I highly recommend.


freddymartinez9:
I have a question here about `OpenOversight/app/auth/views.py`, it looks like we are not invalidating the user's cookie from the current authenticated users? I think that created a security vulnerability. 

```
+def logout():
 +    logout_user()
 +    flash('You have been logged out.')
 +    return redirect(url_for('main.index'))
``` 

same question for this function
```
def password_reset_request():
 +    if not current_user.is_anonymous:
 +        return redirect(url_for('main.index'))
 +    form = PasswordResetRequestForm()
 +    if form.validate_on_submit():
 +        user = User.query.filter_by(email=form.email.data).first()
 +        if user:
 +            token = user.generate_reset_token()
 +            send_email(user.email, 'Reset Your Password',
 +                       'auth/email/reset_password',
 +                       user=user, token=token,
 +                       next=request.args.get('next'))
 +        flash('An email with instructions to reset your password has been '
 +              'sent to you.')
 +        return redirect(url_for('auth.login'))
 +    return render_template('auth/reset_password.html', form=form)
```

freddymartinez9:
I spoke to @redshiftzero in person about this and we agreed that this can be merged in without looking at my comments. @redshiftzero believes it should be handled by the `Flask LoginManager`. We can address concerns within #79, thank you for the PR.

Issue #5, opened 2016-07-07
Do file uploads with WTForms


redshiftzero:
To enable users to SAFELY upload certain file formats. These uploads should go to S3 and then a row should be inserted into `officers.raw_images`.

r4v5:
possible inspiration: https://github.com/tornadoweb/tornado/blob/master/demos/s3server/s3server.py

freddymartinez9:
Discussion from our Slack channel:
```
@r4v5  i’m wondering if ‘s3, apply directly to the bucket’ is the right answer here, i keep waffling back and forth on it 
@b_meson
we dont have a high volume of submissions. I’m also thinking maybe S3 since our concern about trolls flooding and wasting out money might be overstated
@akama   
s3 is really handy for stuff like this, it's what I use for another application
@b_meson [13:30]  
i’ll leave it up to Dr Helsby, lead developer of OO to decide. I think S3 might be fine. make it amazon’s problem. POST to s3, then return it. We need to routinely do database and S3 backups anyways, which means we can handle malicious files (i.e. malware) in our export. once I write the python fabric for this, I can execute it in a QubesVM so any malware would be isolated. and just remove any non images
@redshiftzero
basically to do the Qubes convert to trusted PDF for images
```

redshiftzero:
Relevant: https://www.owasp.org/index.php/Unrestricted_File_Upload

redshiftzero:
In progress in branch `image-uploading`

redshiftzero:
This was implemented and merged in #149, closing!

Issue #148, opened 2017-01-03
Privacy-respecting captcha or other rate limiting mechanism on photo, tag, and classification submissions


redshiftzero:
Before we remove the login_required from the `/submit` view, we should put up some rate limiting in the form of a privacy-respecting captcha.

redshiftzero:
I was going to integrate a CAPTCHA into the submit page, but I see that there does not appear to be a nice CAPTCHA flask extension that would work well with this project (Flask-WTF has a CAPTCHA, but it's ReCAPTCHA, and the privacy implications are prohibitive given the nature of OpenOversight).

In PR #224, I've instead decided to put stricter rate limiting on the submission page. Depending on the level of spam or misbehavior we see, we can reopen this issue and take further steps. If we see misbehavior on the volunteer pages, we can do the same stricter rate limiting there (I think we will likely see lower spam rates elsewhere on the application as a confirmed account is required to visit the volunteer pages).

Issue #90, opened 2016-10-18
Automatic metadata stripping


redshiftzero:
We got a request for stripping of metadata from some submitted images, but really we should just be doing this for all images. Potentially the only metadata to keep would be timestamp EXIF data that has when the image was taken (something we are trying to keep track of in the database), so we should grab that and dump the rest.

JoshuaOpolko:
I'll add this to the Flickr download script so that exif data is removed when photos are downloaded.  The 'taken' timestamp will be added to the database and then removed from photos.

freddymartinez9:
The easiest way to do that would be to use exiftools in a bash script. I'm sure others have written some as well, so I can take this task on.

freddymartinez9:
does this conflict with this comment in `models.py`
```
 74     # We might know when the image was taken e.g. through EXIF data
 75     date_image_taken = db.Column(db.DateTime, index=True, unique=False, nullable=True)
```

redshiftzero:
That is the column that stores the timestamp EXIF data, described here (from issue description):
> Potentially the only metadata to keep would be timestamp EXIF data that has when the image was taken (something we are trying to keep track of in the database), so we should grab that and dump the rest.

redshiftzero:
We should adopt the approach of Qubes: 

1. Open the image in a sandboxed environment (e.g. spin up a new instance on amazon EC2 or use amazon serverless architecture)
2. Use exiftool to grab the date the image was taken if available 
3. Create a 'trusted' image from the raw RGB values in the image
4. Export the RGB values and the date the image was taken
5. Teardown machine

This should do a decent job of handling potentially malicious images as well as stripping out metadata

redshiftzero:
I wrote a CLI tool in Golang that does the above (using go `libexif` bindings) https://github.com/redshiftzero/scrubber minus the sandboxing part - reads in an image file, outputs it and also outputs JSON of the metadata so we can pull out the date the image was taken. If one does this on openoversight, they might call out to that.... or they can use whatever they want

if you build it, i will merge it

PR #149, opened 2017-01-03
Photo submissions to S3

redshiftzero:
This PR closes issue #5, to implement in-app photo uploads to S3. 

Submission route `/submit` is now available to logged in users of the application:

screen shot 2017-01-03 at 12 53 03 am

Anonymous users will continue to use the Google Drive submission. However, once #90 (sanitize images (both to handle potentially malicious uploads and to remove metadata)) and #148 (privacy-respecting captcha on this form) are implemented then we can remove the login required decorator from this view and deprecate the Google drive submission form. 

Details:
* Uploads are limited in size and in file type (to images).
* Images are uploaded first to `/tmp` and then to the S3 bucket for OpenOversight. We should eventually have the S3 upload and deletion of the file from `/tmp` be a celery task. The first part (saving to `/tmp`) is quite fast
* A row is inserted in the images table with the date of upload, the hash of the image, the path to the image on S3.
* We disallow the upload of duplicate images in the form by first checking that there is not a matching hash already in the images table in the database. 

After we deploy this, we can go in and upload the non-troll pics from Google drive and the useful pics from our social media scrape to S3 using this form. Troll harder CPD

### DIFF
--- a/OpenOversight/app/config.py
+++ b/OpenOversight/app/config.py
@@ -33,8 +33,8 @@ class BaseConfig(object):
     # AWS Settings
     AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID')
     AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY')
-    AWS_DEFAULT_REGION = 'us-west-2'
-    S3_BUCKET_NAME = 'openoversight'
+    AWS_DEFAULT_REGION = os.environ.get('AWS_DEFAULT_REGION')
+    S3_BUCKET_NAME = os.environ.get('S3_BUCKET_NAME')
 
     # Upload Settings
     MAX_CONTENT_LENGTH = 16 * 1024 * 1024
--- a/OpenOversight/app/config.py
+++ b/OpenOversight/app/config.py
@@ -13,10 +13,6 @@ class BaseConfig(object):
     SQLALCHEMY_TRACK_MODIFICATIONS = False
     SQLALCHEMY_COMMIT_ON_TEARDOWN = True
 
-    # File Upload Settings
-    UNLABELLED_UPLOADS = 'uploads/'
-    ALLOWED_EXTENSIONS = set(['png', 'jpg', 'jpeg', 'mpeg', 'mp4'])
-
     # pagination
     OFFICERS_PER_PAGE = os.environ.get('OFFICERS_PER_PAGE', 20)
 
@@ -33,6 +29,15 @@ class BaseConfig(object):
     OO_MAIL_SUBJECT_PREFIX = '[OpenOversight]'
     OO_MAIL_SENDER = 'OpenOversight <OpenOversight@gmail.com>'
     # OO_ADMIN = os.environ.get('OO_ADMIN')
+
+    # AWS Settings
+    AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID')
+    AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY')
+    AWS_DEFAULT_REGION = 'us-west-2'
+    S3_BUCKET_NAME = 'openoversight'
+
+    # Upload Settings
+    MAX_CONTENT_LENGTH = 16 * 1024 * 1024
 
     SEED = 666
 
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -108,20 +108,19 @@ def submit_data():
             try:
                  url = upload_file(safe_local_path, original_filename,
                                    new_filename)
+                 # Update the database to add the image
+                 new_image = Image(filepath=url, hash_img=hash_img, is_tagged=False,
+                                   date_image_inserted=datetime.datetime.now(),
+                                   # TODO: Get the following field from exif data
+                                   date_image_taken=datetime.datetime.now())
+                 db.session.add(new_image)
+                 db.session.commit()
+
+                 flash('File {} successfully uploaded!'.format(original_filename))
             except:
                 flash("Your file could not be uploaded at this time due to a server problem. Please retry again later.")
             os.remove(safe_local_path)
             os.rmdir(tmpdir)
-
-            # Update the database to add the image
-            new_image = Image(filepath=url, hash_img=hash_img, is_tagged=False,
-                              date_image_inserted=datetime.datetime.now(),
-                              # TODO: Get the following field from exif data
-                              date_image_taken=datetime.datetime.now())
-            db.session.add(new_image)
-            db.session.commit()
-
-            flash('File {} successfully uploaded!'.format(original_filename))
         else:
             flash('This photograph has already been uploaded to OpenOversight.')
     elif request.method == 'POST':
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -7,7 +7,7 @@ from flask_login import (LoginManager, login_user, logout_user,
 from werkzeug import secure_filename
 
 from . import main
-from ..utils import grab_officers, roster_lookup, upload_file, hash_file
+from ..utils import grab_officers, roster_lookup, upload_file, compute_hash
 from .forms import FindOfficerForm, FindOfficerIDForm, HumintContribution
 from ..models import db, Image
 
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -81,18 +81,18 @@ def submit_complaint():
 def submit_data():
     form = HumintContribution()
     if request.method == 'POST' and form.validate_on_submit():
-        filename = secure_filename(request.files[form.photo.name].filename)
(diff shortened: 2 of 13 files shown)


freddymartinez9:
assigning to myself and @r4v5 for review. I worry about the dangling `tmp/` but perhaps we can chroot it or something else (make read-only) to clean up infosec concerns.

OpenOversight/tests/test_utils.py
```diff
@@ -53,13 +53,7 @@ def test_filter_by_badge_no(mockdata):
...
-    for e in extension:
-        test_filename = 'test.{}'.format(e)
-	assert OpenOversight.app.utils.allowed_file(test_filename) == True
-
-
-def test_not_allowed_filenames(app):
-    test_filename = 'test.gif'
-    assert OpenOversight.app.utils.allowed_file(test_filename) == False
+def test_hash_file(mockdata):
+    hash_result = OpenOversight.app.utils.hash_file('bacon')
```
freddymartinez9: where is the file we are expecting to hash `bacon` actually coming from? the test data directory?

freddymartinez9:
Besides my one question and @r4v5's review, I think it should work correctly.

OpenOversight/tests/test_utils.py
redshiftzero: Ah no, I'll rename this function when I rebase this PR on current develop to be more clear, it should probably be called simply `hash`. Here it's just taking the sha256 of the string `bacon`

OpenOversight/app/main/views.py
```diff
@@ -74,26 +76,37 @@ def submit_complaint():
...
+    form = HumintContribution()
+    if request.method == 'POST' and form.validate_on_submit():
+        filename = secure_filename(request.files[form.photo.name].filename)
+        image_data = request.files[form.photo.name].read()
+
+        # See if there is a matching photo already in the db
+        hash_img = hash_file(image_data)
+        hash_found = Image.query.filter_by(hash_img=hash_img).first()
+
+        if not hash_found:
```
r4v5: say someone uploads two files named 'Photo01.jpg' with different hashes simultaneously. We run in gunicorn so each thread runs without shared state excepting the filesystem.
`filename = secure_filename('Photo01.jpg')`, then we hash the image(s), then we check the database for hash duplicates, and then we write it out to the filesystem at '/tmp/Photo01.jpg'. Which one gets uploaded?

Also, we're trusting workzeug's `safe_filename`, which is about things that are safe to drop into a folder but not necessarily safe to drop into /tmp itself. What if someone manages to clobber something important-ish (like, say, a socket) in /tmp this way? Can we use the `tempfile` package in stdlib to specify the local filename to use instead?

OpenOversight/app/utils.py
```diff
@@ -1,11 +1,30 @@
...
+
+def hash_file(file_to_hash):
+    return hashlib.sha256(file_to_hash).hexdigest()
+
+
+def upload_file(src_filename, dest_filename):
+    s3_client = boto3.client('s3')
+    s3_client.upload_file('/tmp/{}'.format(src_filename),
+                          current_app.config['S3_BUCKET_NAME'],
+                          dest_filename,
```
r4v5: (loading the s3 web console is a terrible experience with 150k files in the same dir too, trust me)

OpenOversight/tests/test_utils.py
redshiftzero: Fixed in commit ab55d39.

OpenOversight/app/utils.py
redshiftzero: Good point - fixed in commit 17536cb.

OpenOversight/app/config.py
```diff
@@ -34,6 +30,15 @@ class BaseConfig(object):
     OO_MAIL_SENDER = 'OpenOversight <OpenOversight@gmail.com>'
     # OO_ADMIN = os.environ.get('OO_ADMIN')
 
+    # AWS Settings
+    AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID')
+    AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY')
+    AWS_DEFAULT_REGION = 'us-west-2'
+    S3_BUCKET_NAME = 'openoversight'
```
redshiftzero: Makes sense, should I then remove these things from here?

OpenOversight/app/main/views.py
redshiftzero: Addressed the first comment in commit f7bd0f7 by never using the original filename and always using the hash of the file to store both on the webserver filesystem and on S3.

Addressed this in commit ab636de to use `tempfile` to create a temporary directory and then store the file in there, again never using the original filename from the user (I followed [this guide](https://security.openstack.org/guidelines/dg_using-temporary-files-securely.html) to do so).

OpenOversight/app/config.py
r4v5: No, sorry. I meant that `AWS_DEFAULT_REGION` and `S3_BUCKET_NAME` should be set in .env and therefore not hardcoded to values in `config.py`.

redshiftzero: Ahh, word. Okay changed this in commit 48be6c2

OpenOversight/tests/test_utils.py
```diff
@@ -53,13 +53,7 @@ def test_filter_by_badge_no(mockdata):
...
+    expected_hash = '9cca0703342e24806a9f64e08c053dca7f2cd90f10529af8ea872afb0a0c77d4'
+    assert hash_result == expected_hash
```
redshiftzero: OK so I did almost this but not quite: Added a test for the file storage path in commit 66c2bc9 by monkeypatching the boto3 connection (which actually uncovered a bug which I fixed in 47fdf4b). 

Is the library you mean https://github.com/jubos/fake-s3 which looks like https://github.com/jserver/mock-s3 for python?

redshiftzero:
OK guys, I'm finished making changes for now: take a look at these changes and let me know if there are further concerns.

redshiftzero:
And thanks for all the great comments!! Made this way better :+1: 💯

OpenOversight/app/main/views.py
```diff
@@ -74,26 +80,49 @@ def submit_complaint():
...
+
+            # Save temporarily on local filesystem
+            tmpdir = tempfile.mkdtemp()
+            safe_local_path = os.path.join(tmpdir, new_filename)
+            with open(safe_local_path, 'w') as tmp:
+                tmp.write(image_data)
+            os.umask(SAVED_UMASK)
+
+            # Upload file from local filesystem to S3 bucket and delete locally
+            url = upload_file(safe_local_path, original_filename, new_filename)
```
r4v5: this line can throw boto exceptions if s3 is down or the credentials are incorrect; do we want to display an informative message to the user when this happens (`flash("Your file could not be uploaded at this time due to a server problem. Please retry again later.")` or the like)? I think if we don't, we'll just show the default nginx unstyled 500.

requirements.txt
```diff
@@ -1,5 +1,7 @@
 flask>=0.11.1
 python-dotenv
+mock
```
r4v5: these can go into `dev-requirements.txt` once #163 lands.

OpenOversight/app/main/views.py
```diff
@@ -74,26 +80,49 @@ def submit_complaint():
...
+        original_filename = secure_filename(request.files[form.photo.name].filename)
+        image_data = request.files[form.photo.name].read()
+
+        # See if there is a matching photo already in the db
+        hash_img = compute_hash(image_data)
+        hash_found = Image.query.filter_by(hash_img=hash_img).first()
+
+        if not hash_found:
+            # Generate new filename
+            file_extension = original_filename.split('.')[-1]
```
r4v5: `FileAllowed` in `forms.py` asserts that the file uploaded ends in `{.jpg,.jpeg,.png}` right? Because if not, this could be dicey. Can we throw a test or two in that assert that a file with name ending in .jpg, .jpeg, .png gets accepted and more importantly, one with a name ending in, say, `/../../../../../~/etc/passwd` doesn't, so we capture this requirement in our tests?

requirements.txt
redshiftzero: Fixed in c492804

OpenOversight/app/main/views.py
redshiftzero: Good point, added in fc41ad2

OpenOversight/app/main/views.py
redshiftzero: Yep this is true, I've added a couple of tests in 5548ba5 to check that files named "passwd" (due to the way that WTForms validates the file upload form, the file has to exist, so I did this "passwd" check in order to get that test passing both on the dev VM and on Travis) and files with invalid file extensions are not uploaded.

redshiftzero:
OK I think I've addressed the remainder of the comments here - I also added some additional unit tests for the login and logout routes. Check it out and let me know 😃

redshiftzero:
Yeaaaah!! Thanks y'all 😄

PR #155, opened 2017-01-03
Speed up unit tests

redshiftzero:
Tests were taking a while because we were using a realistic number of officers (if we use all ~15k officers Travis builds time out), so I've made that variable config-specific. We use a realistic number of officers for the development config and a small set of officers for the unit tests and Travis config. 

Also updated the instructions in `CONTRIB.md` to instruct people how to use `pytest` as the unit test runner.

### DIFF
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -53,11 +53,9 @@ or
 
 ## Running Unit Tests
 
- Run tests with `nose`:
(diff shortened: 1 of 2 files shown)


redshiftzero:
Also this is a tiny microchange so I am going to go ahead and use my admin powers to merge to `develop` so I can rebase my other PR on this.

PR #156, opened 2017-01-03
Add unit tests for authentication routes

redshiftzero:
Adds a few unit tests for the new auth blueprint


freddymartinez9:
I like it. thanks and merging.

Issue #162, opened 2017-01-07
Add geckodriver, firefox, xvfb install to dev VM


redshiftzero:
For functional tests we need Firefox, geckodriver (selenium 3 requires this to drive Firefox), and xvfb (virtual display framebuffer required for headless programs) set up in the development VM. 

In the `functional-test` branch I've added this to CI - you can see the [Travis.yml](https://github.com/lucyparsons/OpenOversight/blob/functional-test/.travis.yml) here - note xvfb is installed by default in the Travis CI environment.

r4v5:
This has been done on a commit in #163.
```
(oovirtenv)vagrant@vagrant-ubuntu-trusty-64:/vagrant/OpenOversight$ pytest 
============================================================================= test session starts ==============================================================================
platform linux2 -- Python 2.7.6, pytest-3.0.3, py-1.4.32, pluggy-0.4.0
rootdir: /vagrant, inifile: 
plugins: pep8-1.0.6, cov-2.4.0
collected 50 items 

tests/test_functional.py ...    
tests/test_models.py ...................
tests/test_routes.py ....................
tests/test_utils.py ........

========================================================================== 50 passed in 89.80 seconds ==========================================================================
```
:rocket:

PR #163, opened 2017-01-07
Add basic functional tests

### LINKED ISSUE #58: Integration tests
We need some integration tests to check the end-to-end functionality of our webapp. Should probably use selenium for this

redshiftzero:
This closes #58, to add some basic functional tests to the app for the complainant use case. These are run using `pytest`. Requirements are added in this PR to run the functional tests in Travis CI. I've made issue #162 to have the functional test requirements added to the development environment.


OpenOversight/tests/test_functional.py
```diff
@@ -0,0 +1,29 @@
+import time
+
+
+def test_user_can_load_homepage_and_get_to_form(mockdata, browser):
+    browser.get("http://localhost:5000")
+
+    # Complainant loads homepage
+    assert "OpenOversight" in browser.title
+    browser.find_element_by_link_text("Try it").click()
```
r4v5: It's best to assign an id to that button (in https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/templates/index.html#L9) and then find the element by id. This makes it less fragile and in a world where we have i18n concerns (which, in a multicultural city like Chicago should be p. soon) we wouldn't need different integration tests for English, Spanish, Polish, Arabic, Tagalog, Chinese, etc

OpenOversight/tests/test_functional.py
```diff
@@ -0,0 +1,29 @@
...
+    time.sleep(5)
```
r4v5: Explicit sleeps is a selenium antipattern; it makes the suite slower than it needs to be and brittle and doesn't solve the problem you want solved. Check http://stackoverflow.com/a/26567563 for options in how to do it in python. Basically, waiting until an element is visible, or dying if it's not visible in x seconds.

OpenOversight/tests/test_functional.py
```diff
@@ -0,0 +1,29 @@
...
+    assert "Find an Officer" in browser.page_source
+
+
+def test_user_can_use_form_to_get_to_gallery(mockdata, browser):
+    browser.get("http://localhost:5000/find")
+
+    # Complainant submits default search params and gets list of officers
+    browser.find_element_by_name("submit-officer-search-form").click()
+    time.sleep(10)
+    assert "Digital Gallery" in browser.page_source
```
r4v5: Can we have better asserts here, like the results box is a div that exists with sufficient child elements? If not I'm okay with this for now.

requirements.txt
```diff
@@ -5,6 +5,7 @@ Flask-WTF>=0.13.1
 Flask-Login
 Flask-Mail
 psycopg2>=2.6.2
+selenium
```
redshiftzero: Sure makes sense. To the best of my knowledge that's how it's usually handled, so in commit 8043c0b I've split the requirements into two files, one that contains just the requirements for staging/prod (`requirements.txt`) and then another file `dev-requirements.txt` that contains the additional requirements required for dev/testing. We will also want to modify the development VM environment to have it install both files

redshiftzero: (I handled the Travis CI part in 8043c0b)

requirements.txt
```diff
@@ -17,3 +18,4 @@ faker==0.7.3
 pytest-cov==2.4.0
 pytest-pep8==1.0.6
 itsdangerous
+xvfbwrapper
```
redshiftzero: :+1: done in 8043c0b

OpenOversight/tests/test_functional.py
redshiftzero: :+1: fixed in commit 167804a

OpenOversight/tests/test_functional.py
redshiftzero: Word, good feedback here. To address this in commit 167804a I made helper function `wait_for_element` for the functional tests to wait for an element of interest to be present or throw a selenium `TimeoutException` and fail the test. I also added a context manager `wait_for_page_load()` for page clicks such that we wait for the new page to finish loading (I got that from [this page](http://www.obeythetestinggoat.com/how-to-get-selenium-to-wait-for-page-load-after-a-click.html)).

OpenOversight/tests/test_functional.py
redshiftzero: Added an assert looking for one of the officer results in commit 167804a

PR #166, opened 2017-01-16
Add admin management interface

redshiftzero:
I integrated the Flask extension [Flask-Script](https://flask-script.readthedocs.io/en/latest/) with OpenOversight, this creates a Django-like manage.py interface for administration and dev tasks. 

* One can use `python manage.py runserver` to spawn the server (replaces `run.py`, for dev only).

* One can use `python manage.py shell` to create a Python interpreter prompt in the Flask application context (avoids having to do a bunch of annoying imports if you want to try things out interactively (especially useful for testing out SQLAlchemy ORM queries)).

* Finally, one now also uses this to make OpenOversight administrator accounts both in dev and staging/prod:

```
(oovirtenv)vagrant@vagrant-ubuntu-trusty-64:/vagrant/OpenOversight$ python manage.py make_admin_user
Username: redshiftzero
Email: jen@redshiftzero.com
Password: 
Type your password again: 
Administrator redshiftzero successfully added
```

This enables admins to sidestep account email confirmation (nice in dev environment). Note that this is the _only_ way to create admin accounts - there is no function in the webapp to make administrator accounts, only accounts with regular privs.


redshiftzero:
Added new commands for migrations support over to `manage.py`, deprecated `downgrade_db.py`, `upgrade_db` and `migrate_db`. Instead one does e.g.:

```
(oovirtenv)vagrant@vagrant-ubuntu-trusty-64:/vagrant/OpenOversight$ python manage.py migrate_db
New migration saved as /vagrant/OpenOversight/app/db_repository/versions/002_migration.py
```

redshiftzero:
Thanks for review @r4v5! Good call - those two commands I did not put into `manage.py` yet but when I do I will update the puppet manifest, added a reminder note to #167

Issue #147, opened 2016-12-27
Better password validation


redshiftzero:
Right now we are letting users pick their passwords with some extremely simple validation. We could use our weak human brains to come up with some rules and code these as validators in the form, but a better solution would be to just use Dropbox's nice [zxcvbn password strength estimation library](https://blogs.dropbox.com/tech/2012/04/zxcvbn-realistic-password-strength-estimation/) to check password strength.

redshiftzero:
Implemented by @Te-k in PR #168 (committed in 7628327a1844ef669e5995238bd388a1f660c3cd) , so closing!

PR #168, opened 2017-01-19
Implement better password validation #147

Te-k:
Here is a proposal implementation for #147 using zxcvbn


OpenOversight/app/templates/auth/register.html
```diff
@@ -9,8 +9,82 @@
...
+    var password2 = document.getElementById('password2');
+    var meter = document.getElementById('password-strength-meter');
+    var meter2 = document.getElementById('password-confirmation-meter');
+    var text = document.getElementById('password-strength-text');
+
+    function check_confirmation() {
+        var val1 = password.value;
+        var val2 = password2.value;
+        console.log(val1);
+        console.log(val2);
```
redshiftzero: Can remove these two `console.log` lines

Te-k: Oops, yes I will :)

Te-k:
Nope, I will be happy to do both this week :)

redshiftzero:
Awesome! Feel free to ping me for feedback or a re-review 💪

Te-k:
Here it is : 
* Code migrated into a separated js file
* Same client side check on /change-password
* Form requirement of 8 password minimum character (I also changed login to 6 min, 8 is really long, is it ok?)
* No more console.log :stuck_out_tongue:

redshiftzero:
Nice!! Changes look great, just tested them out. Also :+1: to changing login to 6 minimum, I'm cool with that. 

One last thing, it still looks like the password strength is not being used to disallow registration of accounts with weak passwords - for example, I was able to register an account with the password `password` - legitimately scored as "Very Bad" by the code. But since it was 8 characters or greater none of the form validators threw an error.

Te-k:
Yes, the client side is only showing an indication, the only forced constraint is the 8 char minimum server side. I thought it was your idea with zxcvbn, did you want to block the form client side ?

It is possible, I just need to work a bit on it to be sure that the reason is clear for the user.

Te-k:
I just improved the js to disable the submit button if the password is not strong enough. If someone can check that I don't fall into some weird js implementation on their browser it would be cool.

freddymartinez9:
It looks good to me! I'm approving for merge.

redshiftzero:
Tested the changes in Firefox, Safari, and Chrome. Looks great!! Thanks for this @Te-k! Gonna merge in...

redshiftzero:
(Merged in these commits manually to resolve merge conflicts)

Issue #186, opened 2017-02-13
Log certain things like exceptions/errors so we have visibility into the app


r4v5:
Certain things, especially things that cause exceptions or errors have no logging around them to surface that to admins and they should.
Example:
https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/main/views.py#L297
"Uploading a file failed." Why? Was it incorrect S3 permissions? Something else? We should log the original exception instead of swallowing it.
Another example:
Creating an admin user should log _something_.

redshiftzero:
This was completed in #189, so closing!

PR #189, opened 2017-02-22
Add logging of application events and errors

redshiftzero:
Implements #186:

* Logs error when a exception occurs uploading an officer image to S3 (also I've verified that the photo upload does work in the development environment with a test S3 bucket). 
* Logs info event when an administrator account is created.

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -9,6 +9,7 @@ import re
 from sqlalchemy.exc import IntegrityError
 import sys
 import tempfile
+from traceback import format_exc
 from werkzeug import secure_filename
 
 from . import main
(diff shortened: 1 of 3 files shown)


OpenOversight/app/main/views.py
```diff
@@ -297,6 +298,9 @@ def submit_data():
 
                  flash('File {} successfully uploaded!'.format(original_filename))
             except:
+                current_app.logger.error('Error uploading to S3: {}'.format(
+                    sys.exc_info()
```
r4v5: do we just want to to_string this tuple? I'm fine if so, I just want error logging :)

redshiftzero: Full traceback is now saved as of 391fed3 👸

redshiftzero:
Anything else I should add to the logging here @r4v5 or is this one good to go in? If it looks good we can deploy to staging and figure out what the issue was with S3.

PR #195, opened 2017-02-26
Fix filepath field length

### LINKED ISSUE #194: Image uploads fail on staging due to too-narrow db column
```
DataError: (psycopg2.DataError) value too long for type character varying(120)
 [SQL: 'INSERT INTO raw_images (filepath, hash_img, date_image_inserted, date_image_taken,
 contains_cops, user_id, is_tagged) VALUES (%(filepath)s, %(hash_img)s, %(date_image_inserted)s,
 %(date_image_taken)s, %(contains_cops)s, %(user_id)s, %(is_tagged)s) RETURNING raw_images.id']
[parameters: {'contains_cops': None, 'user_id': None, 'date_image_taken': datetime.datetime(2017, 2, 26, 4, 8, 28, 769435),
'filepath': 'https://s3-us-west-1.amazonaws.com/openoversight-staging/3f/0d3a44de1d03863733253ba72c412f0f325e3358580bee35dcb425fa442233.jpg',
'date_image_inserted': datetime.datetime(2017, 2, 26, 4, 8, 28, 769395),
'is_tagged': False, 'hash_img': '3f0d3a44de1d03863733253ba72c412f0f325e3358580bee35dcb425fa442233'}]
```
(that image isn't a cop, but a test image. Mea culpa for polluting our test database.)

redshiftzero:
Fixes #194, where the filepath in the images table was too small. 

Also, in the course of making this change, I ran into a situation where the automatically generated migrations by `sqlalchemy-migrate` did not work, and instead I had to [manually write the migrations](https://sqlalchemy-migrate.readthedocs.io/en/v0.7.1/changeset.html). To make it easier for developers in the future, I've modified `manage.py` a bit (this workflow will seem familiar to Django people):

* First, a developer will run `python manage.py makemigrations` in order to have `sqlalchemy-migrate` attempt to automatically generate the migrations. 
* Then, a developer will inspect and edit them (if necessary) and _then_ will run `python manage.py migrate` to actually apply the migrations. 

For deploying to staging and prod, since we're committing the migrations as we ensure that they work, the admin will just need to run `python manage.py migrate`.

### DIFF
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -64,16 +64,16 @@ In addition to running the development server, `manage.py` (OpenOversight's mana
 ```
 (oovirtenv)vagrant@vagrant-ubuntu-trusty-64:/vagrant/OpenOversight$ python manage.py
 usage: manage.py [-?]
-                 {shell,downgrade_db,runserver,upgrade_db,migrate_db,make_admin_user}
+                 {shell,makemigrations,migrate,downgrade_db,runserver,make_admin_user}
                  ...
 
 positional arguments:
-  {shell,downgrade_db,runserver,upgrade_db,migrate_db,make_admin_user}
+  {shell,makemigrations,migrate,downgrade_db,runserver,make_admin_user}
     shell               Runs a Python shell inside Flask application context.
+    makemigrations      Make database migrations
+    migrate             Migrate/upgrade the database
     downgrade_db        Downgrade the database
     runserver           Runs the Flask development server i.e. app.run()
-    upgrade_db          Upgrade the database
-    migrate_db          Migrate the database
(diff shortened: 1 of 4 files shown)


r4v5:
LGTM. Do we want the migrate step to be run as part of `fab deploy`?

redshiftzero:
Hmm, probably not in `deploy` since we don't always want to run it, however we could add another function in the `fabfile` for the apply database migrations case

freddymartinez9:
Also looks good to me, I can do write a `fab staging migrate` with both options, should I push a commit as part of this PR?

redshiftzero:
Let's merge this in and you can do a followup PR. Gotta get these bits moving

PR #200, opened 2017-03-03
Update travis to use Ubuntu Trusty docker image

r4v5:
I have no idea why it's running our tests twice or freaking out, as the case may be.


freddymartinez9:
yes, this is what we are running in production and in Vagrant.

PR #201, opened 2017-03-03
Don't clean up resources that don't exist

r4v5:
This _might_ fix the travis freakout that's just started happening.
Please let this get a full travis run and a lookover by @redshiftzero.


redshiftzero:
Thanks for working on troubleshooting this @r4v5 ❤️. Since we've since fixed the builds and these resources will always exist (we specify they will exist during setUp), I am gonna close this one.

Issue #245, opened 2017-04-13
Update or remove tutorial


redshiftzero:
The tutorial is old and needs to be updated. 

Though perhaps if there were enough informative tooltips throughout the user flow, a tutorial would be unnecessary... 🤔

ShawnaStewart:
Could you elaborate on tutorial? Or post the link to which part you're specifically referring to? I think you're talking about the instructions throughout the volunteer flow but just want to verify. I can work on the ux and ui for this 👍

redshiftzero:
Ah sorry about that: it's [this page](https://openoversight.lucyparsonslabs.com/tutorial). UX/UI help on this page would be great! 

The minimal thing we need is that page updating (primarily the out of date screenshots 😞)  and making as easy/friendly to read as possible, whether through tooltips or something else. Let me know your thoughts :metal:

ShawnaStewart:
I started working on a flow and layout for this. Will follow up next weekend :-)

redshiftzero:
Awesome! 😄

redshiftzero:
Adding this extremely sick flowchart that @ShawnaStewart made:

<img width="905" alt="screen shot 2017-07-08 at 7 19 49 pm" src="https://user-images.githubusercontent.com/7832803/27990665-7b66537a-6412-11e7-93bf-0b28bb521dc8.png">

jujueyeball:
Dope

On Jul 8, 2017 10:22 PM, "redshiftzero" <notifications@github.com> wrote:

Adding this extremely sick flowchart that @ShawnaStewart
<https://github.com/shawnastewart> made:

[image: screen shot 2017-07-08 at 7 19 49 pm]
<https://user-images.githubusercontent.com/7832803/27990665-7b66537a-6412-11e7-93bf-0b28bb521dc8.png>

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/lucyparsons/OpenOversight/issues/245#issuecomment-313893234>,
or mute the thread

PR #267, opened 2017-05-11
Classify cops quickly with hotkeys :-D

### LINKED ISSUE #266: UX: classifying submissions (contains cop or not) too slow
I know the answer in < 1 second, but it takes 2 seconds to scroll down past the large image and click 'Yes' or 'No'.

Shortcut keys?

elimisteve:
## Status

Ready for review

## Description of Changes

Fixes #266 

 - Added hotkeys for cop classification: No/Skip/Yes keyboard shortcuts are n/s/y
 - Moved No/Skip/Yes buttons to top of page (easier to click; no need to scroll down, plus the buttons don't move)

## Tests and linting
 
 [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/templates/sort.html
+++ b/OpenOversight/app/templates/sort.html
@@ -35,14 +35,6 @@ $(document).bind('keydown', 's', function(){
       </div>
 
       <div class="row">
-        <div class="frontpage-leads">
-          <div>
-            <img class="center-block img-responsive" src="{{ path }}" alt="Picture to be sorted">
-          </div>
-        </div>
-      </div>
-
-      <div class="row">
         <div class="col-sm-4 text-center">
           <form action="{{ url_for('main.classify_submission', image_id=image.id, contains_cops=0) }}" method="post">
             <button type="submit" name="action" id="answer-no" class="btn btn-lg btn-danger">
@@ -62,6 +54,14 @@ $(document).bind('keydown', 's', function(){
               Yes!
             </button>
           </form>
+        </div>
+      </div>
+
+      <div class="row">
+        <div class="frontpage-leads">
+          <div>
+            <img class="center-block img-responsive" src="{{ path }}" alt="Picture to be sorted">
+          </div>
         </div>
       </div>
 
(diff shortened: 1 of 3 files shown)


elimisteve:
@redshiftzero I can squash these if you'd like

OpenOversight/app/templates/sort.html
```diff
@@ -13,36 +35,48 @@
...
+      </div>
+
+      <div class="row">
+        <div class="frontpage-leads">
+          <div>
+            <img class="center-block img-responsive" src="{{ path }}" alt="Picture to be sorted">
+          </div>
+        </div>
+      </div>
+
```
redshiftzero: One nice way that we could communicate these shortcuts to users would be to underline the first letter of each word on the buttons themselves. What do you think?

elimisteve: That is certainly more elegant, but is it clear enough?

elimisteve: If you think that'd be clear enough to the user, I can make that change.

Options:

1. Tell the user about the shortcuts in a pop-up once (but only on desktop), and store in a cookie (or in their account profile, but not exposed in the UI) that they've seen it and know (so we don't show them again
    - Sounds complicated
2. Underline first letter of each + mention these shortcuts in the FAQ
3. Underline first letter of each + mention shortcuts elsewhere (blurb when they first log in?)

jujueyeball: 4) move yes no skip options above foto and have non-disappearing text: "Tip: sort faster by using y, n, and x"

elimisteve: @jujueyeball This PR puts those buttons above the photo, so that part's done. Could add that tip, but that's similar to what I've implemented above, more elegant, but also a bit less clear IMO.

elimisteve: @redshiftzero Replacement PR that makes the change you suggested above: https://github.com/lucyparsons/OpenOversight/pull/269

redshiftzero: It's a fair question re: is it clear enough but I'm hoping that the replacement PR you made (:angel:) should provide enough prompts to the user. Re: other places where the shortcuts list should go, I think that in the FAQ and on the tutorial definitely make sense :+1: :+1:. I've added a note to the checklist in #245.

redshiftzero:
Closing in favor of #269

Issue #296, opened 2017-08-28
Store department name in database


redshiftzero:
Let's store the name of the police department in `config.py` (if you don't think that's a good place, speak up) so that "Chicago Police Department" is not hardcoded everywhere in the HTML templates

r4v5:
Is this as a stepping stone to multiple PDs in one DB, or in lieu of it, or can they both coexist in a way I'm not seeing?

redshiftzero:
Ideally, this codebase would support both (1) multiple police departments in one database as well as (2) enable this code to be easily forked to be used for independent instances of OpenOversight. I was thinking that `config.py` was the right place to store this, but that works for the 2nd goal but not the first, so I've changed this issue to be storing the name of the department in the database. We should move all department-specific data into the database (i.e. name of department, rank levels, possibly more).

redshiftzero:
Done in #300, and removing hardcoded "Chicago Police Department" in index template is now an item in followup #308

PR #300, opened 2017-09-10
Initial multi department support

redshiftzero:
## Status

Ready for review

## Description of Changes

This PR adds first steps towards initial multi-department support (partially #296):

* Adds a 'departments' model to enable the web application to store which police departments the instance of OpenOversight supports. 

* Adds a new page (`/departments`) to list and manage basic data about each new police department. We can extend this to store the ranks for each department and any other department-specific information.

* Adds a new admin-only form to add a new department (`/department/new`)

## Multi-City Plan

1. After this is deployed, I will go into the admin panel and add "Chicago Police Department" as the first department. 
2. In a followup PR I will further decouple the code from Chicago (e.g. once step 1 is done then I can remove the hardcoded "Chicago Police Department" from all templates) and modify the other models (e.g. officers, units, etc.) to foreign key them to this first department. 

## Testing

Especially because this includes a migration, during review one should test this before merging it, so you should do something like this:

1. Provision development VM on 8a272dcccd2895be9cd90f6f3fe1de1262372429 (that commit makes the test user an admin)
2. `vagrant ssh`
3. `source ~/oovirtenv/bin/activate`
4. From your host OS: `git checkout multi-department-support`
5. Back in the VM: `cd /vagrant/OpenOversight`
7. `python manage.py migrate`
8. `python manage.py runserver`
9. Verify the functionality claimed in this PR is accurate, and recall that the test account credentials are `test@example.org` with password `testtest`.

## Screenshots

screen shot 2017-09-10 at 1 24 45 am

## Tests and linting
 
 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/main/forms.py
+++ b/OpenOversight/app/main/forms.py
@@ -105,10 +105,10 @@ class AssignmentForm(Form):
 class DepartmentForm(Form):
     name = StringField(
         'Full name of police department, e.g. Chicago Police Department',
-        default='', validators=[Regexp('\w*'), Length(max=500), DataRequired()]
+        default='', validators=[Regexp('\w*'), Length(max=255), DataRequired()]
     )
     short_name = StringField(
         'Shortened acronym for police department, e.g. CPD',
-        default='', validators=[Regexp('\w*'), Length(max=10), DataRequired()]
+        default='', validators=[Regexp('\w*'), Length(max=100), DataRequired()]
     )
     submit = SubmitField(label='Add')
--- a/OpenOversight/app/main/forms.py
+++ b/OpenOversight/app/main/forms.py
@@ -100,3 +100,15 @@ class AssignmentForm(Form):
     unit = QuerySelectField('unit', validators=[Optional()],
                             query_factory=unit_choices)
     star_date = DateField('star_date', validators=[Optional()])
+
+
+class DepartmentForm(Form):
+    name = StringField(
+        'Full name of police department, e.g. Chicago Police Department',
+        default='', validators=[Regexp('\w*'), Length(max=500), DataRequired()]
+    )
+    short_name = StringField(
+        'Shortened acronym for police department, e.g. CPD',
+        default='', validators=[Regexp('\w*'), Length(max=10), DataRequired()]
+    )
+    submit = SubmitField(label='Add')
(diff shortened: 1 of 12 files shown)


OpenOversight/app/main/forms.py
```diff
@@ -100,3 +100,15 @@ class AssignmentForm(Form):
     unit = QuerySelectField('unit', validators=[Optional()],
                             query_factory=unit_choices)
     star_date = DateField('star_date', validators=[Optional()])
+
+
+class DepartmentForm(Form):
+    name = StringField(
+        'Full name of police department, e.g. Chicago Police Department',
+        default='', validators=[Regexp('\w*'), Length(max=500), DataRequired()]
```
r4v5: Is there a reason `Length` on these don't match the columns? Entries of length 255<x<500 for `name` will throw an exception in OpenOversight/app/main/views.py:add_department()

OpenOversight/tests/test_routes.py
```diff
@@ -606,3 +608,55 @@ def test_user_cannot_change_password_if_they_dont_match(mockdata, client,
...
+        form = DepartmentForm(name='Test Police Department',
+                              short_name='TPD')
+
+        rv = client.post(
+            url_for('main.add_department'),
+            data=form.data,
+            follow_redirects=True
+        )
+
+        assert 'New department' in rv.data
```
r4v5: Probbably worth also asserting that the row is actually in the DB.

OpenOversight/tests/test_routes.py
```diff
@@ -606,3 +608,55 @@ def test_user_cannot_change_password_if_they_dont_match(mockdata, client,
...
+        login_admin(client)
+
+        form = DepartmentForm(name='Chicago Police Department',
+                              short_name='CPD')
+
+        rv = client.post(
+            url_for('main.add_department'),
+            data=form.data,
+            follow_redirects=True
+        )
```
r4v5: I forget; this gets auto-cleaned between runs, right?

redshiftzero: You are correct - the `teardown()` in the `mockdata` pytest fixture in `conftest.py` will remove the data in the database between tests.

redshiftzero:
Thanks for reviewin' @r4v5 :heart:. I've made your suggested improvements in 564d210db89bb5037d472c5c106cc067efb3f1c1.

PR #310, opened 2017-09-24
Enable admin to add new officers and units for any department

### LINKED ISSUE #213: "Add cop" admin only route
Add a new officer to the database.

redshiftzero:
## Status

Ready for review

## Description of Changes

This PR:

  * Adds department_id columns to `Officer` and `Unit` 
  * Adds `manage.py` command to link existing officers and units to the first department
  * Adds admin-only "Add officer" feature (for any department). Closes #213.
  * Adds admin-only "Add unit" feature (for any department).

screen shot 2017-09-24 at 12 00 12 pm
screen shot 2017-09-24 at 12 00 19 pm
screen shot 2017-09-24 at 12 00 26 pm

## Notes for Deployment

```
$ pip install -r ../requirements.txt
$ python manage.py db upgrade
$ python manage.py link_officers_to_department
```

## Tests and linting
 
- [x] `flake8` checks pass


OpenOversight/app/db_repository/versions/006_migration.py
```diff
@@ -0,0 +1,42 @@
...
+)
+
+officers = Table('officers', post_meta,
+    Column('id', Integer, primary_key=True, nullable=False),
+    Column('last_name', String(length=120)),
+    Column('first_name', String(length=120)),
+    Column('middle_initial', String(length=120)),
+    Column('race', String(length=120)),
+    Column('gender', String(length=120)),
+    Column('employment_date', DateTime),
```
r4v5: this has the potential for weird timezone bugs, but i don’t know how much potential it has.

Specifically i don’t remember if postgres’s datetime support includes zone or assumes the application handles it; there’s a potential for discrepancy when initializing these with dates (at time 00:00) and being returned referencing the day before/after we intended.

OpenOversight/app/db_repository/versions/006_migration.py
```diff
@@ -0,0 +1,42 @@
+from sqlalchemy import *
+from migrate import *
+
+
+from migrate.changeset import schema
+pre_meta = MetaData()
+post_meta = MetaData()
+unit_types = Table('unit_types', post_meta,
```
r4v5: what exactly is this table holding? units or unit types? unit type to me implies a general type with a one-to-many relationship from unit type to department but we make department id an int column, implying 1-to-1.

r4v5: also, ‘descrip’ feels like a weird choice of column name; is there a reason to not just use name here?
(asking these questions early while refactoring/migrations are still relatively easy)

OpenOversight/app/db_repository/versions/006_migration.py
```diff
@@ -0,0 +1,42 @@
...
+officers = Table('officers', post_meta,
+    Column('id', Integer, primary_key=True, nullable=False),
+    Column('last_name', String(length=120)),
+    Column('first_name', String(length=120)),
+    Column('middle_initial', String(length=120)),
+    Column('race', String(length=120)),
+    Column('gender', String(length=120)),
+    Column('employment_date', DateTime),
+    Column('birth_year', Integer),
+    Column('pd_id', Integer),
```
r4v5: what is pd_id and how does it differ from department_id?

OpenOversight/app/main/views.py
```diff
@@ -218,6 +219,55 @@ def add_department():
...
+                          last_name=form.last_name.data,
+                          middle_initial=form.middle_initial.data,
+                          race=form.race.data,
+                          gender=form.gender.data,
+                          birth_year=form.birth_year.data,
+                          employment_date=form.employment_date.data,
+                          department_id=form.department.data.id)
+        db.session.add(officer)
+        db.session.commit()
+        assignment = Assignment(officer_id=officer.id,
```
r4v5: worth noting that because we commit the session for officer before assignment, it’s possible to fail to create an officer+assignment duo due to things like db constraints or connectivity loss and have actually created the officer underlying it, which sounds like a way to accidentally get duplicate officers.

OpenOversight/app/models.py
```diff
@@ -33,6 +33,7 @@ class Officer(db.Model):
     pd_id = db.Column(db.Integer, index=True, unique=False)
     assignments = db.relationship('Assignment', backref='officer', lazy='dynamic')
     face = db.relationship('Face', backref='officer', lazy='dynamic')
+    department_id = db.Column(db.Integer)
```
r4v5: do we not want to explicitly state these are foreign keys here and in the migration?

r4v5: also what happens if someone changes departments, say from UCPD to OPD?

OpenOversight/app/models.py
```diff
@@ -62,6 +63,7 @@ class Unit(db.Model):
 
     id = db.Column(db.Integer, primary_key=True)
     descrip = db.Column(db.String(120), index=True, unique=False)
+    department_id = db.Column(db.Integer)
```
r4v5: same FK comment

redshiftzero: Fixed in 495df5d0e783ed831e39b3893b9c9f3c954800ae

OpenOversight/app/main/views.py
redshiftzero: Modified this to commit the two together in a91f3f7833c900bd6c5c498973e394517887c9ed

OpenOversight/app/db_repository/versions/006_migration.py
redshiftzero: hmm so this table holds the unit names / descriptions, I linked them to department as I think it's nice to store it this way because when a user adds an officer assignment, the webapp could show them only the units that are associated with that police department

redshiftzero:
Well, I tried to squash the migrations in this PR, got annoyed at how finicky `sqlalchemy-migrate` is, so I ripped it out and replaced it with `alembic` (via `Flask-Migrate`). I've upgrade tested this and all went well. Deployment instructions in the PR are updated. I'm going to merge this now since I'm about to push more changes on top of the changes in this PR

PR #354, opened 2018-01-21
Set content-type on uploaded images

### LINKED ISSUE #290: Images are served with the wrong Content-Type
Images are currently served with `Content-Type: binary/octet-stream` which is fine for `` elements.
However, attempting to view the image via Right-Click -> View image in Firefox for example prompts the user to download the image.

Please serve images with `Content-Type`s corresponding to their filetypes/file magic.

r4v5:
@bmintz pointed out in #290 that we don't set content-type on the images we upload to S3, leading to some frustrating browser experiences.
This naively sets them based on file extension, with the assumption that browsers will figure it out if someone uploads, say, a .png with a .gif extension.
It also adds a test to make sure that any added ALLOWED_EXTENSIONS are handled properly in `guess_mime_type()` to make sure we have programmatic prompting to keep the list up to date.
I don't actually know how deep inside the s3 library call we want to go to add a test asserting that a mock of the call has mime-type set, so I punted on that.
This doesn't fix historical S3 uploads, unfortunately, which I'll have to do by manually (well, programmatically, but outside of OO) re-PUTing them.

Tested that the test works by adding a new potential filetype to app/config.py and seeing it fail:
```
=================================================================================== FAILURES ===================================================================================
_____________________________________________________________________________ test_guess_mime_type _____________________________________________________________________________

    def test_guess_mime_type():
        for extension in current_app.config['ALLOWED_EXTENSIONS']:
            file_name = "testimage.%s" % extension
>           assert OpenOversight.app.utils.guess_mime_type(file_name) != "binary/octet-stream", "undefined mime-type for extension %s" % extension
E           AssertionError: undefined mime-type for extension gifv
E           assert 'binary/octet-stream' != 'binary/octet-stream'
E            +  where 'binary/octet-stream' = <function guess_mime_type at 0x7f661b216b90>('testimage.gifv')
E            +    where <function guess_mime_type at 0x7f661b216b90> = <module 'OpenOversight.app.utils' from '/vagrant/OpenOversight/app/utils.pyc'>.guess_mime_type
E            +      where <module 'OpenOversight.app.utils' from '/vagrant/OpenOversight/app/utils.pyc'> = <module 'OpenOversight.app' from '/vagrant/OpenOversight/app/__init__.pyc'>.utils
E            +        where <module 'OpenOversight.app' from '/vagrant/OpenOversight/app/__init__.pyc'> = OpenOversight.app

tests/test_utils.py:95: AssertionError
===================================================================== 1 failed, 13 passed in 12.33 seconds =====================================================================
```

## Status

Ready for review

## Description of Changes

Fixes #290.

## Notes for Deployment

None, though I'm going to have to go through and re-PUT all the images after this is deployed to prod.

## Tests and linting
 
 [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -4,7 +4,7 @@ import hashlib
 import random
 from sqlalchemy import func
 from sqlalchemy.sql.expression import cast
-
+import imghdr as imghdr
 from flask import current_app, url_for
 
 from .models import db, Officer, Assignment, Image, Face, User, Unit, Department
@@ -120,11 +120,12 @@ def upload_file(safe_local_path, src_filename, dest_filename):
     # Folder to store files in on S3 is first two chars of dest_filename
     s3_folder = dest_filename[0:2]
     s3_filename = dest_filename[2:]
+    s3_content_type = "image/%s" % imghdr.what(safe_local_path)
     s3_path = '{}/{}'.format(s3_folder, s3_filename)
     s3_client.upload_file(safe_local_path,
                           current_app.config['S3_BUCKET_NAME'],
                           s3_path,
-                          ExtraArgs={'ACL': 'public-read'})
+                          ExtraArgs={'ContentType': s3_content_type, 'ACL': 'public-read'})
 
     url = "https://s3-{}.amazonaws.com/{}/{}".format(
         current_app.config['AWS_DEFAULT_REGION'],
--- a/OpenOversight/tests/test_utils.py
+++ b/OpenOversight/tests/test_utils.py
@@ -90,7 +90,7 @@ def test_compute_hash(mockdata):
     assert hash_result == expected_hash
 
 
-def test_s3_url(mockdata):
+def test_s3_upload_png(mockdata):
     local_path = 'OpenOversight/app/static/images/test_cop1.png'
 
     mocked_connection = Mock()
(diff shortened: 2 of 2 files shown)


OpenOversight/app/utils.py
```diff
@@ -114,17 +114,32 @@ def compute_hash(data_to_hash):
     return hashlib.sha256(data_to_hash).hexdigest()
 
 
+# Map file extension to content type naively
+def guess_mime_type(filename):
```
bmintz: As long as we won't be testing images, we can use a function in the standard library called [`imghdr.what`](https://docs.python.org/3/library/imghdr.html#imghdr.what). It returns a value based on the _contents_ of the file, rather than its extension.

Although I would advise against it, you can still guess based on file extension using https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_extension instead.

OpenOversight/app/utils.py
```diff
@@ -114,17 +114,32 @@ def compute_hash(data_to_hash):
     return hashlib.sha256(data_to_hash).hexdigest()
 
 
+# Map file extension to content type naively
```
bmintz: Should be a docstring.

r4v5:
Thanks for the pointer to imghdr.what()!

bmintz:
No problem! I thought we were gonna use it though...

r4v5:
We did: https://github.com/lucyparsons/OpenOversight/pull/354/files#diff-3110b4868e45e11a2432598cb73a53c7R123

PR #366, opened 2018-02-16
Add per-user default department preference

### LINKED ISSUE #358: User should be able to select a default department
As an OpenOversight user interested only in one department, I want the website to remember the department I care about without having to select it every time.

pierwill:
## Status

Ready for review

## Description of Changes

Updates forms, views, and templates to include the default department feature. If the user chooses a preference, the **Submit Images** button takes them right to the appropriate page. When setting this preference, the user can choose the blank option for no preference.

## Description of Changes

Fixes #358.

## Tests and linting
 
 [x] `flake8` checks pass


OpenOversight/app/main/views.py
```diff
@@ -432,8 +432,12 @@ def submit_complaint():
 @main.route('/submit', methods=['GET', 'POST'])
 @limiter.limit('5/minute')
 def submit_data():
-    departments = Department.query.all()
-    return render_template('submit_deptselect.html', departments=departments)
+    if User.dept_pref is not None:
```
pierwill: I see that my guess on how to get data from the db was wrong. Went looking for something similar. Maybe `if User.query.filter_by(dept_pref=field.data).first():`?

b-meson:
This PR works well functionally ! Do we need to write tests?

redshiftzero:
Thanks for the PR @pierwill - looks great. Two things would make this :100: - unit tests and a schema migration (for the new column in the database)

pierwill:
I've added a unit test here.

pierwill:
@redshiftzero @r4v5 Some of us are hitting a wall trying to figure out how to do database migrations..

redshiftzero:
Thanks for the ping @pierwill, will take a looky see

redshiftzero:
Hey @pierwill, sorry for delay, have taken a look at this - the tl;dr is that I messed up the migrations in the development environment. To step around the issue you are seeing, with the virtualenv activated in vagrant:

```
python manage.py db stamp head
```

Then:

```
python manage.py db migrate
```

This should generate a migration like [this commit here](https://github.com/lucyparsons/OpenOversight/commit/6f5f73af234a7eeac5f6987de7596ddfcbf653df). 

You can test the migration by applying it to the development database via:

```
python manage.py db upgrade
```

I'm going to make some changes in the dev env(s)/docs but this should get you cookin'.

redshiftzero:
And @magpiewings you should be able to do the same thing over in #390 (obviously the migration itself produced will be different). Let me know if y'all have additional trouble. I can also do the database migration parts of the PRs if either of you prefer, just let me know :heart:

OpenOversight/app/main/views.py
```diff
@@ -433,8 +438,18 @@ def submit_complaint():
 @main.route('/submit', methods=['GET', 'POST'])
 @limiter.limit('5/minute')
 def submit_data():
-    departments = Department.query.all()
-    return render_template('submit_deptselect.html', departments=departments)
+    # try to use preferred department if available
+    try:
+        if User.query.filter_by(id=current_user.id).one().dept_pref:
+            department = User.query.filter_by(id=current_user.id).one().dept_pref
+            return redirect(url_for('main.submit_department_images', department_id=department))
```
redshiftzero: I think it's OK to do this but for logged-in users that might want to submit images for a department that is not their own, there should be a note on the submit page only displayed for users that have a default department that says "Want to submit for another department? Change your default department" where "default department" links to the change department form. What do you think?

OpenOversight/tests/test_routes.py
```diff
@@ -870,3 +870,18 @@ def test_admin_can_add_new_unit(mockdata, client, session):
...
+
+        form = ChangeDefaultDepartmentForm(dept_pref=1)
+
+        rv = client.post(
+            url_for('auth.change_dept'),
+            data=form.data,
+            follow_redirects=True
+        )
+
+        assert 'Updated!' in rv.data
```
redshiftzero: Nice test! One though - we could also assert that the correct department is assigned - i.e.:

```
        user = User.query.filter_by(email='jen@example.org').one()
        assert user.dept_pref == test_department_id
```

where `test_department_id` is what you set the department ID to be, i.e. in the example 1.

OpenOversight/migrations/versions/af933dc1ef93_.py
```diff
@@ -0,0 +1,30 @@
...
+    op.add_column('users', sa.Column('dept_pref', sa.Integer(), nullable=True))
+    op.create_foreign_key(None, 'users', 'departments', ['dept_pref'], ['id'])
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, 'users', type_='foreignkey')
+    op.drop_column('users', 'dept_pref')
+    # ### end Alembic commands ###
```
redshiftzero: :metal:

OpenOversight/app/main/views.py
pierwill: how about a `flash()` for this?

pierwill: ✅

OpenOversight/app/main/views.py
```diff
@@ -433,8 +438,19 @@ def submit_complaint():
 @main.route('/submit', methods=['GET', 'POST'])
 @limiter.limit('5/minute')
 def submit_data():
-    departments = Department.query.all()
-    return render_template('submit_deptselect.html', departments=departments)
+    # try to use preferred department if available
+    try:
+        if User.query.filter_by(id=current_user.id).one().dept_pref:
+            flash(Markup('Want to submit for another department? Change your <a href="/auth/change-dept/">default department.</a>'))
```
pierwill: Regarding the period and link, do we like

`flash(Markup('Want to submit for another department? Change your <a href="/auth/change-dept/">default department.</a>'))`

or

`flash(Markup('Want to submit for another department? Change your <a href="/auth/change-dept/">default department</a>.'))`

better?

pierwill: Also: is there a way to dismiss the flash after a few seconds? Something undocumented in flask?

redshiftzero: I like `flash(Markup('Want to submit for another department? Change your <a href="/auth/change-dept/">default department</a>.'))` better but I don't have strong feelings.

So I'm not aware of a feature to auto-dismiss the flashed message in Flask at least, but [this](https://pythonprogramming.net/flash-flask-tutorial/) looks like a pretty decent tutorial to make the alert dismissable.

redshiftzero: Just kidding, the alerts are already dismissable ;) Thanks Bootstrap.

pierwill: 👍

pierwill:
🎉💃

PR #392, opened 2018-02-18
manage.py: add bulk upload command

### LINKED ISSUE #247: Management command to add officer rosters
So as people try to spin up OpenOversight in their own city, or as we want to update our own roster, we want some way of adding officers in bulk from files we get through FOIA. We have implemented issue #214 and have yet to be implemented issue #213 as stopgaps, but we would like to be able to do this in bulk.

tmc:
## Status

Ready for review

## Description of Changes

Fixes #247

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -226,3 +226,11 @@ def compute_leaderboard_stats(select_top=25):
                             .order_by(func.count(Face.user_id).desc()) \
                             .limit(select_top).all()
     return top_sorters, top_taggers
+
+
+def badge_exists(star_no, department_id):
+    """badge_exists checks if a given badge number exists already for a department."""
+    for assignment in Assignment.query.filter_by(star_no=star_no).all():
+        if str(assignment.baseofficer.department_id) == str(department_id):
+            return True
+        return False
--- a/OpenOversight/manage.py
+++ b/OpenOversight/manage.py
@@ -1,11 +1,13 @@
 from getpass import getpass
 import sys
+import csv
 
 from flask_script import Manager, Server, Shell
 from flask_migrate import Migrate, MigrateCommand
 
 from app import app
-from app.models import db, User
+from app.models import db, Assignment, Department, Officer, User
+from app.utils import badge_exists
 
 
 migrate = Migrate(app, db)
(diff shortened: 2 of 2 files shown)


OpenOversight/app/utils.py
```diff
@@ -226,3 +226,10 @@ def compute_leaderboard_stats(select_top=25):
                             .order_by(func.count(Face.user_id).desc()) \
                             .limit(select_top).all()
     return top_sorters, top_taggers
+
```
r4v5: `flake8` is complaining about only having a single newline between this function and the previous one.

OpenOversight/manage.py
```diff
@@ -90,5 +92,49 @@ def link_officers_to_department():
...
+        officer.department_id = department_id
+        officer.last_name = line['last_name']
+        officer.first_name = line['first_name']
+        officer.middle_initial = line['middle_initial']
+        officer.race = line['race']
+        officer.gender = line['gender']
+        officer.employment_date = line['employment_date']
+        officer.birth_year = line['birth_year']
+        db.session.add(officer)
+        db.session.commit()
```
r4v5: I believe this `commit()` will throw constraint-violation-related exceptions if the officer already exists; we might want to wrap the commit in a `try` and log the failed officer on exception? This would also have the advantage of making reimports closer to idempotent.

OpenOversight/app/utils.py
```diff
@@ -226,3 +226,11 @@ def compute_leaderboard_stats(select_top=25):
...
+
+def badge_exists(star_no, department_id):
+    """badge_exists checks if a given badge number exists already for a department."""
+    for assignment in Assignment.query.filter_by(star_no=star_no).all():
+        if str(assignment.baseofficer.department_id) == str(department_id):
+            return True
+        return False
```
redshiftzero: So here's a tricky thing about some police departments we have in OpenOversight: sometimes a badge number may appear several times - e.g. for Chicago Police Department, they allocate an officer a new badge number upon promotion and then re-use the old badge number for a different officer as it has been freed. Since we track the assignments history of officers, the same badge number might appear multiple times in the database if it was indeed allocated to another officer (yep this is silly but it happens 🤷‍♀️).

A more robust approach would be to have this function be `officer_exists` and check that no row exists with the same badge number, first name, or last name. 

Also apologies for the long overdue review. I am happy to take over this PR if you have other stuff you want to work on now 😇 :heart:

dismantl:
I'd be happy to make the requested changes to this PR, but I don't have access to the origin repo.

abandoned-prototype:
@dismantl It seems the problem here is that the original repo was deleted. Googling for this issue lead me to find that you can check out the PR as a branch (I called it pr-392) in your local copy with ` git fetch upstream pull/392/head:pr-392`

redshiftzero:
it'd would be great if you'd want to take this over @dismantl, feel free to pull these changes into a new superseding PR (since the original remote has been deleted)

redshiftzero:
Closing in favor of #524, thanks @tmc :)

PR #396, opened 2018-02-18
Document the development env deficiencies

redshiftzero:
Some people were confused that the photo upload doesn't work, so this PR is just documenting somewhere that it isn't working


redshiftzero:
thanks @r4v5 🥇

Issue #171, opened 2017-01-24
Mock S3 with folder in dev/testing


redshiftzero:
In PR #169 I added this `serve_image` function in `utils.py` that generates a path for the files stored locally in the `static/` folder. However in prod and staging images are actually stored on S3. It'd be nice to [better mock the S3 setup](https://github.com/tornadoweb/tornado/blob/master/demos/s3server/s3server.py) in dev/testing to be as close as possible to prod/staging (suggested by @r4v5).

raq929:
It turns out that I didn't need this for the other issue, but for future reference, boto 1.10 has it's own nice [stubber for s3](http://botocore.readthedocs.io/en/latest/reference/stubber.html).

fritzdavenport:
Finishing this puppy up, encountering an unexpected error in `test_functional` but all other tests and functionality passes. Expect PR soon

fritzdavenport:
Finished - tests pass locally but die on Travis. Struggling with getting much info for it :/

abandoned-prototype:
Hi @fritzdavenport. Thanks for your work on this!
I am pretty excited about having this for local development. I just checked out your branch and ran `make test` and got the exact same errors that travis is getting `16 passed, 330 error`. So maybe you are able to recreate those failures by throwing some docker images away or something

PR #404, opened 2018-03-19
converts tiff and psd photo submissions to jpeg

### LINKED ISSUE #369: Allow upload of additional image formats
We should accept tiff and/or psd formats in the photo upload and display them or convert them to a format we can display in the web application. 

### User Stories

As an OpenOversight volunteer with a collection of photographs, I want to simply upload them in the existing format.

magpiewings:
## Status

this code should work, but cannot test because I can't set up a mock S3 bucket (open issue #171)

## Description of Changes

Fixes #369 .

 - Converts tiff and psd file submissions to jpegs

## Tests and linting
 
 [ ] pytests pass in the development environment on my local machine
 
 [ ] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/config.py
+++ b/OpenOversight/app/config.py
@@ -37,6 +37,7 @@ class BaseConfig(object):
     # Upload Settings
     MAX_CONTENT_LENGTH = 50 * 1024 * 1024
     ALLOWED_EXTENSIONS = set(['png', 'jpg', 'jpeg', 'gif'])
+    REQUIRES_CONV = set(['tiff', 'psd'])
 
     SEED = 666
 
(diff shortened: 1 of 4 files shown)


b-meson:
@magpiewings and I are taking a look at the tests and I think adding a new key to `config.py` is not the right way of doing this. thinking of a line like this ` if allowed_file.extension == '.psd' or allowed_filed.extension == 'tiff': on line 481 in `views.py` is a better way of doing this.

magpiewings:
@b-meson is trying to do something similar in the same place, stripping metadata from jpegs, so we will need to discuss the best way to check file extensions

PR #434, opened 2018-05-04
Changed star_date to start_date

### LINKED ISSUE #427: Change Assignment `star_date` to `start_date`
Make sure to find it everywhere

kurianbenoy:
## Status

## Description of Changes

Fixes #427 

 - Change star_date to start_date

 [x] `flake8` checks pass


kurianbenoy:
I have made the requested changes . Please review the PR 

I found that flake8 failed by default migration files which included 2 import statements ie:
```
from alembic import op

import sqlalchemy as sa 
```
Yet these import statements are present in almost all the migration files. So please do review is it necessary or not . I have removed both the import statements to pass tests

kurianbenoy:
https://github.com/lucyparsons/OpenOversight/tree/develop/OpenOversight/migrations/versions

This is where all migrations are present and all have 2 import statements. So do review if it's necessary or not?
@raq929  Maybe flake8 configuration might be the reason for failing Tests in Travis.yml

OpenOversight/migrations/versions/19f23f6eb170_.py
```diff
@@ -0,0 +1,21 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '19f23f6eb170'
+down_revision = 'af933dc1ef93'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    pass
```
redshiftzero: As you can see, your `upgrade()` and `downgrade()` does nothing here, because Alembic [does not autogenerate migrations for column renames](http://alembic.zzzcomputing.com/en/latest/autogenerate.html#what-does-autogenerate-detect-and-what-does-it-not-detect). Unfortunately, that means that in order to merge the changes in this PR, we need a manually written migration - note that we cannot simply drop the old column and add the new column, as then we'd lose all the data in the column.

kurianbenoy:
```
def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.alter_column('assignments', 'start_no',
                    existing_type=sa.INTEGER(),
                    type_=sa.String(length=120),
                    existing_nullable=True)
    # ### end Alembic commands ###
```

```
def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
    op.alter_column('assignments', 'start_no',
                    existing_type=sa.String(length=120),
                    type_=sa.INTEGER(),
                    existing_nullable=True)
```

So is this manually written migration enough. Is this what you meant? @redshiftzero @raq929

kurianbenoy:
Thanks for spotting the error ☺. I followed your suggestions and made requested corrections.
Can you check if it's correct ?

PR #447, opened 2018-05-21
Filter by race, gender: Return officers that we do not have race, gender info for

### LINKED ISSUE #425: Officer find: For each filter condition, return "Not sure" officers along with the matching filter
Disable search by gender, race, age for departments where we don't have that info (e.g. Berkeley)

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #425

 - For some departments, we do not have per-officer race or gender information. In those cases, we do not currently return them if a user searches for a specific gender. This is confusing behavior - for a query for all male officers, we should return all male officers and those we are not sure about. Longer term, area coordinators should be maintaining this information  if possible.

## Notes for Deployment

No db migrations

## Tests and linting
 
- [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -167,11 +167,15 @@ def filter_by_form(form, officer_query):
         )
     if form['race'] in ('BLACK', 'WHITE', 'ASIAN', 'HISPANIC',
                         'PACIFIC ISLANDER'):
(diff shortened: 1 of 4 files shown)


redshiftzero:
Deployed this one to staging, working as expected, so merging

Issue #173, opened 2017-01-24
Show existing tags overlaid on each image


redshiftzero:
This would be nice in two places:

* on the `/image` route
* as the volunteer is tagging an image on `/cop_face`, boxes could appear around the faces of officers as they are being tagged. This way if there are many many officers in an image and I have extremely poor memory then I would be able to see which officers I have already tagged.

PR #169, opened 2017-01-24
Adds volunteer image classification tool, volunteer officer identification tool, admin/moderation capability, leaderboard, and user profiles

redshiftzero:
This closes #13 (implement tagging functionality for users), closes #14 (implements `face_position` storage in the database), closes #159 (implement user profiles), and closes #158 (implement a tagging leaderboard). Here is a rundown of the changes:

## `/user`: View user profile page

Each user now has a "Profile" link in the upper right hand of the navigation bar, that takes them to an overview of their OpenOversight activity:

screen shot 2017-01-24 at 12 38 04 am

Users can also look at one another's profiles. When an administrator is moderating, this is a good place to start: check for fradulent classifications / tags by clicking on a user's page and looking at a few of their classifications / tags in order to remove fradulent ones. An administrator can click the "Disable account" to withdraw a user's ability to classify images and add tags. They should only do this if they believe the account is fradulent. The "Admin only" panel only appears if the user is logged in and is an administrator. 

## `/image`: Show image and moderate classifications

The new `/image` route enables a logged in user to view the metadata associated with a particular image, as well as the classification of the image and who it was classified by. An administrator has the option here to change the classification if they decide it is incorrect (again, the "Admin only" panel only appears if the user is logged in and is an administrator):

screen shot 2017-01-24 at 12 38 10 am

Classifying the image as being without police officers will remove it from circulation - it won't be displayed to volunteers to classify or tag. 

## `/tag`: Show tag and moderate tags

The new `/tag` route enables a logged in user to view the officer ID, the image and who it was added by. An administrator has the option here to remove the tag if they decide it is incorrect (again, the "Admin only" panel only appears if the user is logged in and is an administrator):

screen shot 2017-01-24 at 12 38 20 am

## `/leaderboard`: Show top volunteers

This adds a leaderboard:

screen shot 2017-01-24 at 12 38 29 am

Admins can use this to quickly find new active users and can click on the profile to inspect their activity in detail and verify that it is not fradulent. 

## Volunteer Application

Volunteers without administrator privileges will use the new classification and tagging feature. When a user clicks "Volunteer" (formerly "Getting Started"), they will be taken to a page instructing them to login or register:

screen shot 2017-01-24 at 12 35 58 am

Upon login or registering and reloading this page, they will be provided with two options:

screen shot 2017-01-24 at 12 36 16 am

### `/sort` Volunteer image classification task

This task has volunteers sort through submitted (or scraped) images in the database and classify them as being with or without cops:

[...]

PR #459, opened 2018-06-04
236 create thumbnails

### LINKED ISSUE #236: Make officer face cutouts a uniform, responsive size
We started displaying just the cutout of the face in officer search results. However there is a wide range of sizes in the cutouts. We make the cutouts using a little CSS (Jinja templating is in here too):

```
background-image: url({{ officer_image }});
background-repeat: no-repeat;
background-position: -{{ officer.face.first().face_position_x }}px -{{ officer.face.first().face_position_y }}px;
width: {{ officer.face.first().face_width }}px;
height: {{ officer.face.first().face_height }}px;
```

We need these to be a uniform, responsive size. Suggestions?

raq929:
## Status

Ready for review

## Description of Changes

Fixes #236 .

 - Instead of saving image crop data, this PR actually crops the image and uploads a cropped image to s3
 - A temporary migration takes all existing faces and changes them to cropped images
 - the new Faces created keep a FK to the original Image

## Notes for Deployment

**IMPORTANT** 
Immediately after migrations are run, the migration that changes existing faces to images should be removed. (The following migration that removes columns from the face table will need to be updated as well at that time.)

screen shot 2018-06-04 at 4 02 53 pm

## Tests and linting
 
 [ x] I have rebased my changes on current `develop`
 
 [ x] pytests pass in the development environment on my local machine
 
 [ x] `flake8` checks pass


OpenOversight/app/models.py
```diff
@@ -102,10 +102,6 @@ class Face(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     officer_id = db.Column(db.Integer, db.ForeignKey('officers.id'))
     img_id = db.Column(db.Integer, db.ForeignKey('raw_images.id'))
-    face_position_x = db.Column(db.Integer, unique=False)
-    face_position_y = db.Column(db.Integer, unique=False)
-    face_width = db.Column(db.Integer, unique=False)
-    face_height = db.Column(db.Integer, unique=False)
```
redshiftzero: hm, so if we ever want to do e.g. #173, it's convenient to have these boxes in the database

OpenOversight/migrations/versions/59e9993c169c_change_faces_to_thumbnails.py
```diff
@@ -0,0 +1,44 @@
...
+revision = '59e9993c169c'
+down_revision = 'bd0398fe4aab'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    for face in Face.query.all():
+        # import pdb; pdb.set_trace()
+        if face.face_position_x \
```
redshiftzero: so I noticed while testing this migration in the dev env that the test data does not have the bounding box fields filled in. since those columns haven't actually been required (this was a holdover from the early days when we manually made cutouts of pics before submitting them to the website), it's probably worth doing:

```
try:
    if face.face_position_x \ 
    [...]
except AttributeError:
    pass  # then skip this face
```

OpenOversight/tests/test_utils.py
```diff
@@ -135,3 +136,39 @@ def test_user_cannot_submit_invalid_file_extension(mockdata):
 def test_unit_choices(mockdata):
     unit_choices = [str(x) for x in OpenOversight.app.utils.unit_choices()]
     assert 'Unit: Bureau of Organized Crime' in unit_choices
+
+
+# Mock calls to upload_file
+@patch('OpenOversight.app.utils.upload_file', MagicMock(return_value='https://s3-some-bucket/someaddress.jpg'))
```
redshiftzero: ah.. that works too :-)

raq929:
OK, this should be good to go. I removed the migration that took away the columns, and added a column to keep track of the original image, since you would also need that data for #173. 
(You'll have to downgrade your db before you pull and then upgrade again afterwards, since I also changed it so that the original image id column is added __before__ the temp migration takes place.)
Also, fyi the creating new faces migration may take a while to run on staging/production, as it involves a fair amount of writing to disk. If we think it will be too much, I can investigate my TODO about using a BytesIO object.

raq929:
@redshiftzero can you be sure to check the migration before pushing to staging? I'm worried about it dropping FKs when it adds the column.

redshiftzero:
Noted! and yeah at this stage in the project it's fine if the migration takes a while, I'll just do it on off hours 🌚

OpenOversight/migrations/versions/4a490771dda1_add_original_image_fk.py
```diff
@@ -0,0 +1,34 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('faces', sa.Column('fk_face_original_image_id', sa.Integer(), nullable=True))
+    op.drop_constraint(u'faces_img_id_fkey', 'faces', type_='foreignkey')
+    op.create_foreign_key('fk_face_image_id', 'faces', 'raw_images', ['img_id'], ['id'], onupdate='CASCADE', ondelete='CASCADE', use_alter=True)
+    op.create_foreign_key(None, 'faces', 'raw_images', ['fk_face_original_image_id'], ['id'], onupdate='CASCADE', ondelete='SET NULL', use_alter=True)
+    # ### end Alembic commands ###
```
redshiftzero: Verified tags (mapping of face -> officer) are indeed preserved after this migration ✅

redshiftzero:
I rebased this one on top of the migration added in #475

Doing a wee bit more investigation - it looks like the reason this migration was hanging for me locally was due to the database being locked:

```
openoversight-dev=> SELECT relation::regclass, * FROM pg_locks WHERE NOT GRANTED;
  relation  | locktype | database | relation | page | tuple | virtualxid | transactionid | classid | objid | objsubid | virtualtransaction |  pid  |      mode       | granted | fastpath 
------------+----------+----------+----------+------+-------+------------+---------------+---------+-------+----------+--------------------+-------+-----------------+---------+----------
 raw_images | relation |    16385 |    16574 |      |       |            |               |         |       |          | 4/14               | 15990 | AccessShareLock | f       | f
 faces      | relation |    16385 |    16654 |      |       |            |               |         |       |          | 3/26               | 15951 | AccessShareLock | f       | f
(2 rows)
```

redshiftzero:
Progress, I deployed this to staging by running this migrations separately (as I did _not_ get a db lock when doing this in the dev environment.. 🤷‍♀️ ):

```
python manage.py upgrade db 4a490771dda1
python manage.py upgrade db 59e9993c169c
```

I added one wee change, and then I think this is good to deploy. 

The plan will be to:

1. Backup everything
2. Run each migration separately as above
3. Manually re-tag any images that were added prior to the use of `face_position_x` (#169) - it looks like there are a few images in this situation, but there shouldn't be many

PR #481, opened 2018-06-27
Include officers without assignments in search results

### LINKED ISSUE #440: Search function
Right now in the `find_by_form` function in `utils.py`, line 160:
`officer_query = officer_query.join(Assignment)`
is filtering out any officers that don't have assignments. I think it should probably have an or Officer.assignment == none attached, so that it catches officers that don't have assignment data. (I'm not sure if this is a problem IRL, but it is with the test data).

raq929:
## Status

## Description of Changes

Fixes #440 .

 - use a left join on Assignment
 - Added a test to be sure it worked

## Tests and linting
 
 [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -228,7 +228,7 @@ def filter_roster(form, officer_query):
             Officer.last_name.ilike('%%{}%%'.format(form['name']))
         )
 
-    officer_query = officer_query.join(Assignment)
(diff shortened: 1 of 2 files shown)


raq929:
Note the second commit which makes tests pass. An alternate behavior here would be to allow officers without assignments to be in search results, which would require changing the tests. It's kind of a judgement call, so let me know if you want it changed. I feel like what it is right now is the behavior I'd expect.

PR #490, opened 2018-07-02
Add frontend asset pipeline for SCSS 

### LINKED ISSUE #471: Use SASS
For front-end developer merriment, let's start using SASS:

https://getbootstrap.com/docs/4.0/getting-started/theming/

We need to:
- [ ] Install necessary dependencies in dev env (sass ruby gem)
- [ ] Integrate Flask-Assets and pyScss into the Flask application

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #471

 - This uses `node-sass` to bundle up SCSS
 - Swaps out Bootstrap 3 CSS for Bootstrap 3 SCSS

## Testing

`make dev`

After a while, the webapp should be available on 127.0.0.1:3000

Check watching: i.e. edits to OpenOversight/app/static/scss/index.scss (e.g. to change the background color) should not require a rebuild of the container

## Tests and linting

 - [x] `flake8` checks pass


redshiftzero:
So the remaining thing to do here is to get this to live reload. Note that `node-sass` is running in watch mode in the container so the `css` will continually get rewritten, but it does require a browser refresh to show up, which is kind of annoying. That said, this is otherwise working.

I decided NOT to migrate from bootstrap 3 to 4 because it ended up actually being a ton of work. 

But this PR does remove the entire `bootstrap.css` theme and imports Bootstrap 3 in SCSS form. 

Frontend edits should now take place in `static/scss/index.scss`, which has the bootstrap 3 SCSS imported and some variables already overridden. Code in the now-legacy openoversight.css we can slowly migrate to SCSS as time allows/as it is edited.

nmorduch:
I've actually never worked in a dev environment with live reload of any sort, so I think I at least will survive :) The question to me is how long the recompiling takes; I was just working on something where it took _over two minutes_ which was long enough to mess with my productivity.

redshiftzero:
Cool cool. Well, I'll defer live reload to a followup for now, but let me know if it gets annoying and I'll spend some more time on this

Otherwise I think this PR is aighhhhht. Since I'm removing a BIG CSS file that contains (mostly) cruft it also removes the custom fonts including that beautiful Raleway font (:innocent:) in prep for branding updates in #494

nmorduch:
screenshot_2018-07-27 openoversight - a lucy parsons labs project

one small (and hacky tbh) change, but this is working well for me!

redshiftzero:
ok I've rebased this on current and I thiiiink this is good to go - any object with my removing the `vagrant` environment at this stage @r4v5 @dismantl?

OpenOversight/scripts/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [ "${DOCKER_BUILD_ENV:-}" == "production" ]; then
+    gunicorn -w 4 -b 0.0.0.0:3000 app:app
+else
+    yarn build
+    yarn watch &
+    python manage.py runserver
```
dismantl: What's the reason to call `manage.py runserver` here instead of the gunicorn command? Because of [this change](https://github.com/lucyparsons/OpenOversight/pull/523#issuecomment-409296717), running without `DOCKER_BUILD_ENV=production` causes gunicorn to be run on port 5000 rather than 3000.

redshiftzero: `runserver` is here such that people can get tracebacks and use the Flask in-browser debugger when in the dev env (right now this is not the case which is not ideal)

dismantl: Then would it make sense to add the runserver command back into manage.py, specifying port 3000?

dismantl:
Rebased onto most resent develop branch. A quick browse of the site looks fine and changes to index.scss are updated as specified.

Makefile
```diff
@@ -16,7 +16,11 @@ create_db: start
 	done
 	@echo "Postgres is up"
 	## Creating database
-	docker-compose run --rm web /usr/local/bin/python ../create_db.py
+	docker-compose run --rm web python ../create_db.py
+
+.PHONY: assets
+assets:
+	yarn build
```
dismantl: I think this may need to be run in the docker container

OpenOversight/app/templates/base.html
```diff
@@ -10,66 +10,58 @@
...
 
-    <!-- jQuery -->
-    <script src="{{ url_for('static', filename='js/jquery.min.js') }}"></script>
-
     <!-- Font Awesome -->
     <link href="{{ url_for('static', filename='css/font-awesome.min.css') }}" rel="stylesheet">
 
-    {% for item in jsloads %}
-    <script src="{{ url_for('static', filename=item) }}"></script>
-    {% endfor %}
```
dismantl: I'm wondering if this may break some routes that use `jsloads`

docker-compose.yml
```diff
@@ -25,12 +29,15 @@ services:
      S3_BUCKET_NAME: "${S3_BUCKET_NAME}"
      FLASK_APP: app
      FLASK_ENV: "${FLASK_ENV:-development}"
+     HMR: 1
    volumes:
      - ./OpenOversight/:/usr/src/app/OpenOversight/:z
+     - cache_dir:/usr/src/app/.cache
+     - /usr/src/app/OpenOversight/static/dist/
```
dismantl: I hope I merged the docker-compose conflicts correctly, but it would be good to check my work here

OpenOversight/app/templates/base.html
dismantl: confirmed

dismantl:
Ok fixed some issues and it's ready for review now.

redshiftzero:
thanks @dismantl!

PR #506, opened 2018-07-10
371 ac officer creation

### LINKED ISSUE #371: Add photo submission page for logged in users
Some users of OpenOversight have detailed information about officers in their community, and want to upload this information, along with the photographs of the officer, all in one page. These photos will not need to go through the usual volunteer workflow as their identities are already known. Only accounts associated with a known user from the community should have this capability (e.g. #355) - administrators should determine this. 

On this submit page we should be able to submit all information about an officer:
* Photos
* Name
* Ethnicity
* Gender
* Links to lawsuits, videos, media articles (if #223 is implemented)
* Free text - for now we should just store this in the database - though we can display for coordinators/administrators on their officer profile

### User Stories

As a community member with knowledge of the police force, I should be able to submit photographs of officers I know along with all information about the officer in a single page.

raq929:
## Status

Ready for review

## Description of Changes

Fixes #371 .

 - Allows admins and ACs to add officer photos directly via the officer profile page
 - Refactors photo uploads for ease of testing

**NOTES FOR REVIEW**
Since this uses the same route as other photo uploads, double check that all photo uploads are working properly.
Since the function used for photo upload is a modified version of the one used to tag officers, double check that tagging officers is working correctly.  

Officer profile:
screen shot 2018-07-09 at 10 54 25 pm

Upload page:
screen shot 2018-07-09 at 10 55 36 pm

## Tests and linting

 - [x] `flake8` checks pass


McEileen:
I'm going to work on this during the hackathon.  I'll start with fixing the merge conflicts.

camfassett:
Hey @McEileen! How did working on this go at the Hackathon?

McEileen:
Hey, I wanted to comment that I'm aware of the breaking tests and I will ping for a review after fixing them.  Onward!

McEileen:
Status update: One failing test left.  I will ping for a review after fixing the relevant test.

McEileen:
In case it's not clear, I wanted to highlight that after all of the struggles with race conditions in testing, this PR is ready for review.  Thanks, @dismantl, for all your help!

dismantl:
When manually testing this in a docker environment, I got the following error when trying to upload a photo for a particular officer:

```
(OpenOversight) armitage:OpenOversight dismantl$ docker logs -f openoversight_web_1
 * Serving Flask app "app" (lazy loading)
 * Environment: development
 * Debug mode: on
 * Running on http://0.0.0.0:3000/ (Press CTRL+C to quit)
 * Restarting with inotify reloader
[2019-03-27 23:25:09,554] INFO in __init__: OpenOversight startup
 * Debugger is active!
 * Debugger PIN: 485-533-843
[2019-03-27 23:25:10,485] INFO in __init__: OpenOversight startup
172.20.0.1 - - [27/Mar/2019 23:25:14] "GET /submit_officer_images/officer/160 HTTP/1.1" 200 -
[2019-03-27 23:25:22,280] ERROR in utils: Error uploading to S3: <class 'AttributeError'> '_io.BytesIO' object has no attribute 'department_id' Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/werkzeug/datastructures.py", line 2821, in __getattr__
    return getattr(self.stream, name)
AttributeError: 'SpooledTemporaryFile' object has no attribute 'department_id'

During handling of the above exception, another exception occurred:

  File "/usr/src/app/OpenOversight/app/utils.py", line 482, in get_uploaded_image
    department_id=image.department_id,
  File "/usr/local/lib/python3.7/site-packages/werkzeug/datastructures.py", line 2827, in __getattr__
    return getattr(self.stream._file, name)
AttributeError: '_io.BytesIO' object has no attribute 'department_id'

172.20.0.1 - - [27/Mar/2019 23:25:22] "POST /upload/department/1/officer/160 HTTP/1.1" 500 -
```

I have valid AWS credentials and S3 bucket in my `.env` file, so I believe that's not the issue. Any ideas what's causing that?

OpenOversight/app/main/views.py
```diff
@@ -686,62 +684,41 @@ def all_data():
...
-        db.session.add(new_image)
-        db.session.commit()
-        return jsonify(success="Success!"), 200
-    except:  # noqa
-        exception_type, value, full_tback = sys.exc_info()
-        current_app.logger.error('Error uploading to S3: {}'.format(
-            ' '.join([str(exception_type), str(value),
-                      format_exc()])
-        ))
+    image = get_uploaded_image(image=file_to_upload, department_id=department_id)
```
dismantl: `get_uploaded_image()` expects an Image object, but here it's being passed `file_to_uploaded`, which is a `werkzeug.datastructures.FileStorage` object. I think this is what caused the error I ran into, since the passed image doesn't have a `department_id` or `date_image_taken` property (utils.py:482).

OpenOversight/app/utils.py
```diff
@@ -487,6 +506,16 @@ def rm_dirs():
         return None
 
 
+def find_date_taken(pimage):
+    if pimage.filename.split('.')[-1] == 'png':
+        return None
+    if pimage._getexif():
+        if 36867 in pimage._getexif():
```
dismantl: Could you add a comment here to note where this number comes from?

OpenOversight/app/utils.py
```diff
@@ -460,20 +478,21 @@ def rm_dirs():
...
     # Upload file from local filesystem to S3 bucket and delete locally
     try:
-        url = upload_file(safe_local_path, original_filename,
+        url = upload_file(safe_local_path, filename,
                           new_filename)
         rm_dirs()
         # Update the database to add the image
+        # If we're cropping an image, get data from the original image,
+        # else, use data given to us.
         new_image = Image(filepath=url, hash_img=hash_img, is_tagged=True,
```
dismantl: There's a bug here that when I'm uploading photos from the main Submit Images page (not from officer profile), this causes the images to be marked as tagged when that isn't the case. They subsequently don't show up in the tagging queue.

OpenOversight/app/utils.py
```diff
@@ -425,31 +429,45 @@ def create_description(self, form):
         date_updated=datetime.datetime.now())
 
 
-def get_uploaded_cropped_image(original_image, crop_data):
+def get_uploaded_image(image, crop_data=None, department_id=None):
```
dismantl: I think the `get_uploaded_image()` function is a bit confusing as it is, mostly because it attempts to accomplish too many tasks and operates differently depending on which view function it was called from. I would suggest refactoring it out into two different functions:

`upload_image_to_s3(department_id, image_data<io.BytesIO>):`
1. hash `image_data`. if hash found in DB, return existing Image
2. determine image format using `imghdr.what(h=image_data)`; only continue if format recognized by imghdr
3. determine date taken if possible
4. upload to S3 and return new Image instance (don't set `is_tagged`)

`crop_image(image<Image>, crop_data=None):`
1. read in raw image data via `image.filepath` (url or local file) and create a BytesIO buffer from it
2. determine image type using imghdr again, reject if unrecognized
3. crop image: `cropped_image = Pimage.open(image_data).crop(crop_data)` (or `.thumbnail((300,300))` if `crop_data==None`)
4. Save the cropped Pimage into a new BytesIO buffer via `Pimage.save()` using the image format determined from imghdr
5. `return upload_image_to_s3(department_id, cropped_image_buffer)`

A couple things to note: 1) Neither function sets the `is_tagged` field on the resulting Image; that's the responsibility of the caller/view function. 2) The uploaded image (original or cropped) is never saved to disk during this process; it's passed around as a BytesIO stream. That's not functionally necessary, but it does provide added security as it precludes a local file inclusion attack.

McEileen:
@dismantl Thanks for the guidance.  I followed your suggested refactors and now have image submission working through both the `/submit` and `submit_officer_images/officer/<int:officer_id>` routes.  
I wanted to follow up about the below step in your suggestions for `upload_image_to_s3`.  (I called this method `upload_image_to_s3_and_store_in_db` to distinguish it from the existing `upload_obj_to_s3` method and better describe what it does.  I realize this method, besides having an excessively long name, violates single responsibility principles... Your feedback is welcome.)

> 4. upload to S3 and return new Image instance (don't set `is_tagged`)

Since I call  `upload_image_to_s3_and_store_in_db` with the cropped image buffer, that means I'm creating the new `Image` instance with the cropped image.  Since the `Image` objects are saved in the `raw_images` table, it seems more logical to create the new `Image` instance with the original, full-sized image.  Was there a specific reason you wanted me to save the cropped image to `raw_images`?

This PR isn't ready for review yet, because I need to fix the failing test `test_alembic_has_single_head`.  I wanted to ask you about the above before submitting the PR for review.  Thanks again.

McEileen:
@dismantl PR is ready for review

OpenOversight/app/utils.py
```diff
@@ -24,6 +22,9 @@
                      Incident, Location, LicensePlate, Link, Note, Description, Salary)
 from .main.choices import RACE_CHOICES, GENDER_CHOICES
 
+# Ensure the file is read/write by the creator only
+SAVED_UMASK = os.umask(0o077)
```
r4v5: can you update this to remove the permission `o+x`? Just in case it's relevant.

McEileen: @r4v5 I thought because 7 is the last digit in the octal notation passed to `os.umask()`, other wouldn't have read, write, or execute permissions.   I'd expect `os.umask(0o077)` to cause the file to have the below permissions
`u=rwx,g=,o=`
Let me know if I'm mistaken.  Thanks.

McEileen:
@dismantl I fixed the bugs in the officer identification flow, and the PR is now ready for review (again). 

Could you please clarify the difference between the `img_id` and `original_image_id` fields on the `Face` model? On lines 670-672 in `views.py`, the query that confirmed a user didn't submit a duplicate tag failed because it was using `Face.img_id` instead of `Face.original_image_id`.  I looked at the `Face` model, and I'm having a hard time distinguishing between the two fields.  Thanks!

dismantl:
Thanks for your work on this @McEileen! I believe the idea with those fields was to have a reference to both the original image and the cropped image. I'll start reviewing the PR

OpenOversight/app/main/views.py
```diff
@@ -670,15 +669,24 @@ def label_data(department_id=None, image_id=None):
...
             flash('Invalid officer ID. Please select a valid OpenOversight ID!')
         elif not existing_tag:
             left = form.dataX.data
             upper = form.dataY.data
             right = left + form.dataWidth.data
             lower = upper + form.dataHeight.data
-            cropped_image = get_uploaded_cropped_image(image, (left, upper, right, lower))
+            byte_io = BytesIO()
+
+            if 'http' or 'https' in proper_path:
```
dismantl: I believe this should be `if 'http' in proper_path or 'https' in proper_path:` otherwise it always evals as true

dismantl:
@McEileen Ok I just finished reviewing the current branch and I'll share my notes here. This feature is really coming along! Here's a few things I ran into:
- Images that are either directly uploaded to officer profiles or that are the result of tagging and cropping don't get the following fields set on the Image model: `user_id`, `is_tagged`, `contains_cops`. Not having the user_id just means that we don't know who tagged or uploaded the photo, and not having the is_tagged and contains_cops fields means the resulting image gets put back into the workflow of needing to be sorted and tagged, which they shouldn't. It's possible that this issue predates the PR, but I'm not sure.
- When the `dataX` or `dataY` fields of the cropping form are `0` (which is a legit value for those fields), the form throws a validation error. Changing from a `DataRequired` validator to a `InputRequired` validator should solve this. (this issue predates the PR)
- When photos greater than 300x300 are uploaded to the general submit page, they are cropped to 300x300 before being uploaded to S3 and going into the tagging and cropping workflow. I think it would be preferential to keep the full-res original image, and then just make sure the cropped/tagged images get downscaled to within 300x300. Similarly with >300x300 photos uploaded directly to officer profiles, it would be good to keep the original full-res image, and use the 300x300 downscaled image for the Face tag.

I made some suggested changes in the `506-fork` branch that should address those issues. Feel free to use any of that or not, up to you! It also breaks several tests, so I'm happy to resolve those if you choose to merge any of my suggestions. Thanks again for your work on this!

OpenOversight/app/main/views.py
McEileen: Very true!  Thank you for catching this syntax error.

McEileen:
@dismantl  Thank you for the changes on the `506-fork` branch!  I merged your changes and got the tests to pass.  I was having a hard time with the tests breaking on `user_id=current_user.id` in the `upload_image_to_s3_and_store_in_db` method, because even when I used the `login_user(client)` helper method in the test, `current_user` was an `AnonymousUserMixin` that didn't have an id.  I refactored to make `upload_image_to_s3_and_store_in_db` accept `user_id` as a parameter, so it's now easier to test.  If you have an alternative solution, like mocking out `current_user.id` in the tests, please let me know.

dismantl:
Fantastic work on this @McEileen, and thank you for your persistence on this feature! There were a few minor issues I noticed, so I added a commit to fix them. For one, I noticed we were using the `current_user` object from flask_login incorrectly, so I fixed those instances. I also hit an SQL error when submitting a photo with a date taken EXIF tag, so I fixed the formatting on that. Then the submit officer photo page wasn't limiting the upload dialog to only image files, so I made the Javascript more closely mirror that on the general upload page.

I'll wait for Travis tests to finish, but all tests are passing for me and the feature is working as expected. I give this PR my approval 💯

OpenOversight/app/utils.py
brianmwaters: pretty sure @McEileen is right.
```[int10h@katahdin Desktop]$ umask 077
[int10h@katahdin Desktop]$ touch test
[int10h@katahdin Desktop]$ ls -l test
-rw-------. 1 int10h int10h 0 Aug 19 16:25 test
[int10h@katahdin Desktop]$ ```
@r4v5 can you elaborate?

r4v5: Yeah I was backwards on this

PR #508, opened 2018-07-14
[wip] Release 0.4.1

redshiftzero:
## Status

wip, in the process of deploying, do not merge

## Description of Changes

 - When officers are tagged, a cropped version of their face will be created and stored on S3 (#459).
 - Cropped versions of all tagged faces will be created (#459).
 - Admins/ACs can discuss officers on their profiles via notes (#475).
 - Prepopulate officer IDs on incident form (#476).
 - Do not require a report number to create an incident (#477).
 - Update copy on about and index pages (#487).
 - Replace "police" text with "law enforcement" across the application (#472).
 - Bugfix: Include officers without assignments in search results (#481).
 - Bugfix: Don't show incident area on profile if no incidents are present (#503).

## Notes for Deployment

Due to the changes in #459, this requires careful deployment and I will do so manually:

- [ ] Backup staging
- [x] Run migrations
- [ ] Deploy to staging
- [ ] QA
- [ ] Backup production
- [ ] Run migrations
- [ ] Deploy to prod

PR #509, opened 2018-07-15
Incident form: show which fields are optional

### LINKED ISSUE #505: Incident view does not indicate required fields
I think "description" is a required field but it is unclear. Here is a screenshot, I filled out a title but not a description. It is confusing because we can also add a description to a link but I don't know if we should do both? 

screen shot 2018-07-09 at 2 42 13 pm

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #505

 - Quick fix to show which fields are optional in incident form
 - (and a few more fields will be optional)
 - (want to get this in #508)

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/formfields.py
+++ b/OpenOversight/app/formfields.py
@@ -18,7 +18,7 @@ class TimeField(StringField):
             return self.data and self.data.strftime(self.format) or ''
 
     def process_formdata(self, valuelist):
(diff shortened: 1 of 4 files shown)


redshiftzero:
deployed to staging to check this works

PR #522, opened 2018-07-30
fixing #519

### LINKED ISSUE #519: 500 when adding incident to officer profile
Internal service error generated when adding an incident to an officer profile when text is entered in "city" field but "state" field is left blank

abandoned-prototype:
## Status

## Description of Changes

Fixes #519.
The drop-down for selecting the state in the add incident form has an empty value as its default. Since the field is required, leaving the field on this value shouldn't be allowed.
Similarly if a license plate number is added, the state field cannot be left untouched.

 - Adding a boolean flag to the `allowed_values` function so that it is possible to exclude the empty string as an allowed value, so that not changing the license plate raises a validation flag
 - For the license plate, I added a validator to the form `LicensePlateForm` which doesn't allow the number field to be filled, but the state field be empty.

## Notes
This is my first pull request ever, so looking forward to any feedback.
I didn't find any tests for validating the new incident form, so maybe those should be added.

## Tests and linting

 (except for three unrelated tests that also failed before (related to image loading), not sure what I am missing there)

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/main/forms.py
+++ b/OpenOversight/app/main/forms.py
@@ -17,8 +17,8 @@ from ..models import Officer
 import datetime
 
 
-def allowed_values(choices):
-    return [x[0] for x in choices]
+def allowed_values(choices, empty_allowed=True):
+    return [x[0] for x in choices if empty_allowed or x[0]]
 
 
 class HumintContribution(Form):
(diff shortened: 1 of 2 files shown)


abandoned-prototype:
I added two tests for the incident form validation. I basically copy/pasted and adapted existing code, so the validation errors are tested with a simple string search of the error message in the output.
This means that if the message is changed at a later point the test has to be changed too. Hoping that's fine, otherwise I would appreciate ideas on how else to test.

PR #341, opened 2017-12-27
Add Docker Compose development environment

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #287

PR #523, opened 2018-07-30
Use management interface with Docker dev environment

dismantl:
## Status

Ready for review

## Description of Changes

Currently the management interface (manage.py) won't interact with the Docker dev environment when run from localhost, since it requires direct database access. 

* Moves the Postgres port to standard 5432 and exposes it to the host
* Removes the `runserver` command from manage.py since running it would cause a port conflict with the already running instance in Docker. The `runserver` command is still available as it is a default command in Flask-Script and runs on port 5000.
* Updates the CONTRIB documentation to reflect the changes. Now all management commands should work with the Docker env when run from localhost, including database migrations.

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -50,7 +50,7 @@ Make sure you've started VirtualBox, and then in your project directory, run:
 
 `vagrant up`
 
-This creates a new, pristine virtual machine and provisions it to be an almost-copy of production with a local test database. (Behind the scenes, this is all happening via the files in vagrant/puppet.) If everything works, you should get a webserver listening at `http://localhost:3000` that you can browse to on your host machine.
+This creates a new, pristine virtual machine and provisions it to be an almost-copy of production with a local test database. (Behind the scenes, this is all happening via the files in vagrant/puppet.) If everything works, you should get a webserver listening at `http://localhost:5000` that you can browse to on your host machine.
 
 In addition, you can now SSH into it:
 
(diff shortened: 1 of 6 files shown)


redshiftzero:
hey @dismantl - thanks for updating the contribution docs! just a couple questions/comments:

> Moves the Postgres port to standard 5432 and exposes it to the host

so we moved it from the standard port as a few people had conflicts with postgres running locally: https://github.com/lucyparsons/OpenOversight/pull/341#issuecomment-366483664

if we however want to support devs working on a persistent local postgres instance, maybe we can instead document how to do so in CONTRIB.md? what do you think?

one other thing: since port 3000 is forwarded to the host (both for vagrant and for docker dev env), the fact that the server is running on 5000 is a problem no?

dismantl:
@redshiftzero: If we used a non-standard port, we would need to instruct folks to add the non-standard port to `SQLALCHEMY_DATABASE_URI` in their `.env` file in order for the management scripts (manage.py, create_db.py, test_data.py) to access the database. Using the standard port, that wouldn't be necessary. My personal opinion would be to use the standard port for simplicity's sake, and to instruct folks who have another instance of postgres running on their machine to change the port number in the docker-compose file. But I don't feel strongly either way.

As for the `runserver` command: I think it's actually unnecessary for the Docker environment, since the Docker image entrypoint runs gunicorn anyway on port 3000 and exposes it to the host. All that would happen if you ran the `runserver` command using the Docker environment is that it would start _another_ instance of gunicorn on port 5000 running on the host. So not really that useful I don't think.

dismantl:
Also added a line to the postgres docker service so it will actually be a persistent database :)

Issue #510, opened 2018-07-15
Data import: Buzzfeed NYPD disciplinary data


redshiftzero:
We should crosslink these records from OpenOversight from NYPD: https://www.buzzfeed.com/kendalltaggart/nypd-police-misconduct-database

This means, for every entry in their database, we should make an incident and link to the NYPD officers involved. Since we don't have the roster, we can create the officers as we go. I uploaded one incident as a demonstration: https://openoversight.com/incidents/2.

These are the kinds of datasets we could write an import script like in #392 for. Note that developer time is limited, but a lot of people are interested in helping OpenOversight that are not developers, so we _could_ e.g. organize events where people input this kind of data into the database manually (by temporarily making their accounts area coordinators).

abandoned-prototype:
This is an interesting project that I would like to work on. Looking into it, I have two questions, or points of discussion:

- The badge number (or Tax number):
    The last two digits of every cop's badge number are redacted. So I am not sure if they should be added to the database in some way, as on the one hand we don't know the exact number, but on the other hand, the number is necessary to differentiate between cops with the same name.
- How to incorporate data from the pdfs:
  The OCR-txt files provided by the file host documentcloud.org are not very good, and I doubt they can be used to extract data like incident date or (even harder) a description of the incident. The options that I see here are to either utilize better OCR software or leave this part to volunteers.

redshiftzero:
great questions @abandoned-prototype !

> The last two digits of every cop's badge number are redacted. So I am not sure if they should be added to the database in some way, as on the one hand we don't know the exact number, but on the other hand, the number is necessary to differentiate between cops with the same name.

clarifying question re: NY probably for @camfassett: is the tax registry number a static number that does not appear on an officer's uniform? if so then we should implement #462 and input the numbers exactly as is on the BuzzFeed database, e.g. 1234xx, and when we get more information we can fill in the missing digits. 

> How to incorporate data from the pdfs:

so we have someone who is currently working on the OCR (privately off GitHub), will update on the status on that in the next few days. They are _not_ working on the upload though, just extracting the text from the PDFs, so figuring out how to insert that data into OpenOversight (ideally in an idempotent manner) would be a great contribution in the meantime. How were you thinking of doing the upload? manage.py command?

camfassett:
[...]

Issue #462, opened 2018-06-13
Add tax registry number field


camfassett:
For some police departments like NYPD, the unique identifier for each officer is NOT badge number, but rather tax registry number. While lower ranked POs are identified by both badge number and tax no, but higher ranked officers like lieutenants and above do not have badge numbers—just tax nos. We should add a field for tax registry numbers, at least for NYPD once we have the roster.

camfassett:
The BuzzFeed NYPD complaint database has partial tax registry numbers, and I think we should add this field sooner rather than later so we can add those partials (we don't have shield numbers yet so even these partials will help us keep officers with the same names distinct)

McEileen:
Hi, I recently got involved with Lucy Parsons Labs and I am interested in working on this issue.  I have a few domain questions.
- Are tax registry numbers always only digits?  It seems that way based on what I saw in the BuzzFeed NYPD complaint database, but I wanted to confirm.
- Are tax registry numbers associated with the assignment, or the officer?  I thought it was the latter, but I wanted to clarify because I saw that badge numbers are associated with assignment.
- Should I only show the tax registry number field in forms for officers who belong to the NYPD?

Lastly, thanks for using the `good first issue` label in your repo!  It's really helpful for newcomers. :)

redshiftzero:
Hey @McEileen!

> Are tax registry numbers always only digits? It seems that way based on what I saw in the BuzzFeed NYPD complaint database, but I wanted to confirm.

@camfassett can you confirm this one? If we're not sure we should make the field a String

> Are tax registry numbers associated with the assignment, or the officer? I thought it was the latter, but I wanted to clarify because I saw that badge numbers are associated with assignment.

So as background: the reason why we put the badge numbers in the assignments table is because in many police departments, the badge numbers actually change when an officer is promoted (and then that old badge number may even be reallocated to a new officer), so we track the badge number history of each cop. 

For the tax registry number, my understanding is that this is a unique ID per officer that does not change as they get promoted (@camfassett lmk if this is incorrect), so it indeed makes sense to store this piece of data in the `officers` table instead of the `assignments` table. 

> Should I only show the tax registry number field in forms for officers who belong to the NYPD?

yep exactly. I would recommend exposing the tax registry number field in the forms to create an officer for all departments (for maximum flexibility), but then if the field it is not populated for a cop then we should not show the tax registry number row at all in the officer profile (instead of displaying `Tax Registry: None`)

[...]

PR #524, opened 2018-07-31
Bulk upload officers

### LINKED ISSUE #247: Management command to add officer rosters

dismantl:
## Status

Ready for review

## Description of Changes

Fixes #247.

 - Change badge_exists to officer_exists to match on full name
 - Added unique constraint to officer table based on full name, since that's how the `officer_exists` function checks for duplicates.
 - Adding officer suffix field to database migration

## Notes for Deployment

I tested using the following csv:
```
department_id,badge,last_name,suffix,first_name,middle_initial,race,gender,employment_date,birth_year,rank
1,129022,butts,jr.,seymour,e,black,female,01-01-2001,2000,POLICE OFFICER
```

Output from populating a fresh db w/ test data, then using the bulk uploader:
```sh
$ python ../create_db.py 
$ python ../test_data.py -p
[*] Populating database with test data...
[*] Completed successfully!
$ python manage.py bulk_add_officers ../test.csv 
created 1 officers
$ python manage.py bulk_add_officers ../test.csv 
Skipping creation of existing officer seymour butts for department id 1
created 0 officers

```

## Tests and linting

 - [x] `flake8` checks pass


dismantl:
> It is not impossible that there are several cops with the same name especially in big departments.

Are we sure this is true? Just seems unlikely to me. This would mean the officers have the same first, middle, last, and suffix. I'd rather not check against badge numbers because, as in Chicago and Baltimore, badge numbers are recycled (and sometimes not even tracked properly or at all). In my opinion, it would be good to allow uploading rosters of officers even when badge numbers aren't available.

Long term, I think it would make more sense to use generalized unique IDs for officers to check for duplicates (#462) rather than names or badge numbers. I would also suspect those identifiers are easier to get in most places, since they are usually used by a department's or city's HR. That's true for Baltimore at least, but I'd be curious to hear about other cities.

abandoned-prototype:
I think we have to assume that we don't have access to a middle name (or middle initial) in some cities. I am specifically thinking of NYC, and #510. The data we have there only has first and last names. And of the 1840 entries, there are eight with same name, but different "tax-numbers" which is uniquely assigned to an officer.

But I definitely agree that it would be good to have some unique ID, maybe it would possible to generate that one depending on the department

redshiftzero:
Beautiful, thanks for taking this over @dismantl.

> Are we sure this is true? Just seems unlikely to me. This would mean the officers have the same first, middle, last, and suffix.

Bad news... due to common names like Rodriguez, we already have this situation with CPD:

https://openoversight.com/officer/21111
https://openoversight.com/officer/21112

And we know from comrades in NYPD that this is also the case there (they have 40k+ officers so that makes sense). so we can't add this unique constraint in the database

> In my opinion, it would be good to allow uploading rosters of officers even when badge numbers aren't available.

agreed 100%

> I think we have to assume that we don't have access to a middle name (or middle initial) in some cities.

also agreed 100%

For Chicago, we weren't able to get the internal unique IDs (organizations like the [Invisible Institute](https://invisible.institute/) have tried and failed to get them, I believe the argument by law enforcement was that they are "private" records and not subject to FOIA). For departments where we are able to get them, totally agreed this is the best approach.

dismantl:
Yeah yall are totally right about the name uniqueness issue. So would the duplicate check function return true if all three of badge number, first name, and last name match? And once #462 is completed the function can also check against unique identifiers when they are available.

redshiftzero:
> all three of badge number, first name, and last name match

Ah yep, this should be sufficient

dismantl:
Ok I added updates:

- Updated officer duplicate checking to use first name, last name, badge number
- Added more fields and checks to the CSV import command
- Removed unique constraint from previous commit

Ready for review again.

redshiftzero:
fantastic work! do you have bandwidth to add tests to this one?

dismantl:
Yup I can work on that!

dismantl:
Coming back to this now. One thing I'd like to add to the `officer_exists` function is to check against a unique identifier if that is available in the data, otherwise fall back to using name/badge.

I'd also like this feature to support updating an existing database of officers, not just replacing what's there. That way we can hang on to extra added info such as notes, descriptions, incidents, etc.

sibrahi5:
Flask?

dismantl:
@sibrahi5 Yeah I planned to use Flask CLI commands for it, so I rebased the this branch onto the `flask-cli` branch from PR #556, and have been working off of that. Here's what it looks like so far: https://github.com/lucyparsons/OpenOversight/compare/develop...openjusticebaltimore:247-bulk-upload-officers#diff-fa538f187486df65f41ca0d7d8d17c92R94

dismantl:
Ok I added several tests and fixes, so this is ready for review again.

sibrahi5:
LGTM. :O

PR #529, opened 2018-08-03
Add department dropdown list to photo upload

### LINKED ISSUE #375: Photo upload should have a dropdown for department selection
Right now we have an initial page that selects the department: 

screen shot 2018-02-16 at 9 58 49 pm

This should be a dropdown box on the image upload page itself

nmanfred:
## Status

Ready for review

## Description of Changes

Fixes #375 .

 - Add a drop down list to the photo upload page containing the list of police departments
 - Remove the department selection page before a photo upload
 - By default, select the user's preferred department in the dropdown list

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/main/views.py
```diff
@@ -543,26 +544,20 @@ def submit_complaint():
 @main.route('/submit', methods=['GET', 'POST'])
 @limiter.limit('5/minute')
 def submit_data():
+    preferred_dept_id = 1
```
redshiftzero: The only hiccup here is if primary key = 1 doesn't exist (since this ID is from an autoincrementing primary key, this can happen if the current primary department is not the first one that was created, i.e. if a department was made during testing and deleted). We can instead do something like:

```
preferred_dept_id = Department.query.first().id
```

to be more robust

nmanfred:
Thanks for catching this. I just fixed it and squashed the commits together.

PR #531, opened 2018-08-08
Dump departmental data as CSV files

### LINKED ISSUE #489: Add button for download of all department data
For each department, it would be nice to have an endpoint for the download of all data from a given department in CSV format. Instead of dumping out all the photos, which is expensive for us, as a stopgap we could just dump out the URLs

tomx4096:
## Status

## Description of Changes

Fixes #489 

 - Download a department's officer data as CSV
 - Download a department's incident data (if available) as CSV
 - Unit tests

## Notes for Deployment

The final presentation and formatting of data in the CSVs is undecided and open to feedback. Better __str__ methods in some of the model objects would improve presentation

screenshot from 2018-08-08 07-21-10
screenshot from 2018-08-08 07-21-44

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/main/views.py
```diff
@@ -565,6 +565,85 @@ def submit_department_images(department_id=1):
...
+                "race, born, employment_date, assignments, notes\n"
+
+    assign_dict = {}
+    assign_records = Assignment.query.all()
+    for r in assign_records:
+        if r.officer_id not in assign_dict:
+            assign_dict[r.officer_id] = []
+        assign_dict[r.officer_id].append("(#%s %s %s %s %s)" % (check_input(r.star_no), check_input(r.rank), check_input(r.unit), check_input(r.star_date), check_input(r.resign_date)))
+
+    notes_dict = {}
```
redshiftzero: Let's remove the notes from the CSV, the reason being is that the notes are actually for admins/area coordinators to add private notes to a profile, so we don't want to expose them to regular users

OpenOversight/app/main/views.py
```diff
@@ -565,6 +565,85 @@ def submit_department_images(department_id=1):
...
+        abort(404)
+    dept_name = records[0].department.name.replace(" ", "_")
+    first_row = "id,report_num,date,description,location,licences,links,officers\n"
+
+    record_list = ["%s,%s,%s,%s,%s,%s,%s,%s\n" %
+                   (str(record.id),
+                    check_input(record.report_number),
+                    check_input(record.date),
+                    check_input(record.description),
+                    check_input(record.address),
```
redshiftzero: Let's add a string representation for `Location` so the output here is nice, i.e.:

```
diff --git a/OpenOversight/app/models.py b/OpenOversight/app/models.py
index c4b944b..9e87a13 100644
--- a/OpenOversight/app/models.py
+++ b/OpenOversight/app/models.py
@@ -222,6 +222,20 @@ class Location(db.Model):
     def validate_state(self, key, state):
         return state_validator(state)
 
+    def __repr__(self):
+        if self.street_name and self.cross_street2:
+            return 'Intersection of {} and {}, {} {}'.format(
+                self.street_name, self.cross_street2, self.city, self.state)
+        elif self.street_name and self.cross_street1:
+            return 'Intersection of {} and {}, {} {}'.format(
+                self.street_name, self.cross_street1, self.city, self.state)
+        elif self.street_name and self.cross_street1 and self.cross_street2:
+            return 'Intersection of {} between {} and {}, {} {}'.format(
+                self.street_name, self.cross_street1, self.cross_street2,
+                self.city, self.state)
+        else:
+            return '{} {}'.format(self.city, self.state)
+
```

OpenOversight/app/templates/browse.html
```diff
@@ -14,14 +14,20 @@ <h1><small>Browse a Department</small></h1>
...
-        </p>
-        {% endif %}
+            <a class="btn btn-lg btn-primary" href="{{ url_for('main.download_incidents_csv', department_id=department.id) }}">
+               Download Incident Data
+            </a>
+          {% endif %}
+          <a class="btn btn-lg btn-primary" href="{{ url_for('main.download_dept_csv', department_id=department.id) }}">
+               Download Officer Data
+          </a>
+          </p>
```
redshiftzero: Not necessary for merge on this PR, but just a thought: we should have the officers and incidents links be primary, and the raw data downloads be somewhere less central. One good place on the site might be a link on the footer to a new route that has the CSV download links for each department:

screen shot 2018-08-18 at 6 52 23 pm

If you feel like implementing this (or if you have another better idea!) feel free, else we can defer to a followup.

tomx4096:
Thanks for the reviews @redshiftzero . I removed the notes from the CSV, added your Location representation, and changed the location of the download links to the footer:
screenshot from 2018-08-20 11-27-57
screenshot from 2018-08-20 11-27-48

Issue #527, opened 2018-08-03
Graph/network analysis of department


redshiftzero:
Suggested by @dismantl: a nice way to view command structure in a police department would be a visualization that displays a graph of all officers, where each node in a graph is an individual officer, and the edges represent a command relationship. 

(this is a good issue for anyone interested in d3.js)

tomx4096:
This could be visually cool, but would it just be based purely on rank or the actual department structure? I think information like "officer X reports directly to Y" could be hard for outside observers to document.

tomx4096:
@dismantl i was working on this but was unsure about where the data would come from (i ended up with the above PR). namely, it could be hard for an outsider to observe and document "officer X reports to Y". can it be in FOIA reports? some departments have the higher ranks listed online (http://assets.lapdonline.org/assets/pdf/new_org_chart.pdf) but not the entire roster hierarchy.

dismantl:
@tomx4096 yeah I actually got some of that data in a FOIA response for the Baltimore PD roster. Many, but not all, of the officer records list the ID of their supervisor, which could be used to build a partial hierarchy of the department (though unsure how disjointed it would be due to not all officers having this listed). But there currently isn't a data field in OO for a supervisor ID, so that would need to be added, probably to the `assignments` table I would think.

dismantl:
@tomx4096 you can take a look at the [roster I got for BPD](https://cdn.muckrock.com/foia_files/2018/05/11/Active_employees_as_of_May_3_2018.xlsx). The column `EMPLID` shows city employee ID number, and then the `Supv ID` column indicates the employee ID number of their supervisor, if any. So a graph could potentially be built with that.

tomx4096:
@dismantl thanks for the info and roster. i started a 2nd PR (#541 ) for an actual graph visualization

ejfox:
@tomx4096 @redshiftzero Your PR (https://github.com/lucyparsons/OpenOversight/pull/541) looks awesome to me. Is there any remaining work needed to bring it over? Would love to help bring this feature to the finish line if I can help in any way.

tomx4096:
> @tomx4096 @redshiftzero Your PR (#541) looks awesome to me. Is there any remaining work needed to bring it over? Would love to help bring this feature to the finish line if I can help in any way.

thanks :) it worked fine when I tested it on the Baltimore data, but that was quite some time ago. performance could be much better (it takes more than a few seconds to load the visualization) but I'm not sure how to improve it.

[...]

Issue #500, opened 2018-07-08
Unify browse and find views


redshiftzero:
Another very good suggestion by @nmorduch: The browse view looks like: 

<img width="879" alt="screen shot 2018-07-08 at 4 09 51 pm" src="https://user-images.githubusercontent.com/7832803/42424892-5c5cbeac-82c9-11e8-9e58-7a9b091e60e9.png">

And the officer find results look like: 

<img width="856" alt="screen shot 2018-07-08 at 4 10 17 pm" src="https://user-images.githubusercontent.com/7832803/42424896-6b5b024c-82c9-11e8-80c3-31a953c3fff6.png">

We should unify the two displays (favoring the browse view). Note that I do still think it is useful to have the series of "Find" questions as an entrypoint to the database as it is less overwhelming and is more useful for someone trying to find a specific cop

redshiftzero:
Closed by #662

Issue #474, opened 2018-06-24
Improve design of find the officer flow


redshiftzero:
Now that #459 is working its way in, we should remove the carousel in the search results (it's bad), in favor of e.g. a page that shows N results and images at a time.

nmorduch:
Noting here that I my design will probably address #500 and #502

Issue #326, opened 2017-12-10
Admin can add ranks


redshiftzero:
Admin or area coordinator should be able to add ranks to the database

redshiftzero:
And these ranks should be per-department

redshiftzero:
Closed by #619

PR #533, opened 2018-08-12
browse department as directory

tomx4096:
## Status

## Description of Changes

Related to #527 #474 #500 , provides a directory view of the department that users can quickly Ctrl-F through if they know name or star number. 

 - Add a departmental view button to browse.html
 - Leads to directory page that listing all officers in a department, with links to officer profiles
 - 

## Notes for Deployment

- Depends on maintaining ranks in RANK_CHOICES in high-to-low order, excluding Not Sure
- Only Commander, Captain, PO, and Not Sure rank included in template right now

screenshot from 2018-08-12 03-19-20
screenshot from 2018-08-12 03-19-09
screenshot from 2018-08-12 03-18-44
screenshot from 2018-08-12 03-18-57

## Tests and linting

 - [x] `flake8` checks pass


dismantl:
This is great, thanks! I was able to get it to work with the test data, but when I tried to use Baltimore data the ranks don't match up with what's in `RANK_CHOICES` and so it throws an error. Is there a way this could gracefully handle departments that have different sets of ranks?

tomx4096:
Thanks for the feedback. You're right, it could be hard to manually maintain proper rank names for many departments with slight variations. I pushed another approach that makes it independent of RANK_CHOICES. It makes up a list of ranks found in the data and uses those to demarcate sections in the directory view. The html was also generalized so all ranks found would be displayed, with Not Sure at the end, and a random (blue-ish) background color. 

The disadvantage of this is that they aren't guaranteed to be ordered from high to low rank in the directory view. 

Let me know if it fixes your error, I can also try your dataset if you like, just let me know where to find it.

OpenOversight/app/main/views.py
```diff
@@ -305,6 +305,47 @@ def list_officer(department_id, page=1, from_search=False):
...
+            assign_dict[record.officer_id] = record
+        else:
+            if record.star_date > assign_dict[record.officer_id].star_date:
+                assign_dict[record.officer_id] = record
+
+    officer_nodes = {}
+    rank_types = ["NOT SURE"]
+
+    for officer in officers:
+        rec = assign_dict[officer.id]
```
dismantl: It's possible there exist officers that don't have assignments, so this should probably be wrapped in a try block:
```
try:
    rec = assign_dict[officer.id]
except KeyError:
    continue
```

dismantl:
Cool, I like that approach and I was able to get it working for Baltimore data. We have like dozens of job titles, some with only 1 or 2 employees, so that made the list pretty long and unwieldy. But the solution to that is probably just to filter out what officers I import (e.g. not forensic scientists).

I wonder if the best way to allow department-specific ranks but not have the disadvantages you listed above, would be to have a per-department list of ranks provided when a department is created. Maybe by uploading an ordered CSV or otherwise storing it in the database? But that may also be a follow-on issue beyond the scope of this PR :)

OpenOversight/app/main/views.py
```diff
@@ -305,6 +305,47 @@ def list_officer(department_id, page=1, from_search=False):
...
+        if len(rec.rank) <= 1:
+            rec.rank = "NOT SURE"
+        safe_rank = rec.rank.upper()
+        if safe_rank not in rank_types:
+            # keep NOT SURE at the end so it will display last on page
+            rank_types.insert(0, safe_rank)
+        if safe_rank not in officer_nodes.keys():
+            officer_nodes[safe_rank] = []
+        else:
+            officer_nodes[safe_rank].append({'id': officer.id,
```
dismantl: This shouldn't be in an `else` block or it won't add the first officer of a rank it finds.

OpenOversight/app/main/views.py
tomx4096: thanks, i hadn't considered this case. fixed in last commit

OpenOversight/app/main/views.py
tomx4096: ah ha! thanks for catching this, fixed in last commit

redshiftzero:
> I wonder if the best way to allow department-specific ranks but not have the disadvantages you listed above, would be to have a per-department list of ranks provided when a department is created. Maybe by uploading an ordered CSV or otherwise storing it in the database?

:+1: to this, storing the ranks in the database is the right way to go (fine to do in a separate issue/PR imo, see #326)

OpenOversight/app/main/views.py
```diff
@@ -305,6 +305,50 @@ def list_officer(department_id, page=1, from_search=False):
...
 
 
+@main.route('/department/<int:department_id>/chart')
+def dept_org_chart(department_id):
+    department = Department.query.filter_by(id=department_id).first()
+    if not department:
+        abort(404)
+
+    officers = Officer.query.filter(Officer.department_id == department_id).all()
+    # map each officer to their most recent assignment
```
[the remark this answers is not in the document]
tomx4096: at first i tried using this relationship, but I found that accessing assignments on each Officer was very slow (sometimes timing out the browser on the test data). i found it much faster to query all Assignments at once and then associate them to officers in the loop. additionally, i wasn't sure that the assignments would be guaranteed to be in chronological order in the list

PR #552, opened 2018-08-25
First pass at python3 support

r4v5:
## Status

in progress

## Description of Changes

Adds python3 support, builds a python3 development environment instead of python2.

 - python3 support
 - change relative import to explicit relative import (see [the docs](https://docs.python.org/3/tutorial/modules.html#intra-package-references) for more info)
 - urlparse got absorbed into urllib.parse in python3, so conditionally import it from the new place but fall back to the python2 place
 - fixups in manage.py and backup_db.py
 - change path in test data to a non-relative path so images display (in production these are AWS URLs so this shouldn't be an issue, but they were 404ing from /department and /officer/<id>)

## Notes for Deployment

 - don't, yet, at all

## Tests and linting

 - [ ] `flake8` checks pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ addons:
   firefox: "45.4.0esr"
 python:
   - "2.7"
+  - "3.6"
 before_install:
   - wget https://github.com/mozilla/geckodriver/releases/download/v0.11.1/geckodriver-v0.11.1-linux64.tar.gz
   - mkdir geckodriver
--- a/OpenOversight/app/__init__.py
+++ b/OpenOversight/app/__init__.py
@@ -9,7 +9,7 @@ from flask_limiter.util import get_remote_address
 from flask_login import LoginManager
 from flask_mail import Mail
 
-from config import config
+from .config import config
 
 
 bootstrap = Bootstrap()
--- a/OpenOversight/app/auth/views.py
+++ b/OpenOversight/app/auth/views.py
@@ -1,3 +1,5 @@
+from future.utils import iteritems
+
 from flask import render_template, redirect, request, url_for, flash, current_app
 from flask.views import MethodView
 from flask_login import login_user, logout_user, login_required, \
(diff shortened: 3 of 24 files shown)


redshiftzero:
just played around a bit with this in the dev env and this looks good! the thing to work out here is: what do you think about the deployment story here (since we need to use a new `python3` virtualenv)?

redshiftzero:
for posterity: I did not test S3 upload functionality, which we should definitely test on a staging server prior to deploying this on prod

r4v5:
I also haven’t yet tested s3. My thoughts are a completely new staging server, to test the recipe in chef and the code, but i’m not yet certain how we’d want to do the cutover for prod because we’d need letsencrypt tls keys on the box before cutting over dns.

Proposal: creation of oo-staging.lucyparsonslabs.com on a new clean DO VM, and if s3 testing and letsencrypt work there, we also move staging.openoversight.com and staging.oo.lpl.com to it.

If that all succeeds, I hand-copy the /etc/letsencrypt from the existing prod server to a new prod-oo.lpl.com so it has valid certs for them, deploy the code and secrets, and retarget dns for the production domains.

r4v5:
Note that as currently written, this should be both 2.7 and 3.x compatible.

r4v5:
Okay. So as this stands every test should pass, with one exception; test_utils.py:test_get_uploaded_cropped_image_s3_error [fails on python3](https://travis-ci.org/lucyparsons/OpenOversight/jobs/455827158#L1055) because of ...some horrible interaction of magicmock and [the traceback-generating code in the exception handler](https://github.com/lucyparsons/OpenOversight/pull/552/files#diff-3110b4868e45e11a2432598cb73a53c7R473).

@redshiftzero do you have any ideas how that would go about being fixed, short of removing the code on now-line-473? I'm truly stumped.

r4v5:
:tada: tests are green in both py2 and py3

PR #566, opened 2018-10-28
479/badge number optional

### LINKED ISSUE #479: Make badge number field optional
For departments where getting badge number information is difficult, we should allow the star_no field in the assignments table to be optional (`nullable=True`)

McEileen:
## Status

Ready for review

## Description of Changes
Makes officer badge number optional.  Fixes #479  .

 - Make `star_no` in the `assignments` table nullable.
## Tests and linting

The test `test_admins_can_edit_incident_links_and_licenses` fails, however, I don't think it is caused by my changes. 
 - [ ] `flake8` checks pass
The `flake8` checks fail on the generated migration because of indentation errors.  I wasn't sure how to resolve this, since they're generated files, so I left them as is.

### DIFF
--- a/OpenOversight/app/models.py
+++ b/OpenOversight/app/models.py
@@ -109,7 +109,7 @@ class Assignment(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     officer_id = db.Column(db.Integer, db.ForeignKey('officers.id'))
     baseofficer = db.relationship('Officer')
-    star_no = db.Column(db.String(120), index=True, unique=False)
+    star_no = db.Column(db.String(120), index=True, unique=False, nullable=True)
     rank = db.Column(db.String(120), index=True, unique=False)
     unit = db.Column(db.Integer, db.ForeignKey('unit_types.id'), nullable=True)
     star_date = db.Column(db.Date, index=True, unique=False, nullable=True)


OpenOversight/migrations/versions/6199adca9784_.py
```diff
@@ -0,0 +1,285 @@
...
+# revision identifiers, used by Alembic.
+revision = '6199adca9784'
+down_revision = '0acbb0f0b1ef'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('departments',
```
r4v5: I feel like the upgrade/downgrade steps here are way bigger than they should be; could you try it again after making sure the DB was created properly? I _think_ it should just be something like
```suggestion
    op.alter_column('assignments', 'star_no', nullable=True)
```

McEileen: Sure, I will look into it.  Also, I got a new computer and I will be creating the DB from scratch, anyway.  I'm part of the docker on linux crew now!

McEileen:
I looked at this with @r4v5 during the hackathon.  We realized that the database already had the `star_no` field in assignments as nullable, which means that the database wasn't preventing `star_no` from being optional.  Instead, the ORM (`SQL_ALCHEMY`) caused this bug, which I already addressed by updating the `assignment` model.  Since no database change is needed, I deleted the migration.
Please review, @redshiftzero!

redshiftzero:
this looks like it needs a rebase on `develop` (got lots of unrelated commits in here)

PR #595, opened 2018-12-16
Makefile improvements

dismantl:
## Status

Ready for review

## Description of Changes

Separate out commands in Makefile for easier use
Wait until Postgres is ready before creating DB or populating test data

## Notes for Deployment

`make dev` will now create the docker environment and the database, but will no longer populate it with test data. `make populate` was added to do that.

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ create_db: start
 	docker-compose run --rm web /usr/local/bin/python ../create_db.py
 
 .PHONY: dev
-dev: build start create_db
+dev: build start create_db populate
 
 .PHONY: populate
 populate: create_db  ## Build and run containers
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,36 +1,53 @@
...
+	@until docker exec -it openoversight_postgres_1 psql -h localhost -U openoversight -c '\l' postgres &>/dev/null; do \
+		echo "Postgres is unavailable - sleeping..."; \
+		sleep 1; \
+	done
+	@echo "Postgres is up"
+	## Creating database
+	docker-compose run --rm web /usr/local/bin/python ../create_db.py
+
+.PHONY: dev
+dev: build start create_db
```
redshiftzero: actually - it seems more reasonable to load the test data by default in the dev env, no?

dismantl: Just made the change

PR #601, opened 2018-12-17
include age next to birth year in officer profile

### LINKED ISSUE #569: Replace birth year with age on officer profile
Officer profiles currently display birth year when that data is available. It would probably be easier for the end user if it simply displays the officer's age (assuming DOB is available) instead of requiring them to do the math in their head.

asabine:
## Status

Ready for review

## Description of Changes

Fixes #569 
Added a template filter to calculate age based on birth year, and updated the officer profile to include the age alongside the birth year in parenthesis e.g., `1955 (63)`.

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/templates/officer.html
+++ b/OpenOversight/app/templates/officer.html
@@ -55,8 +55,8 @@
                 <td>{{ officer.gender }}</td>
               </tr>
               <tr>
-                <td><b>Birth Year</b></td>
-                <td>{{ officer.birth_year }}</td>
+                <td><b>Birth Year (Age)</b></td>
+                <td>{{ officer.birth_year }} ({{ officer.birth_year|get_age }})</td>
               </tr>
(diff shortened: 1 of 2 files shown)


OpenOversight/app/templates/officer.html
```diff
@@ -55,8 +55,8 @@ <h3>General Information
                 <td>{{ officer.gender }}</td>
               </tr>
               <tr>
-                <td><b>Birth Year</b></td>
-                <td>{{ officer.birth_year }}</td>
+                <td><b>Birth Year (Age)</b></td>
+                <td>{{ officer.birth_year }} ({{ officer.birth_year|get_age }})</td>
```
redshiftzero: We should probably not display `None` in these fields when birth year is not populated, however, that issue was not introduced in this diff, so merging

PR #618, opened 2019-01-06
Update requirements.txt

allanice001:
psycopg2 has been renamed to psycopg2-binary

## Status

## Description of Changes

Fixes # .

 - 
 - 
 - 

## Tests and linting

 - [ ] `flake8` checks pass

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ Flask-Migrate==2.1.1
 Flask-SQLAlchemy==2.3.2
 Flask-WTF==0.14.2
 Pillow==5.1.0
-psycopg2>=2.6.2
+psycopg2-binary==2.7.6.1
 SQLAlchemy==1.1.15
 gunicorn==19.9
 Fabric==1.14.0


redshiftzero:
Hey @allanice001 thanks for the PR! Check out the docs in http://initd.org/psycopg/docs/install.html#binary-install-from-pypi - this indicates that for production use we should not use the wheel, i.e. it looks like we should continue to use `psycopg2` (there's also some useful discussion/background on this issue over in the [psycopg2 bugtracker here](https://github.com/psycopg/psycopg2/issues/674)). Take a look and let me know what you think!

allanice001:
Hey @redshiftzero , my PR is for easier (less dependencies on the host level) support
On many production servers i work with, we remove development libraries as well as c compilers for security reasons, thus moving to the pre-built binary makes sense.

Also from the docs:

- release psycopg 2.7.4 as a normal 2.7.x, both as source and binary packages;
- also release a package called psycopg2-binary version 2.7.4;
- importing the wheel package installed from 'pip install psycopg2'
will result in a
  warning:
      The psycopg2 wheel package will be renamed from release 2.8;
      to keep using the binary package please install 'psycopg2-binary' instead.
- importing the wheel package installed from 'pip install
psycopg2-binary' will cause
  no such warning;
- starting from release 2.8, the binary packages will only be released under
  psycopg2-binary.

Thus updating the requirements file is future-proofing the package.  (at least in my mind it is)

redshiftzero:
hi @allanice001, thanks for this PR but since the docs are pretty clear in suggesting to avoid this wheel for prod use, the conservative choice here seems to be to use the source package (fwiw this is what most projects are doing over in the upstream issue I pointed to above). As such I'm going to close this one in favor of #631

But please check out the issues tagged [good first issue](https://github.com/lucyparsons/OpenOversight/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) if you'd like to make another PR!

PR #619, opened 2019-01-12
Department-specific ranks

### LINKED ISSUE #572: Per-department configurable ranks
Each police department may have a different set of ranks. Instead of hardcoding the list in `choices.py`, it would be good to allow the admin of a department to input an ordered list of ranks. Maybe this could be included on a per-department settings page of some sort?

dismantl:
## Status

Ready for review

## Description of Changes

Fixes #572, #326.

This PR gets rid of the default ranks and instead stores department-specific ranks in the new `ranks` table. When assignments are added or edited, the officer's rank is entered in a regular text field now instead of a dropdown menu. Therefore officers can have any rank, not just the ones listed in the `ranks` table for that department. However, the department-specific ranks are used to populate the filters in the browse and find pages.

## Notes for Deployment

When a new department is created, a 'Not Sure' rank is added by default, as well as any ranks entered in the fields on the add/edit department page. To test, create or edit a department and set the ranks however you like. Then, go to `/browse` or `/find` and you should see the ranks you just entered in the Rank dropdown field.

Next, add or edit an officer's assignments to set the rank to whatever you like. This should then be displayed on the officer's profile page, as well as in the officer list in `/department/<dept_id>`.

image

## Tests and linting

 - [x] `flake8` checks pass


sibrahi5:
"Therefore officers can have any rank, not just the ones listed in the ranks table for that department. However, the department-specific ranks are used to populate the filters in the browse and find pages"
I'm not sure why someone would have a rank not listed in the ranks table for that department. Could you clarify this a bit, as right now it seems like we could have a problem with folks misspelling the name of the department rank, and therefore we'd have iterations like 
police fficer
police office
etc eventually that would require manual cleanup, when they can just be presented the department rank options pertinent to their pd.

dismantl:
@sibrahi5 the reason I did that is because the Baltimore PD roster has 100 different unique job titles, but there's only like 10 "main" ranks: https://en.wikipedia.org/wiki/Baltimore_Police_Department#Rank_structure_and_insignia. By not requiring each rank to match one within a defined set, we can still store and display all those ~100 ranks, but still search for officers by the main ranks.

There certainly is the downside that when people manually enter or edit officers they may make a typo in the rank, but that seems like less of a problem (at least for Baltimore) than not being able to display everyone's ranks as listed in the roster.

sibrahi5:
@dismantl Perhaps a separation of data then- job title table different from rank table. I think it's important to differentiate between the two, as ranks are more rigid over time, whereas job titles can change all the time.

dismantl:
Are you thinking they would also be displayed separately? For our roster, we have only the one datapoint of "job title" and some of them fit the standard ranks but many of them don't.

OpenOversight/app/templates/add_edit_department.html
```diff
@@ -15,10 +16,49 @@ <h1>{% if update %} Update {% else %} Add {% endif %} Department</h1>
...
+                {%- endif %}
+              </div>
+            </fieldset>
+            {% endfor %}
+          {% else %}
+            <fieldset>
+              <div class="form-group ">
+                <input class="form-control" custom_name="ranks-1" id="ranks-1" name="ranks-1" type="text" value="">
+              </div>
+              <button class="btn btn-danger js-remove-button" style="margin-bottom: 2rem;">Remove</button>
```
redshiftzero: I think there is a bug here, check out the following STR:

0. Go to add a "New Department"
1. Enter first rank name.
2. Click add a new rank and enter a second rank name.
3. Click remove on the new rank.

I would expect that just the most recently added rank would disappear, but the entire rank section disappears

OpenOversight/app/models.py
```diff
@@ -37,6 +37,22 @@ def __repr__(self):
         return '<Department ID {}: {}>'.format(self.id, self.name)
 
 
+class Rank(db.Model):
+    __tablename__ = 'ranks'
+
+    id = db.Column(db.Integer, primary_key=True)
+    rank = db.Column(db.String(255), index=True, unique=False, nullable=False)
```
redshiftzero: hmm, shouldn't there be a relationship between the rank and assignments tables (e.g. a rank added there should point to a column here)? i see that you're keeping the `rank` field in the assignments table as a separate field. is this preserved for the jobs vs ranks distinction you mention in the PR comments? If so, I think it will behoove us to have a generic table like `Jobs` and we could have a field in there that indicates whether or not it is a "officer rank", i.e. `is_sworn_officer`. what do you think? let me know if i misunderstand the purpose here

dismantl: Yup, that's basically the idea. The entries in this table basically just populate the rank dropdowns in the find and browse forms, while the `rank` field in the assignments table would list the job title, whether it was a "rank" or not. 

As for generalizing the table to `Jobs` with a `is_sworn_officer` field, I guess I just don't think it's necessary since the jobs that aren't sworn officer ranks aren't used in any forms or queries, and are already in the assignments table in the `rank` field. Are there other things you think a jobs table could be used for?

OpenOversight/app/models.py
```diff
@@ -37,6 +37,22 @@ def __repr__(self):
...
+    order = db.Column(db.Integer, index=True, unique=False, nullable=False)
+    department_id = db.Column(db.Integer, db.ForeignKey('departments.id'))
+    department = db.relationship('Department', backref='ranks')
```
redshiftzero: I think we want a unique constraint on `rank` name and `department_id` here, right now we can add duplicate ranks for the same department (see `Face` table)

OpenOversight/app/models.py
redshiftzero: Re-reading the above I think what I said was unclear: I was suggesting _instead_ of a `Ranks` table, we have a `Jobs` (or `EmployeeRole` or something) table, and the `is_sworn_officer` field in that field would determine whether or not that role should appear in the rank dropdowns (we don't need the `is_sworn_officer` flag for openoversight.com but I think y'all do for Baltimore hence this suggestion). 

The reason I'm advocating this is because there is duplication of data in the assignments table, which will complicate maintainability / extensibility (violates [2NF of database normalization](https://en.wikipedia.org/wiki/Second_normal_form)). For example, the assignments table and ranks/jobs table linked via foreign key would be useful:
* if we add data to the rank that we want to easily access on the officer's profile. Some possibilities are using a picture of the rank insignia, a description of what tasks that rank tends to do (maybe available in a tooltip), typical salary range of the rank (e.g. if at some later date we wanted to display "officer is 10% above the average" on their profile)))
* if for example a rank is renamed, e.g. we want to rename all `PO`s to `Police Officer` (that is exactly [this situation](https://github.com/lucyparsons/OpenOversight/pull/619/files#r257579546) which we have on openoversight.com due to the initial bad denormalized design which now requires a migration to address)

What do you think?

OpenOversight/app/templates/add_edit_department.html
redshiftzero: (I think this is not yet resolved, lmk if I'm wrong or you were unable to repro)

OpenOversight/app/utils.py
```diff
@@ -249,15 +249,8 @@ def filter_by_form(form, officer_query, is_browse_filter=False):
                                                 Officer.birth_year == None))  # noqa
 
     officer_query = officer_query.outerjoin(Assignment)
-    if form['rank'] == 'PO':
-        officer_query = officer_query.filter(
-            db.or_(Assignment.rank.like('%%PO%%'),
-                   Assignment.rank.like('%%POLICE OFFICER%%'),
-                   Assignment.rank == 'Not Sure')  # noqa
```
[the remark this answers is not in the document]
redshiftzero: I can write this migration (unless someone else wants to) and append in a commit when this is ready for merge (just so that we don't merge something into `develop` that we can't yet deploy), it should update ranks named exactly "PO" to "POLICE OFFICER"

OpenOversight/app/main/views.py
```diff
@@ -355,6 +391,15 @@ def list_officer(department_id, page=1, from_search=False, race='Not Sure', gend
         max_age=form_data['max_age'])
 
 
+@main.route('/department/<int:department_id>/ranks')
+@main.route('/ranks')
+def get_dept_ranks(department_id=None):
```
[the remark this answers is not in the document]
redshiftzero: I wrote a couple tests and added them in a subsequent commit

OpenOversight/app/main/forms.py
```diff
@@ -165,8 +166,8 @@ class AddOfficerForm(Form):
     star_no = StringField('Badge Number', default='', validators=[
         Regexp('\w*'), Length(max=50)])
     unique_internal_identifier = StringField('Unique Internal Identifier', default='', validators=[Regexp('\w*'), Length(max=50)])
-    rank = SelectField('Rank', default='PO', choices=RANK_CHOICES,
-                       validators=[AnyOf(allowed_values(RANK_CHOICES))])
+    rank = StringField('rank', default='', validators=[
```
redshiftzero: I think there is a danger in making these string fields instead of selecting from the database just because based on the error rates we've seen with people adding data to the openoversight.com database, we'll end up with a bunch of mispellings/whitespace all for the same officer rank. I think we need to provide some more guard rails here for people doing data entry. How about we make these rank selections as `QuerySelectField` for `FindOfficerForm` and `AssignmentForm` and then have a "Add/edit/delete rank" page.

OpenOversight/app/main/views.py
```diff
@@ -274,6 +274,23 @@ def add_department():
                                     short_name=form.short_name.data)
             db.session.add(department)
             db.session.commit()
+            db.session.add(Rank(
+                rank='Not Sure',
+                order=0,
+                department_id=department.id
+            ))
+            db.session.commit()
+            if form.ranks.data:
```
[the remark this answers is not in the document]
redshiftzero: (looks like we still need additional test coverage, lmk if I'm wrong)

OpenOversight/app/models.py
dismantl: Yup that makes sense and I agree that's a good way to go. I'll refactor.

OpenOversight/app/main/forms.py
dismantl: Good call, we don't want to end up with a bunch of typos. @camille-s is working on normalizing our rank data since our initial rosters were pretty inconsistent, so we'll be able to create the rank choices from that.

dismantl: For Baltimore we have like 15 sworn officer ranks and then like 80 other job titles. That seems like too many to display in a form on the Add/Edit Department page. Does it seem okay to just include the sworn officer job titles on the Add/Edit Department page and assume admins will import the other job titles via the `bulk-add-officers` import command?

redshiftzero: Sure that's OK with me, we can file a followup for adding editing of non sworn officer roles via the web interface

OpenOversight/app/models.py
redshiftzero: btw we should have `is_sworn_officer` be default True

dismantl:
@redshiftzero Ok I've refactored and everything seems to be working well. Here's the major changes:
- Fixes #655. The fix requires using SQL window functions, which aren't available in sqlite3 (used for tests) until version 3.25.0. Upgrading the sqlite3 version in Travis proved extremely difficult so instead I configured Travis to run the tests within docker.
- Refactored `Rank` model, now `Job` that includes a `is_sworn_officer` field.
- Add Officer form now dynamically sets the job title `<select>` field options with Javascript, allows selection of any job title in the department, not just sworn officer ranks.
- Assignment form on officer profile now uses a `<select>` field for job title, rather than a text field for rank as before, and shows unit descriptions rather than unit IDs.
- Edit Assignment form now allows selection of no unit.
- When assignment has no unit selected, label it as 'None' rather than 'Unknown'.
- The ranks form on the Add/Edit Department page uses jQuery UI to allow drag-and-drop re-ordering.

redshiftzero:
Merged #662 which contains these changes

PR #623, opened 2019-01-25
Pull FLASK_ENV from host environment so tests will run in docker

dismantl:
## Status

Ready for review

## Description of Changes

Commit f850c0f44262617c9f101b257951b208f61b7869 added the `FLASK_ENV` environment variable to the docker web service. Because it is set to `development`, pytest will fail when run within docker, such as when you run `make test`.

This fix pulls FLASK_ENV from the host environment so that `make test` will run without error.

## Tests and linting

 - [x] `flake8` checks pass


redshiftzero:
nice, thanks!

PR #640, opened 2019-03-12
add documentation about pdb

### LINKED ISSUE #638: Documentation about pdb
I will add documentation about how to run the app with pdb and how to run a single test with pdb.

McEileen:
## Status

Ready for review

## Description of Changes

Fixes #638  .

 - Add explanation of how to use pdb with both tests and the app itself.
 

## Tests and linting

I didn't run pytests because I only changed documentation.

 - [ ] `flake8` checks pass
I didn't run `flake8` because I only changed documentation.

### DIFF
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -179,3 +179,17 @@ Password:
 Type your password again:
 Administrator redshiftzero successfully added
 ```
+
+## Debugging OpenOversight - Use pdb with the app itself
+Debugging OpenOversight requires a few minor changes to `docker-compose.yml`.  Add the below two lines to the `web` service in the `docker-compose.yml`, below the line that specifies the port.
+```yml
+   stdin_open: true
+   tty: true    
+```
+To set a breakpoint in OpenOversight, add `import pbd` to the file you want to debug, then write `pdb.set_trace()` where you want to drop a breakpoint.
+Next, in your terminal run `docker ps` to find the container id of the `openoversight_web` image, then run `docker attach ${container_id}` to connect to the debugger in your terminal.  You can now use pdb prompts to step through the app.
+
+## Debugging OpenOversight - Use pdb with a test
+If you want to run an individual test in debug mode, use the below command.
+`docker-compose run --rm web pytest --pdb -v tests/ -k <test_name_here>`
+Again, add `import pdb` to the file you want to debug, then write `pdb.set_trace()` wherever you want to drop a breakpoint.  Once the test is up and running in your terminal, you can debug it using pdb prompts.
--- a/CONTRIB.md
+++ b/CONTRIB.md
@@ -181,15 +181,17 @@ Administrator redshiftzero successfully added
 ```
 
 ## Debugging OpenOversight - Use pdb with the app itself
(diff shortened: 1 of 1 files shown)


CONTRIB.md
```diff
@@ -140,3 +140,17 @@ Password:
 Type your password again:
 Administrator redshiftzero successfully added
 ```
+
+## Debugging OpenOversight - Use pdb with the app itself
+Debugging OpenOversight requires a few minor changes to `docker-compose.yml`.  Add the below two lines to the `web` service in the `docker-compose.yml`, below the line that specifies the port.
```
pierwill: How about:


> In `docker-compose.yml`, below the line specifying the port number, add the following to the `web` service:

CONTRIB.md
```diff
@@ -140,3 +140,17 @@ Password:
...
+```yml
+   stdin_open: true
+   tty: true    
+```
+To set a breakpoint in OpenOversight, add `import pbd` to the file you want to debug, then write `pdb.set_trace()` where you want to drop a breakpoint.
```
pierwill: How about:

> To set a breakpoint in OpenOversight, first import the pdb module by adding `import pbd` to the file you want to debug. Call `pdb.set_trace()` on its own line wherever you want to break for debugging.

CONTRIB.md
pierwill: ^ The above is a suggestion for the second sentence on line 145

McEileen:
@redshiftzero @pierwill 
Eep, the documentation about debugging is incomplete. About a week ago, I found a bug where if I placed the pdb breakpoint in a class, like `FindOfficerForm()` in `forms.py`, the website would hang and never load.  When I used `docker attach ${web_container_id}`, the terminal didn't display the pdb prompt.  I was able to access the pdb prompt by killing the docker web container, then restarting it with `docker-compose run --rm -p 3000:3000 web flask run --host=0.0.0.0 --port=3000`.  However, when I entered a command into the pdb prompt, it mutated the input and didn't recognize it.  See below for an example:

```bash
[34] > /usr/src/OpenOversight/OpenOversight/app/main/forms.py(36)FindOfficerForm()
-> 'name', default='', validators=[Regexp('\w*'), Length(max=50),
(Pdb) list
*** NameError: name 'lt' is not defined
```

Turns out that the Werkzeug reloader was starting one pdb process when the class got loaded at input time, and then another pdb process when I actually navigated to the form.  The two pdb processes competed over the input from `stdin`, which caused "list" to get processed as "lt."  Adding `FLASK_DEBUG: 0` stops the reloading and make it possible to place pdb breakpoints in classes.  I didn't see this bug when I first added the documentation about debugging, because then I only put pdb breakpoints in views, which don't get loaded at import time.

I will update the documentation and push to this branch.  If you'd prefer that I create another branch instead of pushing to a recently deleted branch, please let me know.  Thanks to both of you for your help with the documentation updates. :pray: 

*TLDR*: I need to update the documentation about debugging with an additional change to the `docker-compose.yml`.  I'm going to push the change to the branch redshiftzero just deleted and recreate the branch.  Please let me know if I should instead create a new branch.

Issue #589, opened 2018-12-14
Add "year" selector to browse officers/departments


camfassett:
As OpenOversight obtains new rosters for our departments in 2019, it's important that our historical data remain easily available to the public, so that changes over time in officer assignments and who is an active officer can be determined.

I think the best way to do this is to add a "year" selector to the top of the "browse officers" page for each department. 
<img width="1238" alt="screen shot 2018-12-14 at 13 14 10" src="https://user-images.githubusercontent.com/20425838/50019878-3db72e00-ffa2-11e8-99d7-0412a73c8884.png">

Additionally, perhaps there should be a simple way to view changes over time on each officer's profile as well? We have "assignment history", but what about the year that the officer left the department? Should we have a way to indicate the circumstances regarding a departure, such as a resignation? (Doesn't seem useful to have that be its own incident.)

@McEileen @redshiftzero particularly interested in your thoughts here.

McEileen:
@camfassett I'm going to work on this now because I think it ties into what I did in PR #636. I still think that we should only display "last employment date" and "last employment details/explanation" (I'm not particularly invested in what we call that field) if the fields are not null.  Your thoughts?

McEileen:
Also, what years do we want displayed as options for the year selector?  I'm thinking 2016 - 2019, since the project started in 2016.

camfassett:
Thanks @McEileen! That sounds good to me. I think "last employment details" for the latter field works. And yes, agreed re: those year selectors! This is perfect :)

PR #642, opened 2019-03-13
589/add year selector

### LINKED ISSUE #604: Add a "last employment date" field
When officers resign or quit, we should have a way of displaying that information. I think this could look like having two fields—last employment date field (which should default to N/A) and some sort of explanation/additional details field—i.e. if they resigned or were fired for some sort of misconduct issue.

McEileen:
## Status

Ready for review 
## Description of Changes

Fixes #604 and #589 .

 - Add `last_employment_date` and `last_employment_details` fields to all officers.
 - Add `last_employment_date` and `last_employment_details` fields to the add/edit officer forms.
 - Add a `year` dropdown selector to the page that displays all of a department's officers.

YearSelector
## Tests and linting

(There are two failing tests, `test_admin_can_disable_user` and `test_lastname_capitalization`, however, they were failing before this PR.
 - [x] `flake8` checks pass


dismantl:
I'm not understanding what the 'Year' dropdown in the filter section does. Does that mean year of employment?

McEileen:
Yes, so for example, if someone were trying to find an officer they interacted with in 2018, they could select "2018" in the dropdown.  @camfassett pointed out in this issue that as we obtain rosters for 2019, we want historic data to remain accessible to the public.  It's a good way to see how officer's assignments or roster sizes change over time.

McEileen:
Could someone please review this when they get a chance?  Thanks! :pray:

OpenOversight/app/main/forms.py
```diff
@@ -372,4 +378,6 @@ class BrowseForm(Form):
                           validators=[AnyOf(allowed_values(AGE_CHOICES))])
     max_age = SelectField('maximum age', default=100, choices=AGE_CHOICES,
                           validators=[AnyOf(allowed_values(AGE_CHOICES))])
+    year = SelectField('year', default='', choices=YEAR_CHOICES,
+                       validators=[AnyOf(allowed_values(YEAR_CHOICES))])
```
dismantl: I think it might be better to do a `QuerySelectField` that will get the full range of years from officer employment dates, rather than using a fixed list like `YEAR_CHOICES`. For one, departments could have employees with last employment dates before 2016. Plus, we'd have to update `YEAR_CHOICES` once every year going forward. A `QuerySelectField` is nice because it will dynamically fetch the possible choices from the database. I used one in PR #619, and set the query for the field in the corresponding view function, in case that is helpful at all.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": true
+}
```
dismantl: Not sure if it was intentional to include this file, since `.vscode/**` was added to `.gitignore`?

OpenOversight/app/utils.py
```diff
@@ -241,6 +243,13 @@ def filter_by_form(form, officer_query, is_browse_filter=False):
                                                     Officer.gender == 'Not Sure',
                                                     Officer.gender == None))  # noqa
 
+    if form['year']:
+        year = int(form['year'])
+        min_employment_date = datetime.date(year, 1, 1)
+        if year in (2019, 2018, 2017, 2016):
```
dismantl: Also here it would probably be better to query the range of possible dates from the database rather than a static list.

OpenOversight/migrations/versions/f01cb1bc6293_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message
```
dismantl: It's always helpful to put a description here and in the filename, but not required :)

OpenOversight/app/main/forms.py
McEileen: Good point, having to each year update the hardcoded list of `YEAR_CHOICES` is a downside.  I agree about using a `QuerySelectField`, but I'm not sure the best way to represent in the database the years for which we have available data.  Here are two approaches I thought of:
1) Create a `QuerySelectField` for each department that includes every year represented in every department officer's `employment_date` or `last_employment_date`.  (This is the approach I am leaning toward.)  Don't change the officer or department model at all.
2) Add a `Years_Of_Available_Data` field to each department and list in that field the years for which we have roster data....... Then, create the `QuerySelectField` based on this new field.  I  think this approach would be overengineering.

I'm also interested in what @camfassett thinks, as she opened the original issue.

.vscode/settings.json
McEileen: Not intentional, I think it was added before I added `.vscode/**` to the `.gitignore`.  Thanks for pointing it out.

OpenOversight/app/utils.py
McEileen: Agreed!

OpenOversight/app/main/forms.py
McEileen: Hey, I'm reviewing PRs that I still have pending, and I wanted to give a heads-up that I will be taking the first approach.

McEileen:
@r4v5 @redshiftzero Could one of you please review this PR?  @dismantl already reviewed it once and his main feedback was to remove the hardcoded lists of years and to instead use a `QuerySelectField`.  Thank you! :pray:

McEileen:
I found a bug in this PR where officers that started working for a department after the year selected would still appear in results.  For example, if a user selected "1970" from the drop-down selector, an officer that started working in "2005" would still appear in the results.
I will fix this bug and then ping someone for a re-review.

OpenOversight/app/utils.py
```diff
@@ -62,6 +62,14 @@ def dept_choices():
     return db.session.query(Department).all()
 
 
+def year_choices():
+    years = [year[0] for year in db.session.query(cast(func.extract('year', Officer.employment_date), Integer)).filter(Officer.employment_date.isnot(None)).order_by(func.extract('year', Officer.employment_date))]
+    years_deduped = list(set(years))
```
McEileen: I realize this isn't a super efficient way to dedupe the years.  Please let me know if anyone can think of a more performant approach.

brianmwaters:
Still working on my in-depth code review here, but I noticed in the meantime that the "year" selector only appears in the filter bar on the left-hand side of the officer listing page, but not on the "find an officer" page (e.g, `/find`). This seems to sort of make sense, but just want to verify that that's by design and not an oversight (ba-dum tiss)

OpenOversight/app/main/views.py
```diff
@@ -481,12 +476,15 @@ def list_officer(department_id, page=1, race=[], gender=[], rank=[], min_age='16
     if request.args.get('rank') and all(rank in rank_choices for rank in request.args.getlist('rank')):
         form_data['rank'] = request.args.getlist('rank')
 
+    if request.args.get('year') and request.args.get('year') in year_choices():
```
brianmwaters, line 479: It looks like this was already happening the code base before you added this line, but I've always learned to use `is not None` to check for `None` in Python, as opposed to relying on the falsy-ness of `None`. This is recommended by PEP 8 (see the 2nd bullet point under the heading): https://www.python.org/dev/peps/pep-0008/#programming-recommendations

OpenOversight/app/utils.py
brianmwaters: Could we just hardcode the years to be from like the year we founded OO or something (open to discussion how far back to go), to the current year?

OpenOversight/app/utils.py
```diff
@@ -300,6 +310,15 @@ def filter_by_form(form, officer_query, department_id=None):
         officer_query = officer_query.filter(db.or_(db.and_(Officer.birth_year <= min_birth_year,
                                                             Officer.birth_year >= max_birth_year),
                                                     Officer.birth_year == None))  # noqa
+    if form.get('year') and form.get('year') in year_choices():
```
brianmwaters, line 313: Same comment as above w/ regard to checking the truthyness of `None`

OpenOversight/tests/conftest.py
```diff
@@ -81,17 +82,29 @@ def pick_salary():
     return random.randint(100, 100000000) / 100
 
 
+def generate_random_date(year, month):
```
brianmwaters, line 85: Hmmmmm randomness in tests seems bad. What about picking a random number once and then hardcoding that into the tests?

brianmwaters: Unless, if this is standard practice in our test suite already, then I guess it's fine to keep doing it.

OpenOversight/tests/test_utils.py
```diff
@@ -118,6 +119,86 @@ def test_filter_by_partial_unique_internal_identifier_returns_officers(mockdata)
         assert returned_identifier == identifier
 
 
+def test_year_filter_selects_officer_who_quit_that_year(mockdata):
```
brianmwaters, line 122: Looks like you covered a lot of corner cases, nice job

OpenOversight/app/utils.py
McEileen: Even though the current code isn't ideal, I prefer it to hardcoding a list of years.  I think hardcoding is brittle and vulnerable to breakage.

McEileen:
> Still working on my in-depth code review here, but I noticed in the meantime that the "year" selector only appears in the filter bar on the left-hand side of the officer listing page, but not on the "find an officer" page (e.g, `/find`). This seems to sort of make sense, but just want to verify that that's by design and not an oversight (ba-dum tiss)

Yep, I only put the "year" selector on the browse page, since that's where the original issue, #589, requested it.  Still, thanks for opening (ba-dum tiss?) my eyes to this.

OpenOversight/app/main/forms.py
```diff
@@ -202,6 +202,9 @@ class AddOfficerForm(Form):
                             query_factory=unit_choices, get_label='descrip',
                             allow_blank=True, blank_text=u'None')
     employment_date = DateField('Employment Date', validators=[Optional()])
+    last_employment_date = DateField('Last Employment Date', validators=[Optional()])
+    last_employment_details = StringField('Last Employment Details', default='', validators=[
+        Regexp(r'\w*'), Length(max=50), Optional()])
```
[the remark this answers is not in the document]
McEileen: My first thought was, "shoot, you're right."  Then I tested this and was able to successfully enter last employment details with punctuation.

OOGeneralInformation

When I changed the validator from `Regexp(r'\w*')` to  `Regexp(r'\w')`, then I was no longer able to enter punctuation.  It appears that the wildcard adds flexibility to this regex.

OpenOversight/app/commands.py
```diff
@@ -313,9 +313,7 @@ def process_salary(row, officer, compare=False):
                 if Decimal('%.2f' % salary.salary) == Decimal('%.2f' % float(row['salary'])) and \
                         salary.year == int(row['salary_year']) and \
                         salary.is_fiscal_year == is_fiscal_year and \
-                        ((salary.overtime_pay and 'overtime_pay' in row and
-                            Decimal('%.2f' % salary.overtime_pay) == Decimal('%.2f' % float(row['overtime_pay']))) or
-                            (not salary.overtime_pay and ('overtime_pay' not in row or not row['overtime_pay']))):
+                        ((salary.overtime_pay and 'overtime_pay' in row and Decimal('%.2f' % salary.overtime_pay) == Decimal('%.2f' % float(row['overtime_pay']))) or (not salary.overtime_pay and ('overtime_pay' not in row or not row['overtime_pay']))):
```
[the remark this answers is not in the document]
McEileen: This only appears in the diff because of a formatting change.  Since I didn't write the logic, I'm not the right person to add an explanatory comment.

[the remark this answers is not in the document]
dismantl: This is comparing the imported salary with the officer's existing salaries, due to type and formatting differences.

OpenOversight/app/main/views.py
McEileen: TIL!

OpenOversight/tests/conftest.py
McEileen: I think it's okay because a lot of the other methods in `conftest.py`, like `pick_birth_date()` or `pick_gender()`, use `random.randint()` or `random.choice()`.  Let me know if you disagree.

McEileen:
I responded to feedback.  Here's [a video ](https://www.dropbox.com/s/hi1dvp1bohhrfdn/YearSelectorDemo.mp4?dl=0)of the feature in action.

OpenOversight/app/utils.py
abandoned-prototype: I am not sure I understand what exactly we want to get here. I think whatever years we offer to select should be an uninterrupted range. If it was 1990, 1995, 2003 and 2010 a user would probably be somewhat confused. If you want 2005, do you select 2003 or 2010? I would either hardcode the start year (maybe 2014 or something, I don't think we have much data from before that) and go until current year or take the lowest that's in the database and offer every year from that year on until current year

OpenOversight/app/main/forms.py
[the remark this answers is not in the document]
abandoned-prototype: The issue here is that `Regexp(r'\w*')` doesn't do anything. It matches literally any string. `Regexp(r'\w')` only fails if the first character of the string is punctuation. So the question here really is which strings should be allowed and which shouldn't?

OpenOversight/tests/conftest.py
abandoned-prototype: It is true that we do this in a lot of places and I think it is actually a problem. I saw an issue with the randomness of our test-data just yesterday.
But given that it is already so widespread I guess it's fine to keep doing it for now. And if it's just the actual value randomness is not as bad as deciding if a field is None or not based on randomness.

dismantl: PR #756 offers an alternate implementation that might have better random distribution since it hashes the officer name: https://github.com/lucyparsons/OpenOversight/pull/756/files#diff-4eef8ece740b94822be839e43d28d7a7R42

OpenOversight/app/main/forms.py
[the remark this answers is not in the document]
McEileen: I think my 2019 self used this regex because it's being used in most of the other string fields in `main/forms.py`.  Since @abandoned-prototype is right that it matches literally any string, I changed the regex to require alphanumeric characters in its input.  The field now allows punctuation, but will  throw an error if the input is only punctuation.

OpenOversight/app/utils.py
McEileen: My goal with `year_choices()` is to display every year for which a department has roster information.  
> take the lowest that's in the database and offer every year from that year on until current year

I went for this approach.

OpenOversight/tests/conftest.py
McEileen: > I saw an issue with the randomness of our test-data just yesterday.

What was the problem you saw with test data randomness?  Is it something that we need to open an issue for?

I can use the implementation from PR #756, but I would also like to know more about the problems with our existing use of random test data.

abandoned-prototype: I will open an issue on the concrete problem I saw and probably add some ideas I have for moving forward

McEileen:
I responded to feedback.  Thanks to everyone who offered their thoughts.

OpenOversight/tests/conftest.py
```diff
@@ -81,17 +81,40 @@ def pick_salary():
...
+
+    def bytes_to_float(b):
+        return float(unpack('L', sha256(b).digest()[:8])[0]) / 2 ** 64
+
+    if seed is None:
+        seed = str(uuid.uuid4()).encode('utf-8')
+
+    return datetime.datetime(start_year, 1, 1, 00, 00, 00) \
+        + datetime.timedelta(days=365 * (end_year - start_year) * bytes_to_float(seed))
+
```
abandoned-prototype, line 98: Nice approach. If we want to go with this I think we should put the `bytes_to_float` function in a utils-file are something so that we can use it everywhere, and I suggest rewriting the function as 
```
    def bytes_to_float(seed = None):
        if seed is None:
            seed = uuid.uuid4()
        return float(unpack('L', sha256(str(seed).encode('utf-8')).digest()[:8])[0]) / 2 ** 64
```
that way any object that has a `str` function can be used as seed.

if we want the result to be deterministic we have to make sure that the input is deterministic too. In this case it seems that the last name is used, which is unfortunately randomly chosen.

OpenOversight/app/utils.py
```diff
@@ -77,6 +77,16 @@ def dept_choices():
...
 
 
+def year_choices():
+    oldest_year = db.session.query(cast(func.extract('year', Officer.employment_date), Integer)) \
+                            .filter(Officer.employment_date.isnot(None)) \
+                            .order_by(func.extract('year', Officer.employment_date)) \
+                            .limit(1).first()[0]
+    current_year = datetime.datetime.now().year
+    year_choices = [str(year) for year in range(oldest_year, current_year + 1)]
+    return year_choices
```
abandoned-prototype, line 87: This looks good. I think for the oldest year we should use a `SELECT min(employment_date) FROM officers` to make it as efficient as possible. I looked up how to do that in sqlalchemy and it seems
```oldest_year = db.session.query(func.min(Officer.employment_date)).scalar().year```
should be do the trick.
What do we want to do in the case that there is no officer with a (not NULL) employment date?

File: OpenOversight/app/templates/officer.html [hunk 14]
```diff
@@ -70,6 +70,18 @@ <h3>General Information
...
               </tr>
+              {% if officer.last_employment_date %}
+                <tr>
+                  <td><b>Last Employment Date</b></td>
+                  <td>{{ officer.last_employment_date }}</td>
+                </tr>
+              {% endif %}
+              {% if officer.last_employment_details %}
+                <tr>
+                  <td><b>Last Employment Date Details</b></td>
```
abandoned-prototype, line 81: Can we change the label that we display to the users to something that is easier understood? Like "Details on Resignation" or "Details on End of Employment" or something? Not exactly sure how this field will be used though

OpenOversight/app/utils.py
abandoned-prototype: I keep coming back to this and I am think we should just hard-code the `oldest_year`. I think we now have some officers in the database whose employment_date is before 1950, which will make the drop-down menu pretty long and we really don't have good data for those year anyways. So my suggestion would be to hard-code it to 1980 or something, but I am also fine with the suggestion I made, if we can come up with a solution for an officer table without any non-NULL employment dates.

PR #643, opened 2019-03-14
write tests for uniq_internal_id, implement in utils and add frontend form portion

### LINKED ISSUE #633: Look up officers by unique internal identifier
For departments that use unique internal identifiers (e.g. "sequence numbers" in Baltimore PD), it would be nice to be able to look up officers based on that identifier. I would imagine this could just be an added field on Step 2 of the Find an Officer form.

McEileen:
## Status

Ready for review 

## Description of Changes

Fixes #633.

 - Add ability to search by unique internal identifier

UniqueInternalIdentifier

## Tests and linting

 - [x] `flake8` checks pass


McEileen:
@dismantl I wasn't sure what wording to use in the form, since different departments have different terms for `unique_internal_identifier`.  For example, NYPD has `tax_registry_number` and BPD has `sequence_number`.  Feedback on the form wording (and code, of course) would be appreciated!

dismantl:
@McEileen hmm yeah that's a good point. I suppose one way to deal with the different names for unique IDs would be to add a property to departments that is the name of the unique identifier, e.g. "sequence number". So basically just another text field to the `departments` table, and presumably a form field on the add/edit department pages. I know that's more work, but it's the best way I can think of to handle it. Thoughts?

McEileen:
@dismantl I like the idea you proposed and I don't think the change will require much work.  I do feel a little hesitant about adding a new database field solely for the purpose of having a clearer display.  On the flip side, we obviously want the site to be as user-friendly as possible.  I'm going to ask in slack if anyone has any other proposed solutions.  If we don't hear another idea in the next day, I'm going to implement your suggestion.

dismantl:
Another idea would be to use an i18n package and provide a 'translation' for each department.

OpenOversight/app/templates/input_find_officer.html
```diff
@@ -80,6 +80,15 @@ <h2><small>Do you remember any part of the Officer's badge number?</small></h2>
                   {% endfor %}
                 </div>
 
+                <h2><small>Do you know any part of the Officer's unique internal identifier?*</small></h2>
+                <h4><small>*This term varies by department, for example, in Baltimore it is the sequence number and in New York City it is the tax registry number.</small></h4>
+                <div class="input-group input-group-lg col-md-4 col-md-offset-4">
+                  {{ form.unique_internal_identifier(class="form-control") }}
+                  {% for error in form.badge.errors %}
```
dismantl: I think this should be `form.unique_internal_identifier.errors`, yes?

McEileen: Oy, yes, that should definitely be `form.unique_internal_identifier.errors`.  Thanks for pointing out the error.

McEileen:
Could someone please re-review this?  I now dynamically display the department-specific term for `unique internal identifier` in the find officer form.  The re-review shouldn't take long.  Thanks! :pray:

OpenOversight/app/models.py
```diff
@@ -32,10 +32,18 @@ class Department(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String(255), index=True, unique=True, nullable=False)
     short_name = db.Column(db.String(100), unique=False, nullable=False)
+    uii = db.Column(db.String(100), unique=False, nullable=True)
```
redshiftzero: it looks like this field is to denote that the department uses a unique internal ID in addition to badge number - how would this be set / updated? if the previous statmeent is correct, should it be a Boolean? 

Nit: can we name this something more expressive/explicit e.g. `is_using_unique_internal_id`

OpenOversight/tests/test_utils.py
```diff
@@ -94,13 +94,40 @@ def test_filter_by_badge_no(mockdata):
...
+    identifier = OpenOversight.app.models.Officer.query.first().unique_internal_identifier
+    partial_identifier = identifier[:len(identifier) // 2]
+    results = OpenOversight.app.utils.grab_officers(
+        {'race': 'Not Sure', 'gender': 'Not Sure', 'rank': 'Not Sure',
+         'min_age': 16, 'max_age': 85, 'name': '', 'badge': '',
+         'dept': department, 'unique_internal_identifier': partial_identifier}
+    )
+    for element in results:
+        returned_identifier = element.unique_internal_identifier
+        assert returned_identifier == identifier
```
redshiftzero: tests look :100:, very clear!

two other tests that could be added would be covering the following cases:
* department does have `uii` set, the `div` with `id="uii-question"` is present
* department does not have `uii` set, the `div` with `id="uii-question"` is not present

OpenOversight/app/models.py
McEileen: The field `uii` contains the exact term that the department uses for the unique internal id, for example, for NYPD this would be the tax registry number.  I created this field in order to display the appropriate term in step 2 of the Find Officer Form.  (At first I just asked for the "unique internal identifier," but I worried that would be unclear to users.)
I don't think the field should be a Boolean.  Do you want me to leave the field as it is, named `uii`, or do you still want something more explicit?  Perhaps `uii_term`, `unique_internal_identifier_label`, or `uii_phrase`?  I'm fine with leaving it as `uii`, but let me know what you think.

OpenOversight/tests/test_utils.py
McEileen: Sure thing, I'll implement those tests.  Are there any similar tests that you'd suggest I look at as a model to follow?  I initially tried writing the below
```python3
def test_find_officer_cannot_see_uii_question_for_depts_without_uiis(mockdata, client, session):
    with current_app.test_request_context():
        dept_without_uii = Department.query.filter_by(uii=None).one_or_none()
        form = FindOfficerForm(dept=dept_without_uii)

        rv = client.post(
            url_for('main.get_officer'),
            data=form.data,
            follow_redirects=True
        )
        assert not "Do you know any part of the Officer's" in rv.data.decode('utf-8')
```
The test doesn't work, because even when the form gets filled out with a department that has no uii, `rv.data.decode('utf-8')` includes the uii question.  I think this is a reflection of how the template always contains that question, but the `complaint.js` script hides the question depending on the department id. 
(Also, I might try testing for if the `<div id=uii-question>` is hidden, rather than testing for the question wording.  I realize testing for specific phrases can get brittle.)

dismantl: Does the javascript hide the `#uii-question` div based on the department selection? If so you'll probably need to use selenium to test whether the div gets hidden since that uses an actual browser, and the response you store in the `rv` variable above is just the basic HTML, without the javascript being evaluated.

OpenOversight/app/models.py
dismantl: IMO keeping the field as is and renaming it to `unique_internal_identifier_label` makes the most sense. No need for a separate boolean field when checking for a null value in the existing column would work just as well.

OpenOversight/tests/test_utils.py
McEileen: Thank you!  I wrote two tests using selenium and have them passing.  I'll push the changes tonight.

McEileen:
Update on the changes: `Flake8` now gives me this error in response to the functional test for departments that have uuis: `OpenOversight/tests/test_functional.py:140:60: E711 comparison to None should be 'if cond is not None:'`.  I think the linter is misreading the query as a conditional statement, and I think what I have is fine.  If anyone disagrees, please let me know.
Also, I pulled from `origin develop` and then rebased my local `633/search-internal-id` branch on `develop` before doing a force push.

redshiftzero:
hey thanks @McEileen this looks good now - just needs a rebase + conflicts resolving (`down_revision` in migrations also should be updated to the short ID of the latest migration on `develop`)

dismantl:
Works great for me! :+1:

Issue #611, opened 2018-12-29
Unable to remove links from officer profiles


dismantl:
After adding some links to an officer profile, I go to the edit officer page and remove the links and click Update. The action appears to be successful, yet the links remain on the officer profile.

dismantl:
If we move the links form to its own page (a la #574) and add the little edit and delete buttons to each entry (like Notes and Descriptions), that would probably be a good way to solve this.

redshiftzero:
closed by #612

dismantl:
#612 failed to resolve this, so reopening.

abandoned-prototype:
FIxed by https://github.com/lucyparsons/OpenOversight/pull/646

PR #646, opened 2019-03-18
Refactoring links forms

### LINKED ISSUE #574: Add Link button on officer profile page
Currently to add a link to an officer profile, you have to edit the whole profile and fill out the Links form at the bottom. A new user might not even know this functionality exists.

IMO it would make more sense for their to be a "New Link" button on the profile page similar to "New Note" and "New Description".

dismantl:
## Status

Ready for review

## Description of Changes

Fixes #574, #611.

 - add New Link/Video button to officer profiles
 - add edit and delete buttons to links on officer profiles
 - add LinksApi for basic CRUD operations on links, including html templates
 - remove links subform from edit officer page
 - change user_id to creator_id in links table
 - add links to officer profiles in test data

image

image

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/models.py
```diff
@@ -18,10 +18,6 @@
 db = SQLAlchemy()
 
 
-officer_links = db.Table('officer_links',
```
redshiftzero: so as background, the motivation for this table was to prevent data duplication (similar concerns as https://github.com/lucyparsons/OpenOversight/pull/619#discussion_r275175521) in the case where one link relates to multiple officers - let me know if I misunderstand the intention here

dismantl: That makes sense and I can leave the officer_links table as is. Just a note tho that we currently don't have any uniqueness checks on links, so I assume it's okay to leave it that way?

dismantl:
Restored the `officer_links` table and rebased on develop. Ready for review again.

redshiftzero:
status of this review: rebased on latest `develop` but it looks like when links are created via an officer's profile the link is indeed created but the `officer_id, link_id` isn't being created. Thus when you create the link you see a flashed message indicating the link now exists but one does not see the link appear on the officer profile.

dismantl:
Ready for review. This PR has been rebased on develop and the issue you mention @redshiftzero should be fixed. I've also added relevant unit tests.

OpenOversight/app/main/views.py
```diff
@@ -1304,3 +1300,99 @@ def dispatch_request(self, *args, **kwargs):
     '/officer/<int:officer_id>/description/<int:obj_id>/delete',
     view_func=description_view,
     methods=['GET', 'POST'])
+
+
+# This API only applies to links attached to officer profiles, not links
+# attached to incidents.
```
fritzdavenport, line 1306: maybe move the code comment as a doc comment inside the function?

OpenOversight/app/main/views.py
```diff
@@ -1304,3 +1300,99 @@ def dispatch_request(self, *args, **kwargs):
...
+    @ac_or_admin_required
+    def new(self, form=None):
+        if not current_user.is_administrator and current_user.ac_department_id != self.officer.department_id:
+            abort(403)
+        if not form:
+            form = self.get_new_form()
+            if hasattr(form, 'creator_id') and not form.creator_id.data:
+                form.creator_id.data = current_user.get_id()
+
+        if form.validate_on_submit():
```
fritzdavenport, line 1329: Can you add an `else` to catch if there are errors and log them, in the spirit of https://github.com/lucyparsons/OpenOversight/issues/786 ?

dismantl: The ModelView class it inherits from is written in the same way, so I wouldn't want to break that pattern here without doing it elsewhere, to keep things consistent. And I think it would be best to come to consensus on a solution for #786 and apply it in a separate PR.

fritzdavenport (dismissed):
LGTM. Code looks good, and poking around right now in the interface. Looks good to both add, remove, and with none/one/many. Add Link+Video is a little weird when they are two different categories, but I think I don't hate the fact that the add button is conjoined. I gave to my partner to poke around too and got a :+1: from them :) 

Improvement request: Can we do more in-line form editing rather than as a separate page? That's maybe more of a general thing if nothing else really has that.

dismantl:
> LGTM. Code looks good, and poking around right now in the interface. Looks good to both add, remove, and with none/one/many. Add Link+Video is a little weird when they are two different categories, but I think I don't hate the fact that the add button is conjoined. I gave to my partner to poke around too and got a 👍 from them :)
> 
> Improvement request: Can we do more in-line form editing rather than as a separate page? That's maybe more of a general thing if nothing else really has that.

Yes I like that idea for inline forms rather than separate pages! Thanks for the review @fritzdavenport!

abandoned-prototype (approved):
Thanks @dismantl this looks great. Sorry, I seemed to have missed this PR earlier.
One minor thing: Since you changed the field that the advanced-csv-import expects for links, could you also update the field in the documentation: https://github.com/lucyparsons/OpenOversight/blob/develop/docs/advanced_csv_import.rst#links-csv?

abandoned-prototype (approved):
Re-approving. And thanks for adding the documentation-changes @dismantl!

PR #650, opened 2019-03-21
Decode generated tokens to fix environment-specific bug

dismantl:
## Status

Ready for review. Big thanks to @McEileen for helping to troubleshoot this!

## Description of Changes

This fixes a weird bug we were seeing that seemed to only occur in Linux environments. As far as I can tell, the generated tokens used in the user registration system were being deserialized as bytes rather than strings in some environments, causing 3 tests to fail. This fixes that by ensuring generated tokens are deserialized and decoded into utf-8 strings.

## Tests and linting

 - [x] `flake8` checks pass


McEileen:
RIP to a very strange bug.

PR #412, opened 2018-03-24
WIP: photo conversion (Supercedes #404)

ssempervirens:
## Status

In progress, still need to resolve tests and write unit tests before merge 

## Description of Changes

Fixes #369. Supersedes PR #404 

Changes proposed in this pull request:

 - Converts tiff and psd file submissions to jpegs

## Tests and linting
 
 [x] I have rebased my changes on current `develop`
 
 [] pytests pass in the development environment on my local machine
 
 [] `flake8` checks pass

PR #691, opened 2019-10-15
converts bad photo file extensions (supercedes 404/412)

### LINKED ISSUE #369: Allow upload of additional image formats

ssempervirens:
Resolves #369 and supercedes PRs #404 and #412. I made a real mess of this issue, and I figured a new branch is the cleanest fix since that was a while ago.

## Status

WIP

## Description of Changes

Fixes # 369.

 - Convert psd and tiff files to jpeg
 - Test that tiff and psd files can be uploaded
- Note: avoided using os.open, I don't believe this is necessary and avoids dealing with safety warnings

## Tests and linting

*Note*: I get 'TypeError: attrib() got an unexpected keyword argument 'convert' when I `make test`. This seems to be an issue with the package attrs rather than anything I changed, but please let me know if this is not correct. 

 - [ ] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -189,6 +189,21 @@ def allowed_file(filename):
            filename.rsplit('.', 1)[1].lower() in current_app.config['ALLOWED_EXTENSIONS']
 
 
+def requires_conversion(filename):
+    return '.' in filename and \
+        filename.split('.', 1)[1].lower() in current_app.config['PHOTO_REQUIRES_CONV']
+
+
+def convert_image(filename):
+    try:
+        img_to_change = Pimage.open(filename)
+    except Pimage.DecompressionBombWarning:
+        return None
+    basename = os.path.splitext(filename)[0].split('.')
+    fixed_img = img_to_change.save(basename + 'jpeg')
+    return fixed_img
+
+
 def get_random_image(image_query):
     if image_query.count() > 0:
         rand = random.randrange(0, image_query.count())
--- a/OpenOversight/app/config.py
+++ b/OpenOversight/app/config.py
@@ -37,7 +37,8 @@ class BaseConfig(object):
 
     # Upload Settings
     MAX_CONTENT_LENGTH = 50 * 1024 * 1024
-    ALLOWED_EXTENSIONS = set(['png', 'jpg', 'jpeg', 'gif'])
+    ALLOWED_EXTENSIONS = set(['png', 'jpg', 'jpeg', 'gif', 'psd', 'tiff'])
+    PHOTO_REQUIRES_CONV = set(['tiff', 'psd'])
 
     # User settings
     APPROVE_REGISTRATIONS = os.environ.get('APPROVE_REGISTRATIONS', False)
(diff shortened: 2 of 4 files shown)


OpenOversight/app/utils.py
```diff
@@ -189,6 +189,17 @@ def allowed_file(filename):
            filename.rsplit('.', 1)[1].lower() in current_app.config['ALLOWED_EXTENSIONS']
 
 
+def requires_conversion(filename):
+    return '.' in filename and \
+        filename.split('.', 1)[1].lower() in current_app.config['PHOTO_REQUIRES_CONV']
+
+
+def convert_image(filename):
```
redshiftzero: does this _convert_ e.g. tiff and psd files to jpegs or just rename them? we can use imagemagick for the conversion, i think it should work for both tiff and psds (though we should test that, each are good candidates for unit tests)

ssempervirens: Right, so I didn't realize imagemagick had a python binding? Should I just use PIL? 

(So there is no way to avoid the os.open/decompression bomb issue?)

redshiftzero: oh right right, Pillow (Python 3 version of PIL) should do the trick! there's also a decent imagemagick API called `wand` if you find that Pillow is lacking support for a filetype that you want to support.

But yeah I'm not seeing a way to convert these images without using something like Pillow (short of using a third-party service).

McEileen:
Hey @ssempervirens, I'm following up about the error message you got, `TypeError: attrib() got an unexpected keyword argument 'convert' `, after you ran `make test`.  I didn't hit this error when I ran the tests on my own machine, but I did see it in the [Travis build for another PR](https://travis-ci.org/lucyparsons/OpenOversight/jobs/600209495?utm_medium=notification&utm_source=github_status).  I [read up on stackoverflow](https://stackoverflow.com/questions/58189683/typeerror-attrib-got-an-unexpected-keyword-argument-convert) and realized that the error was caused by OO using an older version of pytest, 4.0.2.  (I had a more recent version of pytest installed on my local machine, and I think that's why I didn't get this error locally.)

I updated pytest to 5.2.2 in `dev-requirements.txt` and it fixed the error for me.  Let me know if this fixes the error for you, too.  

Also, since this PR is superseding #404 and #412, could you please close those other PRs?  Thanks!

OpenOversight/app/utils.py
brianmwaters: According to the Pillow docs for this method (https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.save),

"Saves this image under the given filename. If no format is specified, the format to use is determined from the filename extension, if possible."

So it would appear this is actually converting the files to the new format, not just changing the name. 

However, maybe we should have a test for this if we don't already?

brianmwaters:
For what it's worth, I'm seeing the same thing as @McEileen. Namely, I was getting `TypeError: attrib() got an unexpected keyword argument 'convert'`, when I ran `make test` on @ssempervirens' `369-convert-photo-extension` branch. However, when I rebased this branch onto the current `develop`, which contains @McEileen's patch bumping the pytest version, it worked.

I think if we ran this through Travis CI again, the tests would probably pass.

PR #703, opened 2020-02-28
Changes needed for Vermont OO

brianmwaters:
## Status

In progress

## Description of Changes

Adds three new flags to the `bulk_add_officers` command, and make one change to default behavior (because the change is minimal and a fourth flag feels unwieldy: 

- Changes to `bulk_add_officers`:
  1) Add a flag --update-by-name to `bulk_add_officers`, which allows officers to be updated by name (without star_no or unique_internal_identifier) within a department.
  2) Add a flag --update-static-fields to `bulk_add_officers`, which allows normally-static fields like race, employment date, etc. to be updated.
  3) Unconditionally relaxes a check on static fields in `bulk_add_officers`, so that if a static field is `None`, and an imported CSV contains data for it, it will be updated, regardless of whether --update-static-fields is passed. This seems sensible and does not break any unit tests.
  4) Add a flag --no-create to `bulk_add_officers`, which causes `bulk_add_officers` to never create new officers, and only update existing ones. Any unknown officers in the imported CSV will simply be ignored.
  5) Adds a mode to the `get_officers` function, whereby if `star_no` is passed as `None`, `get_officers` just returns the first officer with the given first and last name. (This is required for --update-by-name.)
- Add `add_department` and `add_job_title` commands.
- Make changes to templates so that OO can show officers without any assignments without throwing errors.

## Notes for Deployment

I have not added unit tests and documentation, but I intend to.

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/commands.py
```diff
@@ -362,26 +372,61 @@ def bulk_add_officers(filename):
...
+
+
+@click.command()
+@click.argument('department_id')
+@click.argument('job_title')
+@click.argument('is_sworn_officer')
+@click.argument('order')
+@with_appcontext
+def add_job_title(department_id, job_title, is_sworn_officer, order):
+    """Add ranks from a CSV file."""
```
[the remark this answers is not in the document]
brianmwaters: whoops, good catch

brianmwaters:
Linking to https://github.com/brianmwaters/OpenOversight/pull/1, since @abandoned-prototype is working on getting these tests written and has already written a bunch.

OpenOversight/app/commands.py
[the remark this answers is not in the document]
brianmwaters: Fixed

test_data.py
```diff
@@ -22,7 +22,7 @@
 
     if args.populate:
         print("[*] Populating database with test data...")
-        mockdata(db.session)
+        add_mockdata(db.session)
```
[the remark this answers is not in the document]
brianmwaters: Done

OpenOversight/tests/test_commands.py
```diff
@@ -79,66 +106,639 @@ def test_add_job_title__success(session):
...
+    Assignment.query.delete()
+    Officer.query.delete()
+
+    assert Officer.query.count() == 0
+
+    df = pd.read_csv(csvfile)
+    df.loc[0, "job_title"] = "Commander"
+    df.to_csv(csvfile)
+
+    n_created, n_updated = bulk_add_officers([csvfile], standalone_mode=False)
```
brianmwaters, line 275: @abandoned-prototype what is standalone mode? I don't see this argument in the definition of `bulk_add_officers`

OpenOversight/tests/test_commands.py
```diff
@@ -79,66 +106,639 @@ def test_add_job_title__success(session):
...
+    df.loc[0, "job_title"] = "CAPTAIN"
+    df.to_csv(csvfile)
+
+    n_created, n_updated = bulk_add_officers([csvfile], standalone_mode=False)
+    assert n_created == 0
+    assert n_updated == 1
+
+    officer = Officer.query.filter_by(id=officer_id).one()
+    assert len(list(officer.assignments)) == 2
+    for assignment in officer.assignments:
```
brianmwaters, line 297: I see what you are doing w/ this loop, but it's not strict enough. If both assignments were "CAPTAIN", or both assignments were "Commander", the test would pass. It should check that there's one of each.

brianmwaters: Side note, I may have a pull request coming soon that would be relevant to this. But, this test can stay as-is (aside from the comment about the for loop above), I'll submit a new test along with it. I will make it a separate PR so that we can get this one finished with.

OpenOversight/tests/test_commands.py
```diff
@@ -79,66 +106,639 @@ def test_add_job_title__success(session):
...
+
+    assert Officer.query.count() > 0
+    n_created, n_updated = bulk_add_officers([csvfile], standalone_mode=False)
+    assert n_created == 0
+    assert n_updated == 1
+    assert Officer.query.count() == officer_count
+
+    officer = Officer.query.filter_by(id=officer_id).one()
+    assert len(list(officer.salaries)) == 2
+    for salary in officer.salaries:
```
brianmwaters, line 401: This loop uses the same pattern, and has the same problem, as the one that was commented above. It should be re-written to be a bit more strict

OpenOversight/tests/test_commands.py
```diff
@@ -79,66 +106,639 @@ def test_add_job_title__success(session):
...
+        assert float(salary.salary) == 123456.78 or float(salary.salary) == 150000.00
+
+
+# NEW TESTS
```
brianmwaters, line 405: This comment was helpful for code review but we should probably delete it before merging because these tests won't be "new" for very long!

OpenOversight/app/commands.py
```diff
@@ -362,26 +372,61 @@ def bulk_add_officers(filename):
...
+                else:
+                    raise Exception('Officer {} {} missing badge number and unique identifier'.format(row['first_name'],
+                                                                                                      row['last_name']))
+            else:
                 officer = Officer.query.filter_by(
                     department_id=department_id,
-                    unique_internal_identifier=row['unique_internal_identifier']
+                    last_name=row['last_name'],
+                    first_name=row['first_name']
                 ).one_or_none()
```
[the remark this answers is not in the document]
brianmwaters: @abandoned-prototype wrote a test for this (https://github.com/brianmwaters/OpenOversight/blob/2431f8e62687914d6fbec8228eea40a2426c0d4c/OpenOversight/tests/test_commands.py#L549), and it is passing. The docs for `one_or_none()` (https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.one_or_none) also specify that `MultipleResultsFound` is thrown in this case. So I'd say this is resolved.

OpenOversight/tests/test_commands.py
brianmwaters: Turns out this is an issue with code that was already in the codebase before this PR. It is showing up in the commit because the code was moved into a new file. I am marking this issue "resolved." @abandoned-prototype has indicated in Slack he might submit a separate PR for this next week. I also am happy to do it.

OpenOversight/tests/test_commands.py
brianmwaters: Turns out this is an issue with code that was already in the codebase before this PR. It is showing up in the commit because the code was moved into a new file. I am marking this issue "resolved." @abandoned-prototype has indicated in Slack he might submit a separate PR for this next week. I also am happy to do it.

OpenOversight/tests/test_commands.py
brianmwaters: Turns out this is an issue with code that was already in the codebase before this PR. It is showing up in the commit because the code was moved into a new file. I am marking this issue "resolved." @abandoned-prototype has indicated in Slack he might submit a separate PR for this next week. I also am happy to do it.

OpenOversight/tests/test_commands.py
```diff
@@ -79,66 +106,639 @@ def test_add_job_title__success(session):
...
+# --update_by_name && two with same name
+# --update_by_name && success
+# update null static field with and without --update-static-fields
+
+
+def test_bulk_add_officers__success(session, department_with_ranks, csv_path):
```
brianmwaters, line 411: This test looks like it is for the base case, and seems redundant with the `test_csv_import` tests, but not necessarily problematic. @abandoned-prototype is there something this test does that the older ones don't?

brianmwaters:
Okay!!

I went through @abandoned-prototype's tests. They look good, except that there's one for which I'm not sure what it brings to the table (some older tests seem to be testing the same thing, but maybe I'm wrong). I also made sure all of @redshiftzero's comments from earlier have been addressed. I think we should wait to hear from @abandoned-prototype about that one test, and then if we can get a final sign-off, then I think we can merge this.

OpenOversight/tests/test_commands.py
abandoned-prototype: This test is for the base-case aka "happy path" of using the flag `--update-by-name`. This would fail if the flag wasn't set, so I don't think there is any other test that tests this scenario

OpenOversight/tests/test_commands.py
abandoned-prototype: oops. I put this in as a quick note to myself and then forgot to take them out. Thanks for fixing this!

OpenOversight/tests/conftest.py
```diff
@@ -432,6 +432,33 @@ def mockdata(session):
...
+            is_sworn_officer=True,
+            department=department
+        ))
+    session.commit()
+    return department
+
+
+@pytest.fixture
+def csv_path(tmp_path):
+    return os.path.join(tmp_path, "file.csv")
```
dismantl, line 459: I had to change this to `return os.path.join(str(tmp_path), "file.csv")` or else I would get the following error: `TypeError: join() argument must be str or bytes, not 'PosixPath'`.

dismantl:
@brianmwaters do you have some example CSV files I could use to try out the new command flags? Or if you have other guidance for how to review this PR let me know.

OpenOversight/tests/conftest.py
brianmwaters: This sounds like maybe an issue w/ the Python version, which me and @abandoned-prototype hadn't caught before. I'll change it to your version and see if all the tests pass locally and on Travis, and if so I guess we can keep it.

brianmwaters:
> @brianmwaters do you have some example CSV files I could use to try out the new command flags? Or if you have other guidance for how to review this PR let me know.

@dismantl check out https://github.com/brianmwaters/vt-police-tools, which are our data-cleaning scripts. See `data/README.md` for documentation of the data itself, and also `scripts/migrations/0000-initial.sh`, which cleans the data on-the-fly and pipes it right into the OO import commands. Let me know if anything is confusing, there are probably a few things in there that could use more documentation/comments.

OpenOversight/tests/conftest.py
abandoned-prototype: Looks like the tests are also failing in travis for python 3.5. @brianmwaters can you add the `str` as dismantl suggested?

OpenOversight/app/templates/list_officer.html
```diff
@@ -145,13 +149,13 @@ <h3 class="panel-title accordion-toggle">Age range</h3>
               <div class="col-md-6 col-xs-12">
                 <h2>
                   <a href="{{ url_for('main.officer_profile', officer_id=officer.id) }}" id="officer-profile-{{ loop.index }}">{% include "partials/officer_name.html" %}</a>
-                  <small>#{{ assignment.star_no }}</small>
+                  <small>{{ star_no }}</small>
```
dismantl, line 152: If the `{% if officer.assignments.all() %}` clause evals as false, `star_no` won't get set so won't this throw an error? If so it might need to change to `<small>{{ star_no|default('') }}</small>`.

dismantl: Nevermind, I tried it out and it didn't throw an error.

dismantl:
This PR looks good to me, aside from the one change to conftest.py above. Once that's resolved, should be good to merge.

brianmwaters:
@dismantl okay, change made. Tests are passing on 3.7 locally, and on 3.5, 3.6, and 3.7 on Travis

PR #705, opened 2020-03-28
refactor template for add/edit department

### LINKED ISSUE #704: The page to add/edit department ranks throws a 500 on staging
The page to add or edit department ranks currently throws a 500 on staging.  Here's the error message from logs:
```bash
File "/home/nginx/oovirtenv/venv/OpenOversight/OpenOversight/app/templates/add_edit_department.html", line 28, in block "content"
{% for subfield in (form.jobs|rejectattr('data.is_sworn_officer','eq',False)|sort(attribute='data.order')|list)[1:] %}
File "/home/nginx/oovirtenv/venv/lib/python3.5/site-packages/jinja2/filters.py", line 278, in do_sort
return sorted(value, key=key_func, reverse=reverse)
TypeError: unorderable types: NoneType() < NoneType()
```

The error occurs because [this line ](https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/templates/add_edit_department.html#L28) in the template attempts to sort the different jobs by order.  However, none of the jobs in the staging database, as of March 2020, had any order.  Additionally, [the migration that originally created the jobs table ](https://github.com/lucyparsons/OpenOversight/pull/619/files#diff-501b654b0683fbebfa84c9ea22300d17R36)didn't populate the order field.

Proposed fix: remove the sort by order and keep everything else in the template the same.

McEileen:
## Status

Ready for review

## Description of Changes

Fixes #704 .

 - Remove the code that attempts to sort department ranks by their order, because none of the ranks currently in the staging database have a value in the `order` column.  As a result, this page was throwing a 500.

Here's how the page appears for Chicago after I removed the sort by order.  (There are multiple screenshots because this page included so many ranks.)
1.
DepartmentRanks1

2.
DepartmentRanks2

3.
DepartmentRanks3

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/commands.py
+++ b/OpenOversight/app/commands.py
@@ -266,7 +266,7 @@ def process_assignment(row, officer, compare=False):
                 # create new job
                 job = Job(
                     is_sworn_officer=False,
-                    department_id=officer.id,
+                    department_id=officer.department_id,
                     order=auto_order
                 )
                 set_field_from_row(row, job, 'job_title', allow_blank=False)
--- a/OpenOversight/app/commands.py
+++ b/OpenOversight/app/commands.py
@@ -258,10 +258,16 @@ def process_assignment(row, officer, compare=False):
                                 department_id=officer.department_id)\
                      .one_or_none()
             if not job:
+                num_existing_ranks = len(Job.query.filter_by(department_id=officer.department_id).all())
+                if num_existing_ranks > 0:
+                    auto_order = num_existing_ranks + 1
+                else:
+                    auto_order = 0
                 # create new job
                 job = Job(
                     is_sworn_officer=False,
-                    department_id=officer.id
+                    department_id=officer.id,
+                    order=auto_order
                 )
                 set_field_from_row(row, job, 'job_title', allow_blank=False)
                 db.session.add(job)
--- a/OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
+++ b/OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
@@ -29,13 +29,17 @@ def upgrade():
     )
 
     op.drop_constraint('unique_department_job_order', 'jobs', type_='unique')
+    op.drop_index(op.f('ix_jobs_order'), table_name='jobs')
 
     all_depts = connection.execute(department.select()).fetchall()
     for dept in all_depts:
-        job_order = 0
-        connection.execute(
-            job.update().values({"order" : job_order}).where(sa.and_(job.c.department_id == dept.id, job.c.order.is_(None)))
-        )
+        dept_jobs = connection.execute(department.select().where(job.c.department_id == dept.id)).fetchall()
+        for dept_job in dept_jobs:
+            job_order = 0
+            connection.execute(
+                job.update().values({"order" : job_order}).where(sa.and_(dept_job.c.department_id == dept.id, dept_job.c.order.is_(None)))
+            )
+            job_order += 1
 
     op.alter_column('jobs', 'order',
                     existing_type=sa.INTEGER(),
@@ -43,7 +47,7 @@ def upgrade():
 
 
 def downgrade():
-    op.create_unique_constraint('unique_department_job_order', 'jobs', ['order', 'department_id'])
+    # op.create_unique_constraint('unique_department_job_order', 'jobs', ['order', 'department_id'])
 
     op.alter_column('jobs', 'order',
                     existing_type=sa.INTEGER(),
--- a/OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
+++ b/OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
@@ -7,8 +7,7 @@ Create Date: 2020-04-24 01:58:05.146902
 """
 from alembic import op
 import sqlalchemy as sa
-from app.models import Department, Job, db
-
+from sqlalchemy.sql import table, column
 
 # revision identifiers, used by Alembic.
 revision = '562bd5f1bc1f'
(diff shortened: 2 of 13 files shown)


McEileen:
Hi, could someone please review this PR?  It's blocking deploying staging to production.  If someone prefers another approach to resolving the 500, let me know.

redshiftzero:
Two quick notes:
1. CI didn't run here but a repush _should_ do the trick (I pushed a branch to test and CI builds are getting kicked off as expected, could have been a temporary issue with webhook delivery)
2. Regarding the diff, it seems like adding a data migration (using alembic) to populate the order column (even with 0) and make the column not nullable (since this view assumes the column is populated) is the right resolution (that way we preserve the functionality to reorder ranks in this view). What do you think?

McEileen:
> 2. Regarding the diff, it seems like adding a data migration (using alembic) to populate the order column (even with 0) and make the column not nullable (since this view assumes the column is populated) is the right resolution (that way we preserve the functionality to reorder ranks in this view). What do you think?

I think this is reasonable.  Do you want me to do that before we deploy staging to prod?  Or do you want me to deploy staging to prod with the quick fix that I suggested in this PR, and then make the more lasting fix for our next deploy?  At the moment, I don't have a strong preference.

redshiftzero:
So, I think we should do the data migration before going to prod as otherwise deploying this can cause data pollution (i.e. if someone were to use this view). I think it should be a small migration, but if you want to do this in another PR (to merge this you'll need to rebase and force push to this branch to retrigger CI) or in this PR is up to you.

McEileen:
> So, I think we should do the data migration before going to prod as otherwise deploying this can cause data pollution (i.e. if someone were to use this view).

That makes sense!  I'll do the data migration before the deploy.

McEileen:
@redshiftzero  @r4v5 (And anyone else who is up for doing a review) This PR is ready for re-review now.  I wrote the necessary migration and locally tested that an admin can add or reorder ranks through the edit department page.

OpenOversight/app/commands.py
```diff
@@ -258,10 +258,16 @@ def process_assignment(row, officer, compare=False):
...
+                num_existing_ranks = len(Job.query.filter_by(department_id=officer.department_id).all())
+                if num_existing_ranks > 0:
+                    auto_order = num_existing_ranks + 1
+                else:
+                    auto_order = 0
                 # create new job
                 job = Job(
                     is_sworn_officer=False,
-                    department_id=officer.id
+                    department_id=officer.id,
```
redshiftzero, line 269: unrelated to this diff, but this just jumped out at me: is this a bug? should this be `department_id=officer.department_id`?

OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
```diff
@@ -0,0 +1,40 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_constraint('unique_department_job_order', 'jobs', type_='unique')
+
+    for department in Department.query.all():
+        job_order = 0
+        for job in Job.query.filter_by(department_id=department.id):
```
redshiftzero, line 25: this looks great! one thought: here should we also filter by those jobs that don't have a value for `order` yet? (in case someone running this migration on their data already did define `order` for jobs for a department)

requirements.txt
```diff
@@ -17,3 +17,4 @@ Fabric3==1.14.post1
 itsdangerous==0.24
 us>=1.0
 future
+email-validator==1.0.5
```
redshiftzero, line 20: ah hmm where do we use this dependency?

redshiftzero: Ahhh I see, this is brought in by latest WTForms now (I noticed via CI failing on #706). I'm with ya now

OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
McEileen: That makes sense! :+1: I will add that filter.

OpenOversight/app/commands.py
McEileen: I think you're right.  I'll change it to `department_id=officer.department_id`.

OpenOversight/migrations/versions/562bd5f1bc1f_add_order_to_jobs.py
McEileen: Hey @redshiftzero, I'm having a hard time filtering for jobs where `order` is null.  I tried to write a query using `filter` instead of `filter_by`.  I tested it out in `test_utils.py` with `test_find_jobs_without_order`, and it seemed to work.

But, when I run `flask db upgrade`, it hangs after this line:

`INFO  [alembic.runtime.migration] Running upgrade 6045f42587ec -> 562bd5f1bc1f, Add order to jobs in DB and adjust constraints on order column` 

I tried to see if there are any locks on the database

`SELECT (locktype, pid, mode, granted) FROM pg_locks ORDER BY pid;`

and found this process where `granted` was false 

` (relation,83,AccessShareLock,f)`


I checked `pg_stat_activity` next

`SELECT (pid, application_name, state, query) FROM pg_stat_activity WHERE pid=83;`
` (83,"",active,"SELECT jobs.id AS jobs_id, jobs.job_title AS jobs_job_title, jobs.is_sworn_officer AS jobs_is_sworn_officer, jobs.""order"" AS jobs_order, jobs.department_id AS jobs_department_id +
 FROM jobs                                                                                                                                                                                          +
 WHERE jobs.department_id = 1 AND jobs.""order"" IS NULL")`




It seems to be locked on this query?  I'm not sure why. 


Then I investigated another pid that appeared in the `pg_locks` table.  It seems that the process to remove the unique constraint on job order is idle?

`SELECT (pid, application_name, state, query) FROM pg_stat_activity WHERE pid=82;`
 `(82,"","idle in transaction","ALTER TABLE jobs DROP CONSTRAINT unique_department_job_order")`


I'm having a hard time figuring out why the migration isn't completing.  Any help would be appreciated.  Thanks!

McEileen: Update: after much investigating and duckduckgoing, I figured out that the migration was hanging due to my use of `db.session.commit()`.  I wrote a new migration that uses `connection` instead of `session` and populates the order column for unordered ranks with 0. 

I was able to successfully run this on a mockdata department that had unordered ranks, then locally view the ranks and update them on the edit department page.  However, when I tried to run `make test`, it failed, because the department with unordered ranks broke the new not-null constraint on department order.  Consequently, I added rank order to the department in order to make the tests pass.

Let me know if you have any questions.

PR #727, opened 2020-06-11
persist deleting rank on department edit page

### LINKED ISSUE #708: Bug: Deleting a rank through the department /edit page doesn't persist
I found this bug while doing testing before releasing to prod for #664.  If a user deletes a rank on the department `edit` page, the change doesn't persist.  The relevant code is linked [here](https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/main/views.py#L408).

McEileen:
## Status

Ready for review

## Description of Changes

Fixes #708.

 - Previously, if a user deleted a rank through the department edit page, the change didn't persist.
 - This fix makes the change persist.

## Notes for Deployment
- As of Thursday, June 4, this was the last bug I knew of on staging!

I made [a video ](https://www.dropbox.com/s/uzpp06hjyno3qcm/OORankFix.mp4?dl=0)of the functionality. 

## Tests and linting

 - [ ] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -425,11 +425,15 @@ def edit_department(department_id):
                     db.session.commit()
                 except IntegrityError as e:
                     db.session.rollback()
-                    rank = Job.query.filter_by(id=e.params['id']).one()
-                    home_route = url_for('main.index')
-                    print(home_route)
-                    link = '/department/{}?name=&badge=&unique_internal_identifier=&rank={}&min_age=16&max_age=100&submit=Submit'.format(department_id, rank)
-                    flash(Markup('You attempted to delete a rank, {}, that is in use by <a href={}>the linked officers</a>.'.format(rank, link)))
+                    if len(e.params) > 1:
+                        for param in e.params:
+                            rank = Job.query.filter_by(id=param['id']).one()
+                            link = '/department/{}?name=&badge=&unique_internal_identifier=&rank={}&min_age=16&max_age=100&submit=Submit'.format(department_id, rank)
+                            flash(Markup('You attempted to delete a rank, {}, that is in use by <a href={}>the linked officers</a>.'.format(rank, link)))
+                    else:
+                        rank = Job.query.filter_by(id=e.params['id']).one()
+                        link = '/department/{}?name=&badge=&unique_internal_identifier=&rank={}&min_age=16&max_age=100&submit=Submit'.format(department_id, rank)
+                        flash(Markup('You attempted to delete a rank, {}, that is in use by <a href={}>the linked officers</a>.'.format(rank, link)))
                     return redirect(url_for('main.edit_department', department_id=department_id))
             
             for (new_rank, order) in new_ranks:
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -419,11 +419,15 @@ def edit_department(department_id):
             updated_ranks = form.jobs.data
             if len(updated_ranks) < len(original_ranks):
                 deleted_ranks = [rank for rank in original_ranks if rank.job_title not in updated_ranks]
-                for rank in deleted_ranks:
-                    if Assignment.query.filter(Assignment.job_id.in_([rank.id for rank in deleted_ranks])).count() == 0:
+                if Assignment.query.filter(Assignment.job_id.in_([rank.id for rank in deleted_ranks])).count() == 0:
+                    for rank in deleted_ranks:
                         db.session.delete(rank)
-                        db.session.commit()
-                    else:
+                else:
+                    failed_deletions = []
+                    for rank in deleted_ranks:
+                        if Assignment.query.filter(Assignment.job_id.in_([rank.id])).count() != 0:
+                            failed_deletions.append(rank)
+                    for rank in failed_deletions:
                         formatted_rank = rank.job_title.replace(" ", "+")
                         link = '/department/{}?name=&badge=&unique_internal_identifier=&rank={}&min_age=16&max_age=100&submit=Submit'.format(department_id, formatted_rank)
                         flash(Markup('You attempted to delete a rank, {}, that is in use by <a href={}>the linked officers</a>.'.format(rank, link)))
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -420,22 +420,15 @@ def edit_department(department_id):
             if len(updated_ranks) < len(original_ranks):
                 deleted_ranks = [rank for rank in original_ranks if rank.job_title not in updated_ranks]
                 for rank in deleted_ranks:
-                    db.session.delete(rank)
-                try:
-                    db.session.commit()
-                except IntegrityError as e:
-                    db.session.rollback()
-                    if len(e.params) > 1:
-                        for param in e.params:
-                            rank = Job.query.filter_by(id=param['id']).one()
-                            link = '/department/{}?name=&badge=&unique_internal_identifier=&rank={}&min_age=16&max_age=100&submit=Submit'.format(department_id, rank)
-                            flash(Markup('You attempted to delete a rank, {}, that is in use by <a href={}>the linked officers</a>.'.format(rank, link)))
+                    if Assignment.query.filter(Assignment.job_id.in_([rank.id for rank in deleted_ranks])).count() == 0:
+                        db.session.delete(rank)
+                        db.session.commit()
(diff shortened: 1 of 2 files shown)


abandoned-prototype:
What happens if we delete a rank that is still in use? Is this something we want to prevent?

peculater:
> What happens if we delete a rank that is still in use? Is this something we want to prevent?

I get a 500 Internal Server Error page and the following logs in that case.
```
sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (psycopg2.errors.ForeignKeyViolation) update or delete on table "jobs" violates foreign key constraint "assignments_job_id_fkey" on table "assignments"
DETAIL:  Key (id)=(3) is still referenced from table "assignments".
```

I've got a half-written branch that includes some additional tests and the ON DELETE CASCADE migration that I think will bypass that error by deleting the assignment if the rank goes away.  I don't know if that's the right answer or if we should block the deletion and surface the error in a bit more user-understandable way.  I'm a bit slow in this codebase as I'm still learning my way around. :-)

peculater:
Looking a little closer, I think the badge number is also stored with the assignment.  I feel like that probably shouldn't get deleted when the rank does, but I could be convinced otherwise.  Thoughts?

abandoned-prototype:
from my perspective I don't think we would want any cascading effects here. So throwing an error in that case seems fine to me. If easily done it might make sense to make the error more human readable, but I wouldn't see it as a requirement given that it is functionality that is only for admins and area coordinators and I mostly see it for fixing mistakes when setting it all up

redshiftzero:
Summarizing the cases here (do we have tests for all these? I guess not but it would be wise): 
1. User edits rank (e.g. rank order)
2. User wants to delete rank that is not in use
3. User wants to delete rank that is in use
4. User wants to add a new rank

Case 1 and 4 are handled (and 2 almost is with the raw SQL at the end of the view). What about the following behavior for cases 2 and 3:
2. Keep ranks if unused unless the user requests to delete them
3. We should disallow the user from deleting ranks if in use, i.e. we flash a message "Couldn't delete rank X, in use" (bonus points to linking to the list of officers with that rank, i.e. for Sergeant  https://staging.openoversight.com/department/1?name=&badge=&unique_internal_identifier=&rank=SERGEANT+OF+POLICE&min_age=16&max_age=100&submit=Submit)

What do y'all think?

McEileen:
Good summary, @redshiftzero!

2. I agree with keeping unused ranks unless the user wants to delete them. :thumbs_up
3.  I also agree with showing an error if the user attempts to delete a rank that's in use.

I can work on implementing those changes now.  Also, thank you to @abandoned-prototype and @peculater for pointing out this edge case.

redshiftzero:
It probably makes sense for us to add tests for these four cases now (or check into why our existing tests didn't fail previously - basically to ensure we have regression coverage for the future) - I can look into this if other folks don't have bandwidth. Let me know what folks think.

McEileen:
Hey @redshiftzero,

Thanks for offering to help with tests.  I tested the case `1. User edits rank` in [test_admin_can_edit_rank_order](https://github.com/lucyparsons/OpenOversight/pull/727/files#diff-e77afe4e71b5c0f47d4461d0b9423b2aR407).  [`Test_admin_cannot_delete_rank_in_use`](https://github.com/lucyparsons/OpenOversight/pull/727/files#diff-e77afe4e71b5c0f47d4461d0b9423b2aR454) should cover the cases `2. User wants to delete rank that is not in use` and `4. User wants to add a new rank`.

I got hung up on the test for `3. User wants to delete rank that is in use`.  The test I wrote, [`test_admin_cannot_delete_rank_in_use`](https://github.com/lucyparsons/OpenOversight/pull/727/files#diff-e77afe4e71b5c0f47d4461d0b9423b2aR431), fails because it doesn't raise an `IntegrityError`, and the user can successfully delete a rank that's in use.  

At first I thought it might be a problem with my test data, but I checked, and all of the ranks are in use.  All ten of the job ids appear in the result for `SELECT DISTINCT(job_id) FROM assignments WHERE job_id IS NOT NULL;`

Any help with this test would be very appreciated!  Thanks.

abandoned-prototype:
I looked into this @McEileen and the reason no error is thrown in the test is that we use sqlite for the tests and postgres for local development (and I am assuming production). I am not sure if sqlite is not able to enforce those constrains or that we could change some settings to make it work.

Therefore, I would prefer to not wait for the database to raise the issue and add a check before we commit the changes, for example
```if Assignment.query.filter(Assignment.job_id.in_([rank.id for rank in deleted_ranks])).count() == 0:```
or something a little more elaborate to also retrieve the assignments or ranks that are still in use and cannot be deleted.

I am also open to leaving it as it is and just not having this test, given that it is an edge case and the postgres database will prevent deletion in the current implementation

**Comment by McEileen** [comment]
Thanks for the feedback, @abandoned-prototype.  I made a change and uploaded [a video of the current functionality](https://www.dropbox.com/s/ie03pk4p9yd68oi/EditDepartmentDemo.mp4?dl=0).  

Besides making it easier to test the edge case where someone attempts to delete a rank that is in use, I also fixed the link that displays officers who are that rank.  Before, the link wouldn't properly display ranks that were more than one word.

File: OpenOversight/app/main/views.py [hunk 1]
```diff
@@ -416,6 +416,19 @@ def edit_department(department_id):
...
+                deleted_ranks = [rank for rank in original_ranks if rank.job_title not in updated_ranks]
+                for rank in deleted_ranks:
+                    if Assignment.query.filter(Assignment.job_id.in_([rank.id for rank in deleted_ranks])).count() == 0:
+                        db.session.delete(rank)
+                        db.session.commit()
+                    else:
+                        formatted_rank = rank.job_title.replace(" ", "+")
+                        link = '/department/{}?name=&badge=&unique_internal_identifier=&rank={}&min_age=16&max_age=100&submit=Submit'.format(department_id, formatted_rank)
+                        flash(Markup('You attempted to delete a rank, {}, that is in use by <a href={}>the linked officers</a>.'.format(rank, link)))
+                    return redirect(url_for('main.edit_department', department_id=department_id))
```
brianmwaters, line 430: I believe this return statement being indented inside the for loop means that only one department can be edited or deleted at a time

brianmwaters: `test_admin_can_delete_rank_not_in_use` only deletes a single rank, so we wouldn't be catching this via unit tests, I my understanding of the code is correct

OpenOversight/app/main/views.py
```diff
@@ -416,6 +416,19 @@ def edit_department(department_id):
                 if rank:
                     new_ranks.append((rank, order))
                     order += 1
+            updated_ranks = form.jobs.data
+            if len(updated_ranks) < len(original_ranks):
+                deleted_ranks = [rank for rank in original_ranks if rank.job_title not in updated_ranks]
+                for rank in deleted_ranks:
+                    if Assignment.query.filter(Assignment.job_id.in_([rank.id for rank in deleted_ranks])).count() == 0:
+                        db.session.delete(rank)
+                        db.session.commit()
```
brianmwaters, line 425: Should the `db.session.commit()` be lifted out of the for loop and simply placed before the function returns or exits? That way it's an all-or-nothing deal and the user can't get stuck with partial or inconsistent results. I don't know what best practices are here, but this makes sense to me.

McEileen: Good point, I refactored and the `db.session.commit()` is now outside of the loop.

OpenOversight/app/main/views.py
McEileen: Good catch.  I added functionality to be able to delete multiple ranks that are not in use.  I also added a test for this.

McEileen:
I responded to feedback and uploaded [a video ](https://www.dropbox.com/s/bjpixgrfg2yudli/DemoEditDepartment.mp4?dl=0)that shows the below functionality in action:
1. Change rank order
2. Add a new rank
3. Delete a rank that is not in use
4. Attempt to delete an in use rank and get an error message

Here is a [second video](https://www.dropbox.com/s/fkis647caxgkml5/DemoEditDepartmentDeleteRanks.mp4?dl=0) that shows a user deleting multiple unused ranks at once.

OpenOversight/app/main/views.py
brianmwaters: Looks good, marking resolved

PR #714, opened 2020-06-06
Docker compose local file permissions

peculater:
## Status

Ready for review / in progress

## Description of Changes

Fixes #713 

Changes proposed in this pull request:

 - Moves the dev postgres storage to a docker volume, which removes a lot of the difficulties of managing file permissions shared between the docker container and the host. 
 - Sets the web container to run as the same UID as the current user.  That does a couple of things
   - Makes the `__pycache__` directories that get created on the host's filesystem owned by the user, which makes them manageable and deleteable from outside the container.
    - Gives the container process access to the OpenOversight/app/static/dist directory so that css compilation works like you expect, without needing fancy `sudo` tricks.

## Notes for Deployment

- This will invalidate any dev database you currently have.  We're no longer looking at `container_data/postgres` so if there's any data you've got in there that you're particularly fond of, you'll want to find a way to re-create it.
- If you already have a bunch of `__pycache__` directories I can't promise you'll be able to read them.  Same goes for an index.css that was created with the root user.  Honestly your best bet might be to re-clone your git repo.  Alternatively, `sudo find ./ ! -uid "$UID" -delete` would probably work if you're feeling foolishly brave.

## Tests and linting
 I tested all of this on a Fedora 31 installation with Docker 19.03 and docker-compose 1.25.4.  I'd appreciate it if someone with a different system could tell me if something obviously goes wrong.  I made sure that SCSS compilation worked, but I'm not super sure what other things I should look at.

 - [/] I have rebased my changes on current `develop`

 - [/] pytests pass in the development environment on my local machine

 - [/] `flake8` checks pass

PR #728, opened 2020-06-12
dev env: use development FLASK_ENV by default

redshiftzero:
## Status

Ready for review

## Description of Changes

 - Use `FLASK_ENV=development` key for `make dev` (done in https://github.com/lucyparsons/OpenOversight/blob/develop/docker-compose.yml#L34). This is done such that we get debug mode and live reload in the dev env.
 - Set `FLASK_ENV=testing` just for `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 export DC_UID=$(shell id -u)
-export FLASK_ENV=testing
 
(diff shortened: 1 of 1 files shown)


redshiftzero:
(this is followup to #714 so @peculater lmk if you have thoughts on this)

PR #729, opened 2020-06-12
allow photo submission with only faces, no longer require badges/names in frame

### LINKED ISSUE #723: accept photos that do NOT have names/badge numbers
There are some law enforcement officers that are covering up their name and/or badge number when working protests and a lot of folks are taking photos of them. We should consider relaxing our photo submission to accept photos for any officer in uniform: even if their badge name or name is not exposed. 

In that scenario the modified workflow could be:

1. Not logged in users submits a photo as before.
2. Volunteers use cop classification to determine if the photo fits into one of three categories: (A) cop has name/badge number, (B) cop but no name or badge number, (C) no cop. I'm suggesting this as I do think we should allow volunteers to choose whether they are merely building the OpenOversight dataset by adding faces, OR they are trying to identify a potentially problem cop by matching him with a face that exists in our dataset. These seem like different cases.
3. For cops with name/badge number (case A), same workflow as we have now -> Cop identification for the purpose of building out our dataset. 
4. For cops _without_ name/badge number (case B) -> a very similar workflow (go through roster) but: we instruct users to match the face with one that exists in the dataset. For departments in states without facial recognition constraints, we can at some later point automate this step.

redshiftzero:
## Status

Ready for review

## Description of Changes

This modifies OpenOversight to instruct folks to submit _any_ photos of officers - as long as they are in uniform. We will request that clear faces are in the frame. It will be preferred for badge numbers and/or name tags to also be in frame, but no longer strictly required. 

This will enable us to better handle the following cases: (i) where area coordinators know all cops in the city and can identify officers by face alone, (ii) where officers are in public with their badge covered by tape and there already are other images of that same face in the database.

Initially I was thinking there should be two volunteer "cop identification" buckets for folks to work from, one for identifying cops by badge/name, and one for identifying cops by face, but I'm thinking now that this is overcomplicated and we can just use our existing cop identification workflow (slightly modified the tagger search in this PR to allow this). 

Closes #723

## Notes for Deployment

I'm planning to apply this as a hotfix to prod (since not everything in `develop` is ready to ship) once merged. 

### Submit

Screen Shot 2020-06-12 at 5 37 12 PM

### Tagger search

Screen Shot 2020-06-12 at 5 37 03 PM

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/templates/about.html
+++ b/OpenOversight/app/templates/about.html
@@ -26,7 +26,7 @@
       <span class="fa fa-map-marker fa-2x"></span><p>
         It is the first project of its kind in the United States, and was first implemented in Chicago in October 2016.
         OpenOversight launched in the East Bay of the San Francisco Bay Area in fall 2017 and in New York City in 2018. A
-        Baltimore instance is anticipated to launch in early 2019.
+        Baltimore instance was launched in 2019 at <a href="https://bpdwatch.com">BPDWatch.com</a>.
     </p>
     </div>
     <div class="col-lg-4 col-half-border">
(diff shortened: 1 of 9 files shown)


redshiftzero:
deployed as hotfix to production, tag 0.5.2

PR #735, opened 2020-06-17
new officers csv download - added assignments csv download

abandoned-prototype:
## Status

Ready for review

## Description of Changes

Wanting to work with the CPD data I noticed that some of the data I would like to get is not included in the csv-file generated in the "Download department data" files. Looking into it, I also noticed that instead of the python csv library a "home-grown" solution is used.

 - Deprecating the existing officer-csv download (not removing in case there are any scripts directly accessing the page)
 - Adding a new officer-csv download that uses the python csv library and is therefore more robust (can deal with "," in strings for example)
 - Adding the unique internal identifier to the csv output
 - Replacing the assignments field with providing the (most recent) badge number and job title of each officer
 - Adding a field most recent salary to each officer
 - Finally also adding a separate assignments download that has more detailed historic information on all assignment information of the given department. Each row has officer id and unique identifier together with badge number, job title, start date, end date and unit.

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -842,9 +842,9 @@ def download_dept_officers_csv(department_id):
             most_recent_assignment = None
             most_recent_title = None
         if officer.salaries:
-            most_recent_salery = max(officer.salaries, key=lambda s: s.year)
+            most_recent_salary = max(officer.salaries, key=lambda s: s.year)
         else:
-            most_recent_salery = None
+            most_recent_salary = None
         record = {
             "id": officer.id,
             "unique identifier": officer.unique_internal_identifier,
@@ -858,7 +858,7 @@ def download_dept_officers_csv(department_id):
             "employment date": officer.employment_date,
             "badge number": most_recent_assignment and most_recent_assignment.star_no,
             "job title": most_recent_title,
-            "most recent salary": most_recent_salery and most_recent_salery.salary,
+            "most recent salary": most_recent_salary and most_recent_salary.salary,
         }
         csv_writer.writerow(record)
 
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -1,4 +1,5 @@
 import csv
+from datetime import date
 import io
 import os
 import re
@@ -835,7 +836,7 @@ def download_dept_officers_csv(department_id):
 
     for officer in officers:
         if officer.assignments_lazy:
-            most_recent_assignment = max(officer.assignments_lazy, key=lambda a: a.star_date)
+            most_recent_assignment = max(officer.assignments_lazy, key=lambda a: a.star_date or date.min)
             most_recent_title = most_recent_assignment.job and check_output(most_recent_assignment.job.job_title)
         else:
             most_recent_assignment = None
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -1,3 +1,5 @@
+import csv
+import io
 import os
 import re
 from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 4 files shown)


OpenOversight/app/models.py
```diff
@@ -101,6 +101,7 @@ class Officer(db.Model):
     employment_date = db.Column(db.Date, index=True, unique=False, nullable=True)
     birth_year = db.Column(db.Integer, index=True, unique=False, nullable=True)
     assignments = db.relationship('Assignment', backref='officer', lazy='dynamic')
+    assignments_lazy = db.relationship('Assignment')
```
abandoned-prototype, line 104: This line is necessary, to be able to use the `joinedload` on assignments, to load all assignments for all relevant officers in one go instead of a separate query for every single officer (known as N+1 problem)

OpenOversight/app/main/views.py
```diff
@@ -801,6 +804,110 @@ def download_dept_csv(department_id):
...
 
 
+def check_output(output_str):
+    if output_str == "Not Sure":
+        return ""
+    return output_str
+
+
+@main.route('/download/department/<int:department_id>/officers', methods=['GET'])
+@limiter.limit('5/minute')
```
redshiftzero, line 814: this is an aside, but a good followup here could be to cache the generated CSV

OpenOversight/app/main/views.py
```diff
@@ -801,6 +804,110 @@ def download_dept_csv(department_id):
...
+        abort(404)
+    csv_output = io.StringIO()
+    csv_fieldnames = ["id", "unique identifier", "last name", "first name", "middle initial", "suffix", "gender",
+                      "race", "birth year", "employment date", "badge number", "job title", "most recent salary"]
+    csv_writer = csv.DictWriter(csv_output, fieldnames=csv_fieldnames)
+    csv_writer.writeheader()
+
+    for officer in officers:
+        if officer.assignments_lazy:
+            most_recent_assignment = max(officer.assignments_lazy, key=lambda a: a.star_date)
```
redshiftzero, line 838: Looks like this can cause a TypeError:

```
File "/usr/src/app/OpenOversight/app/main/views.py", line 838, in download_dept_officers_csv
most_recent_assignment = max(officer.assignments_lazy, key=lambda a: a.star_date)
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```

OpenOversight/app/main/views.py
```diff
@@ -766,7 +769,7 @@ def check_input(str_input):
 
 @main.route('/download/department/<int:department_id>', methods=['GET'])
 @limiter.limit('5/minute')
-def download_dept_csv(department_id):
+def deprecated_download_dept_csv(department_id):
```
redshiftzero, line 772: :+1:

OpenOversight/app/main/views.py
abandoned-prototype: Thanks! This should be fixed now. I also checked for other fields that could be `None` and would break things, hoping I didn't miss any this time

OpenOversight/app/main/views.py
abandoned-prototype: yeah, good idea. Do we already have any pages that are cached?

OpenOversight/app/models.py
McEileen: :+1:

OpenOversight/app/main/views.py
abandoned-prototype: I created a separate issue for this: https://github.com/lucyparsons/OpenOversight/issues/742

PR #747, opened 2020-06-29
Handle officer and dept mismatch when tagging

### LINKED ISSUE #325: Application should prevent users from associating faces with wrong department
1. User is identifying images associated with Berkeley Police Department
2. User mistypes OpenOversight ID
3. ID is of a different department
4. App should prevent user from entering the wrong ID by flashing a message ``The officer is not in {{ department.name }}. Are you sure that's the correct OpenOversight ID?``

This also should have a unit test

leifan89:
## Status

Ready for review

## Description of Changes

Fixes #325  .

 - If the tagged officer is with the wrong department, do not proceed
 - Added unit test
 - 

## Tests and linting

 - [x] `flake8` checks pass


leifan89:
(Note that flake8 has flagged lines that were not touched in this PR)

OpenOversight/app/main/views.py
```diff
@@ -636,10 +636,10 @@ def leaderboard():
 def label_data(department_id=None, image_id=None):
     jsloads = ['js/cropper.js', 'js/tagger.js']
     if department_id:
-        department = Department.query.filter_by(id=department_id).one()
+        department = Department.query.filter_by(id=department_id).first()
```
abandoned-prototype, line 639: Just curious @leifan89, can you elaborate why this change is necessary? Are there situations that an invalid department_id is provided that shouldn't result in an error?

leifan89: Thanks for the review!

Unfortunately I'm having issues replicating the problem that I encountered before with `one()` instead of `first()`... do you think that it would be better to throw an error explicitly instead of just swallowing it? Happy to restore back to the original `first()` 🙂

OpenOversight/app/main/views.py
```diff
@@ -667,6 +667,8 @@ def label_data(department_id=None, image_id=None):
                          .filter(Face.original_image_id == form.image_id.data).first()
         if not officer_exists:
             flash('Invalid officer ID. Please select a valid OpenOversight ID!')
+        elif department and officer_exists.department_id != department_id:
+            flash(f'The officer is not in {department.name}. Are you sure that is the correct OpenOversight ID?')
```
abandoned-prototype, line 671: Flake8 fails here because python 3.5 does not support the "f-string" string formatting. I am not sure if we still use python 3.5 in production or anywhere, so I am definitely in favor of dropping 3.5 support, but until we do you might need to use `.format`

leifan89: Thank you for the pointer! I've updated that :)

OpenOversight/app/main/views.py
dismantl: I think leaving it as `one()` makes more sense, since there shouldn't be any cases in which there are multiple departments with the same department_id that would be returned by the query.

leifan89: So I did manage to replicate the original issue that caused me to change it to `first()`...

As it turns out, it was the test that I had added, in which I tested the scenario of an officer with a mismatched department. In that case, since the query with both that department and that officer fails to find anything (the mismatch case), `one()` fails since it expects exactly 1 result, not 0.

abandoned-prototype: @dismantl this particular case is fixed and no longer in the code

abandoned-prototype:
The instance of `first` on views.py:642 makes sense to me since it is possible to provide image_id and department_id that do not match, in which case I don't think we want to trigger a 500 error

dismantl:
> The instance of `first` on views.py:642 makes sense to me since it is possible to provide image_id and department_id that do not match, in which case I don't think we want to trigger a 500 error

Sorry you're right, I was looking at the outdated code on line 639.

dismantl:
Just merged in develop and tested, working great.

PR #750, opened 2020-07-08
Fix leftover issues w/ $UID

brianmwaters:
## Status

Ready for review

## Description of Changes

This is related to https://github.com/lucyparsons/OpenOversight/pull/739, which turned out to be an incomplete solution. This fixes an issue whereby, if the user tries to run docker-compose by hand (outside of the Makefile), they would run into an error:
```
int10h@maxwell ~/projects/OpenOversight $ docker-compose run --rm web flask bulk-add-officers --help
WARNING: The AWS_ACCESS_KEY_ID variable is not set. Defaulting to a blank string.
WARNING: The AWS_SECRET_ACCESS_KEY variable is not set. Defaulting to a blank string.
WARNING: The AWS_DEFAULT_REGION variable is not set. Defaulting to a blank string.
WARNING: The S3_BUCKET_NAME variable is not set. Defaulting to a blank string.
WARNING: The APPROVE_REGISTRATIONS variable is not set. Defaulting to a blank string.
ERROR: Missing mandatory value for "user" option interpolating ${UID:?Docker-compose needs UID set to the current user id number. Try 'export UID=$(id -u)' and run docker-compose again} in service "web": Docker-compose needs UID set to the current user id number. Try 'export UID=$(id -u)' and run docker-compose again
int10h@maxwell ~/projects/OpenOversight $
```
This is unexpected since `$UID` is explicitly set in the user's shell environment. Also, the error message is not helpful because bash does not allow the user to overwrite `$UID`, which is what the message is suggesting the user try to do.

(I think what happened was, https://github.com/lucyparsons/OpenOversight/pull/739 was created as an improvement to https://github.com/lucyparsons/OpenOversight/pull/738, and so because of this iterative development process, we did not see that there was a more elegant solution.)

This PR simply reduces the changes we've made down to the simplest possible thing that works in all scenarios I can think of

Pinging @peculater since they worked on the original PR's.

## Notes for Deployment

None.

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,7 +34,7 @@ services:
      FLASK_ENV: "${FLASK_ENV:-development}"
    volumes:
      - ./OpenOversight/:/usr/src/app/OpenOversight/:z
-   user: "${UID:?Docker-compose needs UID set to the current user id number. Try 'export UID=$(id -u)' and run docker-compose again}"
+   user: "${UID:?Docker-compose needs UID set to the current user id number. Try 'export UID' and run docker-compose again}"
    links:
      - postgres:postgres
    expose:


peculater:
Unfortunately, I don't think this is going to work the way you expect.  Turns out $UID isn't an environment variable so much as a shell builtin like $HISTSIZE.  See https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html for a bit more information.  Docker-compose doesn't find it in its environment if you try to invoke it directly like that.

You're right that UID is read-only in bash.  I think a better documentation suggestion than what got into #739 is just `export UID` which in my testing does copy UID from a builtin to an environment variable, and is then usable in docker-compose.

Either way, the Makefile has a totally different understanding of environment variables than bash does, so it's probably a good idea to leave that export line alone.

brianmwaters:
I hear what you're saying about UID being a shell builtin. However, it appears to be getting passed to docker-compose regardless, and the patch seems to work for me (I've verified that I don't have some unique situation where some profile script added UID to my environment). Commands run via makefile, like `make assets`, continue to work with this patch, but commands run by hand, which didn't work before, such as `docker-compose run --rm web flask abc xyz` now work also, because the error message is no longer stopping them from running.

I agree that removing the `export UID=xyz` from the makefile is unnecessary and am happy to put it back.

What do you think?

PS, you're right that UID isn't passed to subprograms like an environment variable:
```
int10h@maxwell ~/projects/OpenOversight $ echo $UID
1000
int10h@maxwell ~/projects/OpenOversight $ printenv UID
int10h@maxwell ~/projects/OpenOversight $
```

PPS, I'm running bash 5.0 patched up to p17, and docker-compose 1.25.4 on Gentoo. Somehow docker-compose seems to be picking up UID despite its status as shell builtin. Is this not happening for you?

brianmwaters:
I noticed that the very next commits to docker compose after 1.25.4 (which is what I'm running) add support for a new version of the docker-compose.yml file format, so that could be why we are getting different results: https://github.com/docker/compose/commits/1.25.x

brianmwaters:
Ah, crap, you are right. Docker-compose isn't, in fact, picking up the $UID value 🤦

brianmwaters:
I've force-pushed and updated this PR to undo all my changes and simply change the error message to give better advice to the user.

Namely, it suggests running `export UID`, which does in fact solve the problem for the user.

Should be simple to merge.

brianmwaters:
Uh, I guess rebasing on develop to simplify the commit history invalidated @peculater's review and now I'm blocked from merging this. It's just a simple 1-line change to make a comment more accurate, would appreciate someone's sign-off so we can merge it and forget about it!

PR #751, opened 2020-07-09
Make teeny text less teeny

dismantl:
## Status

Ready for review

## Description of Changes

A lot of the text on the [submit page](https://openoversight.com/submit) is super teeny (and I would say not great for accessibility), both due to a missing slash in a `</small>` tag, and the `small` class being applied to paragraph tags. This PR addresses both of those, making the text less teeny and more readable.

## Tests and linting

 - [x] `flake8` checks pass


dismantl:
I didn't see it anywhere else on `<p>` tags, just header tags which seems fine to me.

PR #757, opened 2020-07-17
Test the python versions we think we're testing

### LINKED ISSUE #754: Issue getting dev server running
When following the dev instructions, I get the following error:

```
Starting openoversight_postgres_1 ... done
  File "../create_db.py", line 2, in <module>
    from OpenOversight.app import create_app
  File "/usr/src/app/OpenOversight/app/__init__.py", line 103, in <module>
    app = create_app()
  File "/usr/src/app/OpenOversight/app/__init__.py", line 32, in create_app
    from .models import db  # noqa
  File "/usr/src/app/OpenOversight/app/models.py", line 15, in <module>
    db = SQLAlchemy()
  File "/usr/local/lib/python3.8/site-packages/flask_sqlalchemy/__init__.py", line 715, in __init__
    self.Model = self.make_declarative_base(model_class, metadata)
  File "/usr/local/lib/python3.8/site-packages/flask_sqlalchemy/__init__.py", line 797, in make_declarative_base
    model.query_class = self.Query
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/ext/declarative/api.py", line 79, in __setattr__
    _add_attribute(cls, key, value)
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/ext/declarative/base.py", line 796, in _add_attribute
    type.__setattr__(cls, key, value)
TypeError: can't apply this __setattr__ to DefaultMeta object
make: *** [create_db] Error 1
```

r4v5:
There's a regression in python 3.8.4, which broke all 3 of our builds despite the travis build matrix. This is because we weren't actually running the tests using those versions, because the tests run inside docker where the version was specified by the Dockerfile to be 3.8.

Luckily, flake8 was still running in the travis python version, so we caught things like f-strings not being supported in 3.5.

I expect this will fail on 3.8 and pass on 3.5, 3.6, and 3.7, but I won't know until it's pushed. Once that's verified to work, we can drop 3.8 support until we're running an OS somewhere that needs it or the image is fixed.

## Status

Ready for review

## Description of Changes

Fixes #754 .

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,7 @@ services:
       context: .
       args:
       - DOCKER_BUILD_ENV
+      - TRAVIS_PYTHON_VERSION
       dockerfile: ./dockerfiles/web/Dockerfile
    environment:
      AWS_ACCESS_KEY_ID: "${AWS_ACCESS_KEY_ID}"
--- a/dockerfiles/web/Dockerfile
+++ b/dockerfiles/web/Dockerfile
@@ -1,4 +1,5 @@
-FROM python:3-buster
+ARG TRAVIS_PYTHON_VERSION
+FROM python:${TRAVIS_PYTHON_VERSION:-3.5}-buster
 
 WORKDIR /usr/src/app
 
@@ -6,15 +7,13 @@ ENV DEBIAN-FRONTEND noninteractive
 ENV DISPLAY=:1
 ENV GECKODRIVER_VERSION="v0.26.0"
 RUN echo "deb http://deb.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/backports.list
-RUN apt-get update && apt-get install -y xvfb firefox-esr libpq-dev python3-dev
-RUN apt-get install -y -t stretch-backports libsqlite3-0
+RUN wget -O - https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get update && apt-get install -y xvfb firefox-esr libpq-dev python3-dev nodejs && apt-get install -t stretch-backports libsqlite3-0 && apt-get clean
 
 RUN wget https://github.com/mozilla/geckodriver/releases/download/${GECKODRIVER_VERSION}/geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz
 RUN mkdir geckodriver
 RUN tar -xzf geckodriver-${GECKODRIVER_VERSION}-linux64.tar.gz -C geckodriver
 
-RUN wget -O - https://deb.nodesource.com/setup_8.x | bash -
-RUN apt-get update && apt-get install -y nodejs npm
 RUN npm install -g yarn
 
 RUN mkdir /var/www ./node_modules /.cache /.yarn /.mozilla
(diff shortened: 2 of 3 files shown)


dockerfiles/web/Dockerfile
```diff
@@ -1,20 +1,19 @@
...
 
 WORKDIR /usr/src/app
 
 ENV DEBIAN-FRONTEND noninteractive
 ENV DISPLAY=:1
 ENV GECKODRIVER_VERSION="v0.26.0"
 RUN echo "deb http://deb.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/backports.list
-RUN apt-get update && apt-get install -y xvfb firefox-esr libpq-dev python3-dev
-RUN apt-get install -y -t stretch-backports libsqlite3-0
+RUN wget -O - https://deb.nodesource.com/setup_10.x | bash -
```
brianmwaters, line 10: It looks like we're also moving to a different version of Node here, in addition to changes to Python? What's up w/ that?

docker-compose.yml
```diff
@@ -23,6 +23,7 @@ services:
       context: .
       args:
       - DOCKER_BUILD_ENV
+      - TRAVIS_PYTHON_VERSION
```
brianmwaters, line 26: I'm assuming Travis sets this.

dockerfiles/web/Dockerfile
```diff
@@ -1,20 +1,19 @@
-FROM python:3-buster
+ARG TRAVIS_PYTHON_VERSION
+FROM python:${TRAVIS_PYTHON_VERSION:-3.5}-buster
```
brianmwaters, line 2: :+1:

brianmwaters:
Ok @r4v5 this all looks good; I checked the actual Travis logs and I can see the previous error we were having and that it is gone now. Also, tests pass locally for me FWIW. If you can just explain what's going on w/ that Node.js change, then I will approve and merge and our Travis woes will be gone for now

dockerfiles/web/Dockerfile
r4v5: Oh, that was in response to the EOL notice I saw when doing the docker build (8.x EOLed, and now 10 is the current LTS).

docker-compose.yml
r4v5: Yep!

dockerfiles/web/Dockerfile
brianmwaters: What do we actually use Node for besides generating static assets? I feel like we should do a little testing

brianmwaters: FWIW tests are passing, but I don't know how much automated in-browser testing (e.g, via Selenium or similar) we do. I haven't done a comprehensive click-around-and-make-sure-everything-works; in particular, our image upload feature uses a lot of frontend magic IIRC, and that can't even be tested locally, only on staging (unless you set up a personal S3 bucket).

What do you think about splitting this one change out to a separate PR so that we can merge your (very necessary) Python fix, while making sure we properly QA any JavaScript that might be broken by this?

Issue #746, opened 2020-06-28
Using python auto-formatter


abandoned-prototype:
We currently enforce the code style in our code-base using flake8. One downside of that tool is that flake8 points out issues, but doesn't really help you to solve it. I found this particularly annoying with errors like `continuation line under-indented for visual indent` after breaking up a function call that is too long into multiple lines.

That is why I propose that we select an auto-formatter that we can use from now on. I found this blog post that compares a few: https://www.kevinpeters.net/auto-formatters-for-python. Generally it seems to me that at the current time an auto-formatter called black is most widely used: https://black.readthedocs.io/en/latest/index.html. The strengths of it are that it comes with pretty good default configurations and basically makes code-styling decisions for us, ideally leading to a consistent style throughout the code base. Another advantage of black is that it uses the existing `.flake8` file, so any file formatted with black should automatically be passing flake8.

Hoping to get some feedback from people that have experience with either black or any other auto-formatter. I definitely think adding an auto-formatter is valuable to our project, but I am not insisting on a particular one.

redshiftzero:
yeah I much prefer the autoformatting tools like Golang and Rust have. I use black in most of my other Python projects, I like it a lot, so :+1: from me to adding it to the linters (might be a good time to add a `make lint` target for devs) and running `black --check` in CI (exits 1 if files will be reformatted)

fritzdavenport:
I personally feel like black is too intense about some things, but consistency is always good and the community seems to throw a lot of weight behind black. I use pycharm and know there's both black (and flake8) autoformatting options so :+1: from me

abandoned-prototype:
Has been added via https://github.com/lucyparsons/OpenOversight/commit/7bd564b4641408cc7e96ed26cb6315c230e9e7eb

PR #767, opened 2020-07-26
Implementation of #743

### LINKED ISSUE #743: bulk_add_officers update by ID
When using the bulk_add_officers command officers can be identified by either the `unique_internal_identifier` or a combination of first name, last name and badge number. @brianmwaters also has a PR pending that allows to identify officers by just first name, last name to be used by smaller departments that don't have any name collision issues.

For Chicago, these option do not seem to be sufficient. Badge numbers seem to be assigned to the role and not the person, which means that they change with a promotion of the officer. Internally CPD uses employee numbers as unique identifiers, but these are usually redacted in record requests.

**Therefore, I propose the following:** bulk_add_officers also accepts the ID of the officer in the database as the key for merging the incoming csv file with the existing data. So if an `id` field is present in the provided csv, an officer will be selected solely based on that `id`. If there is no officer with that `id` the command fails. If the field is empty a new officer will be added to the database, regardless of all other fields present.

The down side of this implementation is that if the `id` field is present but left empty, each row will create a new officer, which could be unintentional. Further, updating the existing roster requires working with the data from the website, i.e. downloading the csv from https://openoversight.lucyparsonslabs.com/download/all and then merging accordingly.
On the plus side, this implementation avoids unintentional merges of officers and gives the creator of the csv full control over what will happen once this csv file is merged which seems critical to me when it comes to updating the rosters of police department with 10,000+ officers.

I am happy to work on implementing this, but hoping for comments if this makes sense.

abandoned-prototype:
## Status

Ready for review

## Description of Changes

Fixes #743  .

This implements ideas proposed in #743 
Specifically
 - A csv upload that can populate Officer, Assignment, Salary, Link and Incident Model via separate csv files.
 - The `id` field has to be present in each of the provided csv. if empty a new object will be created and added to the database, if a number the appropriate entry will be loaded from the database and the provided fields will be overwritten (kind of a PATCH). For Officer and Incident, if the field starts with `#` it can be used in other csv to refer to the newly created object (For example to add officers and assignments in one run of the command)
- The `force_create` can be used to instead of updating an entry of given id, create a new entry with that id (regardless of one already existing or not). This is not supposed to be used in production, the idea is to enable people to download the data from the production server and then import the data locally while ensuring that all relevant ids are the same as in production, so a run of the import command can be tested locally
 - Example csv files are provided in the test to make it easier to develop with, an rst documentation is included. More details on the command in the documentation

## Tests and linting

 - [x] `flake8` checks pass

## For merging

Please either wait for me to merge it after approval or make sure to use the squash & merge feature


abandoned-prototype:
To add some context: This is needed to update/fix the Chicago rooster

OpenOversight/app/validators.py
```diff
@@ -1,10 +1,19 @@
 from us import states
+from urllib.parse import urlparse
 
 
 def state_validator(state):
     list_of_states = [st.abbr for st in states.STATES]
 
-    if state not in list_of_states and state != 'DC':
+    if state not in list_of_states and state != 'DC' and state is not None:
```
McEileen, line 8: Why do we need to handle the possibility that the `state` parameter is `None`?

OpenOversight/tests/conftest.py
```diff
@@ -214,6 +214,12 @@ def test_jpg_BytesIO():
     return byte_io
 
 
+@pytest.fixture
+def test_csv_dir():
```
McEileen, line 218: :+1: for testing!

OpenOversight/tests/test_commands.py
```diff
@@ -1,14 +1,31 @@
...
 
+from click.testing import CliRunner
 from sqlalchemy.orm.exc import MultipleResultsFound
 
 import pandas as pd
 import pytest
-from click.testing import CliRunner
-from OpenOversight.app.commands import add_department, add_job_title, bulk_add_officers
-from OpenOversight.app.models import Assignment, Department, Job, Officer, Salary
+from OpenOversight.app.commands import (
```
McEileen, line 13: Did you make these formatting changes due to personal preference?  Or did flake8 request them?  I'm not sure if I'm a fan of putting every `command` and `model` import on a separate line.

OpenOversight/tests/test_commands.py
```diff
@@ -736,3 +753,458 @@ def test_bulk_add_officers__no_create_flag(session, department, csv_path):
     officer = Officer.query.filter_by(department_id=department_id).one()
     assert officer.unique_internal_identifier == officer_uuid
     assert officer.gender == officer_gender_updated
+
+
+def test_advanced_csv_import(session, department_with_ranks, test_csv_dir):
```
McEileen, line 758: :raised_hands: for testing!

OpenOversight/tests/test_csvs/.gitignore
```diff
@@ -0,0 +1 @@
+!*.csv
```
McEileen, line 1: Why was this `.gitignore` necessary?  (I think I might be missing something here.)

docs/advanced_csv_import.rst
```diff
@@ -0,0 +1,196 @@
...
+or create new entries based on the existing data.
+
+The advanced csv upload is for the most part a way to copy data for one department into the database with as little as possible logic added on.
+So the tables provided in csv form represent the data that will be inside the sql tables after running the command.
+(With a few exceptions for many-to-many relationships and auxiliary models like location and license plates)
+
+Before you start
+----------------
+
+CSV uploads should always be tested locally or in other non-production environments and it is strongly recommended
```
McEileen, line 16: :100:

OpenOversight/app/validators.py
abandoned-prototype: The models for `LicensePlate` and `Location` both allow their field `state` to be `NULL` that's why I want to allow it for the import. So I guess the only time it is relevant is when it's unknown which state a license plate belongs to. I am fine with reverting this to not allowing `None`

OpenOversight/tests/test_commands.py
abandoned-prototype: I was using the auto-formatter "Black" on that file, I am assuming it decided that the line will be too long with all those imports. I don't have an opinion on this, but I do hope to adapt Black as our formatting standard at some point: https://github.com/lucyparsons/OpenOversight/issues/746

OpenOversight/tests/test_csvs/.gitignore
abandoned-prototype: we have a `.gitignore` on the repository root level that includes `.csv` files which makes some sense, since we don't want to bring real world data into this repository. So I added this `.gitignore` file to basically revert the ignore on csv files in that particular directory so I can add the `.csv` files used in some of the tests.

OpenOversight/app/commands.py
```diff
@@ -411,6 +421,388 @@ def bulk_add_officers(filename, no_create, update_by_name, update_static_fields)
...
+
+
+def _objects_from_split_field(field, model_lookup):
+    if field:
+        return [model_lookup[object_id] for object_id in field.split("|")]
+    return []
+
+
+def _unify_field_names(fieldnames):
+    return [field_name.lower().replace(" ", "_") for field_name in fieldnames]
```
fritzdavenport, line 467: I've used `inflection` to convert between cases before - not sure that adding a dep is better or worse than using a common-enough function, here.
https://inflection.readthedocs.io/en/latest/

OpenOversight/app/commands.py
```diff
@@ -411,6 +421,388 @@ def bulk_add_officers(filename, no_create, update_by_name, update_static_fields)
...
+                csv_name="officers",
+            )
+
+            for row in csv_reader:
+                # can only update department with given name
+                assert row["department_name"] == department_name
+                row["department_id"] = department_id
+                connection_id = row["id"]
+                if row["id"].startswith("#"):
+                    row["id"] = ""
```
fritzdavenport, line 542: this feels weird? Why just comment out the ID? Is this a dirty-data fix?

OpenOversight/app/commands.py
```diff
@@ -411,6 +421,388 @@ def bulk_add_officers(filename, no_create, update_by_name, update_static_fields)
...
+    row, existing_model_lookup, create_method, update_method, force_create=False, model=None
+):
+    if not row["id"]:
+        return create_method(row)
+    else:
+        if not force_create:
+            return update_method(row, existing_model_lookup[int(row["id"])])
+        else:
+            if model is not None:
+                existing = model.query.filter_by(id=int(row["id"])).first()
```
fritzdavenport, line 434: Could an `upsert` be better here than using the ORM and going one-by-one?

OpenOversight/app/commands.py
```diff
@@ -411,6 +421,388 @@ def bulk_add_officers(filename, no_create, update_by_name, update_static_fields)
...
+                    existing_model_lookup=id_to_salary,
+                    create_method=create_salary_from_dict,
+                    update_method=update_salary_from_dict,
+                    force_create=force_create,
+                    model=Salary,
+                )
+                counter += 1
+                if counter % 1000 == 0:
+                    print("Processed {} salaries.".format(counter))
+        print("Done with salaries. Processed {} rows.".format(counter))
```
fritzdavenport, line 663: Some of the code looks relatively familiar between salary/officer/assignment. Can these update functions be further generified?

OpenOversight/app/commands.py
```diff
@@ -411,6 +421,388 @@ def bulk_add_officers(filename, no_create, update_by_name, update_static_fields)
...
+        select setval('assignments_id_seq', (select max(id) from assignments));
+        select setval('links_id_seq', (select max(id) from links));
+        select setval('incidents_id_seq', (select max(id) from incidents));
+        """
+        try:
+            db.session.execute(raw_sql)
+            print("Updated sequences.")
+        except SQLAlchemyError:
+            print("Failed to update sequences")
+
```
fritzdavenport, line 804: this seems like a lot of custom ETL code in general, especially to bake into a CLI too. It mostly looks good, but could some more-etl specific framework help us here, e.g. pandas or something? compared to vanilla python? Also should this script be in the CLI tool specifically? Will it remain tenable?

OpenOversight/app/model_imports.py
```diff
@@ -0,0 +1,307 @@
...
+        for label, choice_value in given_choices:
+            if value.lower() in [choice_value.lower(), label.lower()]:
+                return label
+        print("'{}' no valid choice of {}".format(value, str(given_choices)))
+    return None
+
+
+def parse_date(date_str: Optional[str]) -> Optional["datetime.date"]:
+    if date_str:
+        return dateutil.parser.parse(date_str).date()
```
fritzdavenport, line 34: good stuff, I'm a fan of this lib :+1:

fritzdavenport: nit: why `"datetime.date"` above? `Optional[datetime.date]` also works, and is what I'm used to; didn't know you could do the quoted version

OpenOversight/app/commands.py
abandoned-prototype: I didn't know that library, thanks! However I wasn't able to find a command that does what I was doing here and I also value that it doesn't require looking up a library to find out what it is happening here

OpenOversight/app/commands.py
abandoned-prototype: This is my approach to be able to import foreign key relationships. Basically, if you set an id-field in the officer csv to `#1` and then use it in assignments csv in the field `officer_id``the assignment will be created for that officer. See the csv files in the tests for examples of how this looks like. I also tried to explain it in the added documentation.
Also, I am not sure if this is the best way to implement these kind of relations, so I am certainly open to suggestions to do it differently.

OpenOversight/app/commands.py
abandoned-prototype: Good point. Although the way I implemented it, there is just a little bit too much logic happening to make it work I think.

In the normal case, if no id field is provided it has to be an insert, and if the id field is provided it needs to be an update, the imported csv should not have the power to decide the id. With the `--force-create` it might make sense, although I think there is a difference to entries related via foreign keys between deleting an entry and creating a new one with the same id and just updating the entry

OpenOversight/app/model_imports.py
abandoned-prototype: Without quotes I need to actually import it, this way I can put the import under `if TYPE_CHECKING`, so it's not imported at runtime. Not really relevant with small built-in libraries like this one, but I like to do it until we have at least python 3.7 and `from __future__ import annotations` which circumvents the problem: https://www.python.org/dev/peps/pep-0563/

OpenOversight/app/commands.py
abandoned-prototype: Definitely a good question. I can't think of a way of doing his with pandas, but I am certainly open to suggestions. Specifically I don't know of solutions (within pandas or any ETL tool) to import several tables and maintain foreign key relationships between them (since the respective primary keys are only decided once loaded into the database)

OpenOversight/app/commands.py
abandoned-prototype: Yeah, I wasn't too excited with how this turned out looking, but I also couldn't come up with a better way to extract more common functionality without making it look weird, but I will give it another try

OpenOversight/app/commands.py
abandoned-prototype: Also, I do agree that this code shouldn't all live in the `commands.py`. I will see where I can move it

abandoned-prototype:
@fritzdavenport I made some changes that hopefully increase at least maintainability a little bit. There are definitely better ways to do this, and if you have any ideas I am very curious to learn about them. At the same time if you don't see any glaring problems I would like to get this into production so we can start importing some data

fritzdavenport:
@abandoned-prototype :+1: from me. I'd be down to talk more on slack about what the imports look like in general, to maybe talk about not needing to maintain as much custom import code for a v2 of this if we want it

Issue #770, opened 2020-07-26
Open Graph support


dismantl:
When sharing officer profiles on social media, it would be nice to have the link show up with the officer's name and picture in the little preview. Open Graph is what facebook uses and twitter has their "cards" thing and there might be other standards worth looking at.

PR #779, opened 2020-07-31
Featured tag

dismantl:
## Status

Ready for review

## Description of Changes

When an officer has multiple tags, the profile displays the images in the order they were tagged, and the first one in the database is used as the thumbnail image in the browse view. This isn't great if the first tag of an officer isn't good quality, but that's the one that always shows up in the browse view despite the officer having other better quality tags.

This PR adds a "featured" boolean flag to the Face model, and a button on the Tag Edit page to set the tag as featured for an officer. Only one tag for an officer can be set as featured at a time; setting another tag as featured simply sets featured=False on the old featured tag.

If an officer has a tag that is featured=True, it will be the first image listed on the officer's profile, and will be the image used for the department browse view.

I plan to use this feature to select the image used for Open Graph tags, in issue #770.

## Notes for Deployment

**NOTE:** This requires a database migration.

image

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/templates/officer.html
```diff
@@ -13,16 +13,33 @@ <h1>Officer Detail: <b>{% include 'partials/officer_name.html' %}</b></h1>
...
+          {% if current_user.is_administrator
+            or (current_user.is_area_coordinator and current_user.ac_department_id == officer.department_id) %}
+          <a href="{{ url_for('main.display_tag', tag_id=faces[loop.index - 1].id) }}">
+          {% endif %}
+          <img src="{{ path }}" alt="Submission">
+          {% if current_user.is_administrator
+            or (current_user.is_area_coordinator and current_user.ac_department_id == officer.department_id) %}
+          </a>
+          {% endif %}
+        {% endif %}
```
abandoned-prototype, line 42: I am not a fan of putting yet more logic into templates. Why not just add a ORDER BY `featured` to where we retrieve the Faces? Then, I don't think we need any of the chances here

OpenOversight/app/main/views.py
```diff
@@ -632,6 +632,39 @@ def delete_tag(tag_id):
...
 
 
+@main.route('/tag/set_featured/<int:tag_id>', methods=['POST'])
+@login_required
+@ac_or_admin_required
+def set_featured_tag(tag_id):
+    tag = Face.query.filter_by(id=tag_id).first()
+
+    if not tag:
+        flash('Tag not found')
```
fritzdavenport, line 642: Is there a standard for doing a 404 vs doing a flash and redirect?

OpenOversight/app/templates/list_officer.html
```diff
@@ -131,11 +131,16 @@ <h3 class="panel-title accordion-toggle">Age range</h3>
     {% endwith %}
     <ul class="list-group">
       {% for officer in officers.items %}
-        {% if officer.face.first() is none %}
+        {% set featured_tag = officer.face.filter_by(featured=True).one_or_none() %}
```
fritzdavenport, line 134: should this be in the html vs in the route itself?

fritzdavenport: officer_image as well, below

OpenOversight/app/main/views.py
dismantl: You're right I think it should be a 404. I used the delete_tag() function as a template, but that should probably be a 404 as well. I'll change that.

OpenOversight/app/templates/officer.html
dismantl: yup that's definitely a better way to go

dismantl:
Thanks for the feedback @abandoned-prototype @fritzdavenport ! I've moved the logic out of the templates and simplified things a bit.

dismantl:
Hmm one of the tests on Travis failed but it wasn't a test I had modified...

Makefile
```diff
@@ -40,8 +40,8 @@ populate: create_db  ## Build and run containers
 .PHONY: test
 test: start  ## Run tests
 	if [ -z "$(name)" ]; \
-	    then FLASK_ENV=testing docker-compose run --rm web pytest -n 4 --dist=loadfile -v tests/; \
-	    else FLASK_ENV=testing docker-compose run --rm web pytest -n 4 --dist=loadfile -v tests/ -k $(name); \
+	    then FLASK_ENV=testing docker-compose run --rm web pytest -n $$(nproc --all) --dist=loadfile -v tests/; \
```
redshiftzero, line 43: heads up that this is not platform independent as is (`nproc` not supported on mac out of the box), we could either put into a separate PR or revert this change? (rest of this PR looks good)

Alternatively the mac equivalent is `sysctl -n hw.logicalcpu` so one could use that if `Darwin` is the output of `uname`

dismantl: Good call, I'll update that to check.

brianmwaters:
Holy shit, you don't know how many hours I spent today working around this by writing SQL queries by hand over and over again. Thank you! Wish I knew this was in the works, I would have just waited

PR #796, opened 2020-08-12
[deployed as hotfix] bugfix: resolve TypeError when trying to concatenate str with NoneType

### LINKED ISSUE #793: Chicago PD's browse page 500s on both prod and staging
When I navigate to the "browse department" page for the Chicago Police Department, there is a 500 on both prod and staging.

When I looked at logs in staging, I saw this

```
$:   File "/home/nginx/oovirtenv/venv/OpenOversight/OpenOversight/app/templates/list_officer.html", line 141, in block "content"
$:     {% set star_no = "#" + assignment.star_no %}
$: TypeError: Can't convert 'NoneType' object to str implicitly
```

I only see this bug for the Chicago Police Department.  My initial theory is that something get messed up with the migrations for Chicago, and the assignments are missing from the database.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #793 which is causing a 500 in production and staging

This has been applied to staging and production already as a hotfix

## Tests and linting
- [x] I have rebased my changes on current develop
- [ ] pytests pass in the development environment on my local machine (relying on CI)
- [ ] flake8 checks pass (relying on CI but no Python code changes in this diff anyway)

### DIFF
--- a/OpenOversight/app/templates/list_officer.html
+++ b/OpenOversight/app/templates/list_officer.html
@@ -138,7 +138,7 @@
           {% endif %}
           {% if officer.assignments.all() %}
             {% set assignment = officer.assignments.order_by(officer.assignments[0].__table__.c.star_date.desc()).first() %}
-            {% if assignment is not none %}
+            {% if assignment is not none and assignment.star_no is not none %}
                 {% set star_no = "#" + assignment.star_no %}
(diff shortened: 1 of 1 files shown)


McEileen:
Thank you for fixing this! :pray:

Issue #795, opened 2020-08-11
refactor: OpenOversight/app/templates/list_officer.html


redshiftzero:
Followup from https://github.com/lucyparsons/OpenOversight/pull/794/files: @McEileen noted we have a lot of logic in this template, if we can move as much of this as possible to the view function it'll make bugs like #793 easier to find/test

PR #800, opened 2020-08-15
SEO Improvements

### LINKED ISSUE #770: Open Graph support

dismantl:
## Status

Ready for review

## Description of Changes

Fixes #770, #795.

 - Customize <title> and meta descriptions for all public page templates.
 - Adds schema.org structured data (JSON-LD) to officer profiles
 - Adds Open Graph meta tags to officer profiles
 - Adds Twitter Card meta tags to officer profiles
 - Adds Google Search Breadcrumb structured data to officer profiles and incident details pages
 - Adds `job_title()` and `badge_number()` methods to the `Officer` class.
 - Removes tests for last name capitalization. This is an opinionated change; I believe name capitalization should be done in the database, not in the view. Capitalization in names is so unpredictable that it is not viable to attempt to capitalize them correctly using a formula. For example, this person's last name is capitalized "VanderBrug": https://bpdwatch.com/officer/3339.

image

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/templates/officer.html
```diff
@@ -1,4 +1,82 @@
 {% extends "base.html" %}
+{% block title %}{{ officer.full_name() }} - OpenOversight{% endblock %}
+{% block meta %}
+  {% set job_title = officer.assignments[0].job.job_title if officer.assignments[0] and officer.assignments[0].job.job_title != 'Not Sure' else 'Employee' %}
```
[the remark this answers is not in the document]
abandoned-prototype: might actually make sense to have a `job_title()` method on the `officer` model that contains whichever logic we decide we want to use

OpenOversight/app/templates/incident_list.html
```diff
@@ -1,5 +1,18 @@
...
+		No Incidents Found
+	{%- endif -%}
+{% endblock %}
+{% block meta %}
+	{% if objects.items|length > 0 %}
+		<meta name="description" content="View all incidents for {{ objects.items[0].department.name }}.">
+	{% else %}
+		<meta name="description" content="No incidents found.">
+	{% endif %}
+{% endblock %}
```
abandoned-prototype, line 15: Can we simplify both of these? And just title it "View incidents" or something regardless of if there are results or not?
I am also fine with keeping the distinction if there are no results to say "No incidents found" and use the `department` object that is already passed to the renderer and used a little further down in the file.
The `/incident` can also be called without the department id in which case all incidents in the database will be displayed not just one department

abandoned-prototype (commented):
Awesome work! I think this is an important improvement.
Were you able to test your implementation do see that google/twitter/etc. show stuff in the way we would want it to? I don't know how this would be done

OpenOversight/app/templates/officer.html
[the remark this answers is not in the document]
dismantl: yep good call

dismantl:
> Awesome work! I think this is an important improvement.
> Were you able to test your implementation do see that google/twitter/etc. show stuff in the way we would want it to? I don't know how this would be done

Yes these changes have been up on bpdwatch.com and the social media previews can be seen in the tweets from twitter.com/whoisbpd. 
image

The search results in Google also look correct:
image

OpenOversight/app/templates/officer.html
```diff
@@ -1,14 +1,92 @@
...
```
abandoned-prototype, line 4: *nitpick* Can we simplify this line now too using the `officer.job_title()`? Still a little awkward since we have to replace both `None` and `'Not Sure'` with `'Employee'`, so maybe it can wait for when we get rid of the 'Not Sure' case everywhere.

abandoned-prototype (approved):
Thanks @dismantl. One small comment, but feel free to ignore. These are some important refactors I think. Unlocks further improvements without having to dive deep into multiple templates.
Also :+1: on using the `officer.full_name()` function instead of a template. And I definitely agree with your decision to just use the capitalization as it is in the database, this makes the most sense to me as well. We should probably take a look after this is in production to maybe do migrations for some departments (CPD for example is all upper case)

dismantl:
Thanks for the review @abandoned-prototype! When I was adding `officer.job_title()` in the template as you suggested, I saw that the `AddOfficerForm.job_title` field was being used inconsistently, so I fixed that in the code and tests. Should be ready for your review again when you have a chance :)

OpenOversight/app/main/forms.py
```diff
@@ -203,7 +203,7 @@ class AddOfficerForm(Form):
     star_no = StringField('Badge Number', default='', validators=[
         Regexp(r'\w*'), Length(max=50)])
     unique_internal_identifier = StringField('Unique Internal Identifier', default='', validators=[Regexp(r'\w*'), Length(max=50)])
-    job_title = StringField('Job Title')  # Gets rewritten by Javascript
+    job_id = StringField('Job ID')  # Gets rewritten by Javascript
```
abandoned-prototype, line 206: This change unfortunately breaks the add-officer form. Like instead of a drop-down for the job/rank it just shows a a text field (in which one has to enter the database-id of the job)

dismantl: ooo thanks for catching that @abandoned-prototype! I'll fix that now.

abandoned-prototype (approved):
Looks good. Thanks @dismantl

PR #807, opened 2020-08-20
When imghdr fails to deduce the image type, use file extension

dismantl:
## Status

Ready for review

## Description of Changes

I ran into what might be an edge case, where `imghdr.what()` failed to figure out the filetype of a submitted photo. So I added in this little snippet in case imghdr fails, to determine image type using the filename.

I'm not really sure what a good way to add tests for this is, since I don't really know why `imghdr` failed for this particular image. For reference, the image was https://bpdwatch.s3.amazonaws.com/8d/095b8184d1d4abd90e867a8b2d41961bcc4fbd2b42e2af6f0ad8c80ca44fc7.jpg

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -482,6 +482,14 @@ def crop_image(image, crop_data=None, department_id=None):
 
     image_buf.seek(0)
     image_type = imghdr.what(image_buf)
+    if not image_type:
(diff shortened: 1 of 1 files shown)


redshiftzero (approved):
just checked the extensions here, lgtm

I think we _could_ add test coverage with some test files for each extension but this is good to merge as is

PR #816, opened 2020-08-23
802: handle EXIF orientations 3, 6, and 8 when cropping photos

### LINKED ISSUE #802: Bug: photos are upright in tagging workflow but rotated 90° when cropped
I’ve seen this happen with a set of photos that came from a particular camera/phone, and I have a hunch for the cause. First — a user story:

1. **Given** I am logged in
2. And I have permissions to tag faces in photos
3. **When** I am tagging a face in a *certain* photo
4. And the photo appears at the correct orientation in the tagging interface
5. **Then** I see the cropped officer tag is rotated 90º from how it had appeared in the tagging interface
6. And the size and shape of the crop are true to how I had input them

**Expected result:** the cropped image is “upright” in the same orientation as the tagging interface
**Actual result:** the cropped image is rotated 90º (most frequently counter clockwise)

Example image: https://www.openoversight.com/officer/12035

My suspicion is the tagging interface is correctly using an EXIF property like `orientation`, e.g:
 - https://jdhao.github.io/2019/07/31/image_rotation_exif_info/
 - https://sirv.com/help/articles/rotate-photos-to-be-upright/

…and then the processing routine is either ignoring that value or incorrectly applying the orientation.

parhamr:
## Status

Ready for review

## Description of Changes

Fixes #802.

 - If an image has EXIF orientation, rotate it “upright” before applying the user-provided crop

## Screenshots

The first two photos for **Haywood BUTZ** are crops of an image having EXIF orientation values; this proves it works because the F is upright. The last photo (cartoon) has no EXIF orientation and it appears upright, as expected.

Screen Shot 2020-08-23 at 14 14 01

------

Here’s what happens with the same image files against the `develop` branch: (the `F` is erroneously rotated 90º counter-clockwise)

Screen Shot 2020-08-23 at 14 19 25

------

## Tests and linting

 - [X] I have rebased my changes on current `develop`

 - [X] pytests pass in the development environment on my local machine

 - [X] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/utils.py
+++ b/OpenOversight/app/utils.py
@@ -21,6 +21,7 @@ import imghdr as imghdr
 from flask import current_app, url_for
 from flask_login import current_user
 from PIL import Image as Pimage
+from PIL import ExifTags
 from PIL.PngImagePlugin import PngImageFile
 
 from .models import (db, Officer, Assignment, Job, Image, Face, User, Unit, Department,
@@ -473,6 +474,32 @@ def create_description(self, form):
         date_updated=datetime.datetime.now())
 
 
+def conditionally_rotate_image(image):
+    try:
+        # NOTE: Pillow >= 6.0 provides `ImageOps.exif_transpose` which acts like the below code
+
+        # find the orientation key
+        for orientation in ExifTags.TAGS.keys():
+            if ExifTags.TAGS[orientation] == 'Orientation':
+                break
+
+        exif = dict(image._getexif().items())
+
+        if exif[orientation] == 3:
+            current_app.logger.info('image had EXIF orientation 3; rotating 180 degrees')
+            image = image.rotate(180, expand=True)
+        elif exif[orientation] == 6:
+            current_app.logger.info('image had EXIF orientation 6; rotating 270 degrees')
+            image = image.rotate(270, expand=True)
+        elif exif[orientation] == 8:
+            current_app.logger.info('image had EXIF orientation 8; rotating 90 degrees')
+            image = image.rotate(90, expand=True)
+        return image
+    except (AttributeError, KeyError, IndexError):
+        # cases: image don't have getexif
+        return image
+
+
 def crop_image(image, crop_data=None, department_id=None):
     if 'http' in image.filepath:
         with urlopen(image.filepath) as response:
@@ -487,6 +514,8 @@ def crop_image(image, crop_data=None, department_id=None):
     SIZE = 300, 300
     if not crop_data and pimage.size[0] < SIZE[0] and pimage.size[1] < SIZE[1]:
         return image
+
+    pimage = conditionally_rotate_image(pimage)
 
     if crop_data:
         pimage = pimage.crop(crop_data)


parhamr:
I’m mixed on writing tests for this. On one hand, it would ensure this works, does not break other things, and does not regress; I agree it’s a best practice. On the other hand, I’m not that comfortable with Python and I’m unsure how to write the tests for this case. Help?!

This page has some useful test images having EXIF orientation values: https://magnushoff.com/articles/jpeg-orientation/

These files have rotations 3, 6, and 8, which this PR covers:
f6t
f3t
f8t

OpenOversight/app/utils.py
```diff
@@ -473,6 +474,32 @@ def create_description(self, form):
...
+
+        # find the orientation key
+        for orientation in ExifTags.TAGS.keys():
+            if ExifTags.TAGS[orientation] == 'Orientation':
+                break
+
+        exif = dict(image._getexif().items())
+
+        if exif[orientation] == 3:
+            current_app.logger.info('image had EXIF orientation 3; rotating 180 degrees')
```
parhamr, line 489: These loggers might be useless. They don’t really provide any actionable information and they expose a behavior the user would otherwise assume “just works.”

OpenOversight/app/utils.py
```diff
@@ -473,6 +474,32 @@ def create_description(self, form):
         date_updated=datetime.datetime.now())
 
 
+def conditionally_rotate_image(image):
+    try:
+        # NOTE: Pillow >= 6.0 provides `ImageOps.exif_transpose` which acts like the below code
```
dismantl, line 479: Any reason not to bump the Pillow dependency version instead?

OpenOversight/app/utils.py
dismantl: I don't see the detriment to keeping them. It would be interesting for admins to see how often we run across these EXIF rotation issues.

dismantl:
> I’m mixed on writing tests for this. On one hand, it would ensure this works, does not break other things, and does not regress; I agree it’s a best practice. On the other hand, I’m not that comfortable with Python and I’m unsure how to write the tests for this case. **Help?!**
> 
> This page has some useful test images having EXIF orientation values: https://magnushoff.com/articles/jpeg-orientation/
> 
> These files have rotations 3, 6, and 8, which this PR covers:
> f6t
> f3t
> f8t

I would suggest adding those sample 'F' images to `OpenOversight/app/static/images` and write tests similar to those in `test_image_tagging.py`. The `test_user_can_add_tag()` and `test_featured_tag_replaces_others()` tests are good examples of how different pieces of the image tagging flow can be mocked. I like the idea of using those 'F' sample images for testing because when you submit to `main.label_data` it actually does the crop and saves a new image in `raw_images`. So the test could then compare the sha256 hash values of the newly cropped images (`raw_images.hash_img`) to see if they match the pre-computed hashes of those same images cropped but also rotated correctly. Does that make sense?

parhamr:
@dismantl thanks for the feedback! I was also thinking the hashes of the image bytes would be useful for testing. I’ll poke around.

Additionally — is there an automated test runner for this code? I finally decided to speed up my test runs by specifying individual files in the `-v tests/` part of the Makefile (https://github.com/lucyparsons/OpenOversight/blob/develop/Makefile#L41-L50), but that wasn’t quite as productive as I had hoped.

OpenOversight/app/utils.py
parhamr: I’ll try that, too. I was a bit hesitant to mess with dependencies.

parhamr:
I’m going to rework this on Pillow >= 6.0

dismantl:
> @dismantl thanks for the feedback! I was also thinking the hashes of the image bytes would be useful for testing. I’ll poke around.
> 
> Additionally — is there an automated test runner for this code? I finally decided to speed up my test runs by specifying individual files in the `-v tests/` part of the Makefile (https://github.com/lucyparsons/OpenOversight/blob/develop/Makefile#L41-L50), but that wasn’t quite as productive as I had hoped.

@parhamr Yes, `make test` should run all the tests in docker, and you can also run individual tests or individual test modules by running `name=<test_or_module> make test`.

PR #822, opened 2020-09-02
[WIP] Twitterbot

dismantl:
## Status

In progress

## Description of Changes

Like @cpdpbot but for OpenOversight. Auto-responding to mentions that include a cop's name, and a more in-depth interactive DM conversation that essentially mimics the Find an Officer form.

Other changes:
- Allows multiple comma-separated values for the Badge No. and UII fields on the browse view, allowing us to construct links for essentially any list of officers.

## Tests and linting

 - [x] `flake8` checks pass


OpenOversight/app/config.py
```diff
@@ -62,6 +72,12 @@ class TestingConfig(BaseConfig):
     NUM_OFFICERS = 120
     APPROVE_REGISTRATIONS = False
     SITEMAP_URL_SCHEME = 'http'
+    TWITTER_CONSUMER_KEY = 'TWITTER_CONSUMER_KEY'
+    TWITTER_CONSUMER_SECRET = 'TWITTER_CONSUMER_SECRET'
+    TWITTER_ACCESS_TOKEN = 'TWITTER_ACCESS_TOKEN'
+    TWITTER_ACCESS_TOKEN_SECRET = 'TWITTER_ACCESS_TOKEN_SECRET'
+    TWITTER_WEBHOOK_ENV = 'testing'
+    SERVER_NAME = 'localhost:5000'
```
dismantl, line 80: `SERVER_NAME` is needed to generate the `url_for(..., _external=True)` links for the twitter bot responses

OpenOversight/app/twitterbot.py
```diff
@@ -0,0 +1,492 @@
...
+        response = self.api_request('direct_messages/welcome_messages/list')
+        if 'welcome_messages' in response:
+            for message in response['welcome_messages']:
+                self.api_request(
+                    'direct_messages/welcome_messages/destroy',
+                    {'id': message['id']}
+                )
+
+    def register_webhook(self):
+        self.delete_all_webhooks()
```
dismantl, line 95: this is redundant, I need to remove this line

abandoned-prototype:
I really like the idea of this. Having seen it with the cpdp bot, this is a neat way to gain some publicity and put anther tool into peoples hands to quickly look up an officer. That said, it is not clear to me why this project should be included in the OpenOversight server code instead of having it as a standalone service that could be deployed on the same server but doesn't have to be. Furthermore, I am assuming we would probably need one bot per department (for which the responsible groups want one) otherwise we risk people looking for an officer in for example Burlington and getting an officer with the same name from Chicago.
The csv-download path `/download/department/<department_id>/officers` allows a download of a csv-file that could be used by the twitter-bot service to have all the most recent data available for the particular department (link to OpenOversight profile, name, rank, most-recent salary & badge-number) or we could build some other API routes a twitter-bot can access.
I hope this makes sense, I really like the idea and the work looks good, I am just hesitant about adding more complexity and dependencies to this project

dismantl:
I initially created a new repo, but I had to have OO as a git submodule because the twitter bot needs to import the OO models, query the database, and use a number of other utilities. If it were to be standalone, there would be a whole lot that would need to be rewritten (essentially duplicating code), which I am not a fan of. I just don't think something like this, which needs to tie in to a lot of OO code, would be feasible standalone which is why I'm integrating it in this way.

> The csv-download path /download/department/<department_id>/officers allows a download of a csv-file that could be used by the twitter-bot service to have all the most recent data available for the particular department (link to OpenOversight profile, name, rank, most-recent salary & badge-number) or we could build some other API routes a twitter-bot can access.

This would require regular exports/syncs of the DB to the CSV. Again, not a fan of that maintenance burden.

> Furthermore, I am assuming we would probably need one bot per department (for which the responsible groups want one) otherwise we risk people looking for an officer in for example Burlington and getting an officer with the same name from Chicago.

I don't think separate bots are necessary. I think it would be possible to use the same bot and add a way to filter by department.

abandoned-prototype:
I didn't previously see the "Questionnaire" functionality, where the bot basically fills the browse-officer form for you. This does indeed create some dependencies on the code. However, including the bot into the OpenOversight repo means that every time we make updates to the `Officer`, `Job` or `Department` model, or change the `BrowseOfficer` we need to make sure that the TwitterBot still works or we might unintentionally break it. That's why I think those two projects should be decoupled. Using the csv-download is just one idea to do that. We could implement other useful interfaces to the data that are actively maintained in the sense that they don't randomly change but are either kept backwards compatible or versioned. And those would also be useful for any other projects people want to start using the OpenOversight data but independently created and maintained

And yes, downloading&storing the most recent csv every 24h or something is additional work and maintenance, however I also mentioned that idea because you could then do the name-permutation work just once per day and store all the different versions, instead of creating them on the fly (something that I don't think is feasible with python and more than 20,000 officers stored in the database)

Finally, if you determine it is impossible to have the twitter-bot as a separate project, I think it should at least have a separate directory and not live in `app`, I think top-level would be best, an example of that would be what's done with the "flickrscraper". That way we can keep the dependencies/links between the two projects down to what's completely necessary

dismantl:
> However, including the bot into the OpenOversight repo means that every time we make updates to the Officer, Job or Department model, or change the BrowseOfficer we need to make sure that the TwitterBot still works or we might unintentionally break it.

Yes but that's the case with any code that depends on those models, which is most of OpenOversight. That's why I wrote exhaustive unit tests for the twitterbot that would indicate if any model or form changes caused the expected twitterbot behavior to break. I am fine with moving the twitterbot code to a top-level directory, tho I'll have to rethink how to hook into the testing infrastructure.

OpenOversight/app/utils.py
```diff
@@ -231,6 +326,16 @@ def compute_hash(data_to_hash):
...
 
 
+def compute_keyed_hash(str_to_hash):
+    h = hmac.new(key=current_app.config['SECRET_KEY'].encode('utf-8'), digestmod='sha256')
+    h.update(str_to_hash.encode('utf-8'))
+    return h.hexdigest()
+
+
+def verify_keyed_hash(digest, str_to_compare):
+    return digest == compute_keyed_hash(str_to_compare)
```
redshiftzero, line 336: instead of byte-by-byte comparison here, let's use `hmac.compare_digest`

redshiftzero (commented):
this bot is very cool! one note inline thus far

dismantl:
RIP Twitter

Issue #737, opened 2020-06-17
add type annotations


redshiftzero:
Once #736 is done, we should
* Add mypy to the dev dependencies
* Run mypy as part of the lint target to enable gradual typing
* Start adding [PEP 484](https://www.python.org/dev/peps/pep-0484/) style type annotations 

This is a good way for folks to get familiar with the codebase

egfrank:
Can I claim this issue? I haven't done type annotations before but it seems like based on mypy docs the order of operations would be:

- Add mypy to the dev dependencies
- Run mypy as part of the lint target to enable gradual typing
- Correct any errors that come from original mypy run
   - Ignore external modules that do not support type checking by adding them to the mypy.ini config
   - Add type annotations to OpenOversight code that errors in the mypy run
- Begin to add PEP 484 style type annotations for all function definitions

My instinct also be that initially we should just worry about files within OpenOversight/app so that would be the directory I would set for the lint target but if anyone has experience with using mypy on tests let me know!

abandoned-prototype:
Hi @egfrank. Yes, you can claim this issue!
In regards to your plan, I think adding mypy to dev requirements and linting can be done independently from adding type annotations. Also I am not sure which errors we can expect mypy to show before having added type-annotations, so fixing errors might be at the end of this, also depending on how many issues come out of it.
Finally, yes, focusing on files in `OpenOversight/app` makes a lot of sense to me

egfrank:
Hi, thanks for the response. I totally agree that `mypy to dev requirements and linting can be done independently from adding type annotations` and in fact I have commits for the dev requirements and linting if you want me to push those right away.

That being said, if you run mypy on the `develop` branch here are the errors you get (note: this is still using a mypy.ini config file that mutes errors on certain imports if it cannot figure out their type. If you ran mypy without this config I think you'd get ~40 more errors saying they cannot infer types from xxx package).
 
```
OpenOversight/app/models.py:26: error: Name 'db.Model' is not defined
OpenOversight/app/models.py:44: error: Name 'db.Model' is not defined
OpenOversight/app/models.py:64: error: Name 'db.Model' is not defined
OpenOversight/app/models.py:77: error: Name 'db.Model' is not defined
OpenOversight/app/models.py:90: error: Name 'db.Model' is not defined
OpenOversight/app/models.py:155: error: Name 'db.Model' is not defined
OpenOversight/app/models.py:170: error: Name 'db.Model' is not defined
[...]

PR #830, opened 2020-09-14
737A Add dev dependencies for mypy and add to lint make target

egfrank:
## Status

Ready for review!

## Description of Changes

Adds mypy to the dev-requirements
Adds mypy.ini file and copies it in to container in Dockerfile
Adds mypy check to `make lint` 

This is the first step outlined in issue 737, specifically mentioned in this comment
https://github.com/lucyparsons/OpenOversight/issues/737#issuecomment-685209038

## Tests and linting

Note: flake8 tests pass but the (new) mypy tests DO NOT. Also I would like to note that the way the `make lint` command is written that both `flake8` and `mypy` should run, even if the `flake8` check brings up errors. 

In another PR (on the not yet pushed branch `737_resolve_mypy_errors`) I will push changes to the actual Python files that will resolve the mypy errors currently being raised. 

## Question
Do we actually need the `start` command to be called within `make lint`?  https://github.com/lucyparsons/OpenOversight/blob/develop/Makefile#L53

Should it be empty, or `make: build` instead? I don't think we need the services running to do the linting, but maybe we do need to build first?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,8 @@ test: start  ## Run tests
 	fi
 
 .PHONY: lint
-lint: start
-	docker-compose run --rm web flake8
+lint: 
+	docker-compose run --rm web /bin/bash -c 'flake8; mypy app --config="../mypy.ini"'
 
 .PHONY: cleanassets
 cleanassets:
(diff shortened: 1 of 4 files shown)


abandoned-prototype:
I am closing this in in favor of #831 which contains the changes of this PR as well

PR #831, opened 2020-09-14
737B resolve mypy errors

egfrank:
## Status

Ready for review!

## Description of Changes

This PR (https://github.com/lucyparsons/OpenOversight/pull/830) added mypy type checking as part of the linting for this code-base. It did not however, resolve the original errors said linting surfaced. This PR makes changes to files within `OpenOversight/app` , adding type annotations and therefore resolving those issues.

The changes in  https://github.com/lucyparsons/OpenOversight/pull/830 are also on this branch.

 - Add dateparser.parser as a specific import to `model_imports.py` (I think just importing dateparser is a bug?)
 - Add type annotations for two Dictionaries in `commands.py` 
 - Import and use the DefaultMeta class from flask_sqlalchemy to have types for the db.Model 
   - I saw how to do this via https://github.com/dropbox/sqlalchemy-stubs/issues/76#issuecomment-595839159
- Define certain types for the base ModelView

All of these changes are addressing this issue  https://github.com/lucyparsons/OpenOversight/issues/737

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ test: start  ## Run tests
 
 .PHONY: lint
 lint: 
-	docker-compose run --rm web /bin/bash -c 'flake8; mypy app --config="../mypy.ini"'
+	docker-compose run --no-deps --rm web /bin/bash -c 'flake8; mypy app --config="../mypy.ini"'
 
 .PHONY: cleanassets
 cleanassets:
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,8 @@ test: start  ## Run tests
 	fi
 
 .PHONY: lint
-lint: start
-	docker-compose run --rm web flake8
+lint: 
+	docker-compose run --rm web /bin/bash -c 'flake8; mypy app --config="../mypy.ini"'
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -50,8 +50,8 @@ test: start  ## Run tests
 	fi
 
 .PHONY: lint
-lint: start
-	docker-compose run --rm web flake8
+lint: 
+	docker-compose run --rm web /bin/bash -c 'flake8; mypy app --config="../mypy.ini"'
```
abandoned-prototype, line 54: Great addition! I think we should add the `--no-deps` flag, otherwise docker starts up the postgres image which we don't need here

abandoned-prototype:
Thanks for this @egfrank. This looks & works great. I have added one comment on the `make lint` command.
Now we have even more reason to move on from python3.5 to change those comments to variable annotations

Makefile
egfrank: Ah thank you, just pushed that change.

abandoned-prototype (approved):
Thanks!

PR #833, opened 2020-09-25
#711: normalize gender in DB and forms, improve filter by gender behavior

### LINKED ISSUE #711: Search doesn't include results with null fields
In the testing instance for Vermont, we have one department that never gave us demographics data. So age, race, and gender are null for all the cops in this department. If a user tries to make a search that is keyed on one of these fields, they will get zero results. (For example, if they try to search for all female cops in one of these departments). This is a problem because it could cause users to not find officers who are indeed in OpenOversight.

To fix this, the search functionality should be changed so that, for example, when a user searches for all female cops, the application will return all female cops, plus all cops with a null gender.

EMCain:
## Status

Note: this is being merged from @juholgado 's branch but @emcain is writing up this PR. 

I'm recognizing that we have been distracted from this project (hello, entire West Coast being on fire!) and I want to make sure that the work we have done can be used if it is still needed, I'm happy to make changes to it but I think we need some help identifying what is left, can we get some help with that please?

Also, as per the discussion on #711 we are recognizing that a lot of the tests are broken for reasons unrelated to our changes and need help fixing this. We have added new tests but seemingly unrelated ones have stopped passing. 

## Description of Changes

Fixes #711 (partial) and some related DB issues. 

The original problem described in the issue was that for departments that are missing some demographic data on their officers (such as gender or race), trying to search on one of those fields would exclude all the officers with the missing data, when in fact they should be included in all searches (assuming some other filter doesn't rule them out). We chose to focus on gender as an example to be followed for other demographic categories, which could have similar implementations. Further research found that there were problems with how the database handled gender; there was no normalization of the gender column despite the fact that the filtering capabilities assume gender will fit into one of a few categories, and furthermore, the search-form concept of "not sure" (i.e. don't filter by this field because I, the end user, don't know) was being conflated with the demographic data missing for a given officer (a concept properly represented by NULL/None, but which was being stored as the string "Not Sure"). 

 - Change the database concept of "not sure" to NULL (which more accurately represents the situation where the project maintainers do not have the data, as opposed to the end user not knowing or remembering) 
 - require officer data saved to the database to fit into one of 3 gender labels (M, F, Other) or NULL; enforce this as a database constraint, and normalize data incoming from forms and CSV input 
- note that when officers are being newly added via CSV a missing gender will be saved as NULL, but when existing officers are updated via CSV, a non-null gender value will not be overwritten by an empty value in the spreadsheet even as other fields are being updated. 
 - change the officer lookup form to disallow selecting more than one gender value at the same time (either people want to filter by gender or they don't; this avoids some ambiguity on how to handle multiple boxes being checked. Note that this particular fix does not carry over well to categories such as race and age) *NOTE: this was left out due to Git problems before but it is now included*

If this PR is successfully merged, a good next step (for me or whoever has time afterward) would be to write up more detailed issue tickets describing how to implement the same sorts of changes for race, age, rank, etc. 

## Notes for Deployment

There's a migration that enforces the gender constraint and normalizes existing data, so for if some reason officers are listed as "man" and "woman" or something similar in your DB, they will be converted to the correct value instead of just being set to null. 

## Tests and linting

 - [x] `flake8` checks pass

Thanks to everyone who supported us on getting started on this ticket! I'm sorry this PR isn't cleaned up better, but I think I'd prefer to work with you to figure out what needs to happen yet than continue putting it off and possibly forget about it.


OpenOversight/migrations/versions/cd39b33b5360_constrain_officer_gender_options.py
```diff
@@ -0,0 +1,73 @@
...
+revision = 'cd39b33b5360'
+down_revision = '562bd5f1bc1f'
+branch_labels = None
+depends_on = None
+
+def get_update_statement(normalized, options):
+    return f"""
+    UPDATE officers 
+    SET gender = '{normalized}'
+    WHERE gender::citext in ({', '.join(["'" + o + "'" for o in options])});
```
EMCain, line 22: `gender::citext` just means the match is case-insensitive

EMCain: so this generates statements that say, basically, "set gender to M if the existing gender is 'male', 'm', or 'man' and ignore case" (but for each gender option)

abandoned-prototype (commented):
Thank you so much @EMCain and @juholgado. This looks great and both the database migration and the csv import work well.

> change the officer lookup form to disallow selecting more than one gender value at the same time (either people want to filter by gender or they don't; this avoids some ambiguity on how to handle multiple boxes being checked. Note that this particular fix does not carry over well to categories such as race and age)

Was this change also included? Which view are you referring to here?

For the tests and formatting, I can offer to take those parts over (since especially the test-stuff is a little tricky since it involves stuff you haven't touched at all as you mentioned) and I don't want to bog you down with that. In that case I would put a PR on @juholgado's repo and we can discuss the changes I am proposing there. Other options I can think of are for me to try to identify issues and report them to the two of you to fix them, or we do some kind of remote pair-programming if you prefer that. Also open to any other ideas

EMCain:
> Thank you so much @EMCain and @juholgado. This looks great and both the database migration and the csv import work well.
> 
> > change the officer lookup form to disallow selecting more than one gender value at the same time (either people want to filter by gender or they don't; this avoids some ambiguity on how to handle multiple boxes being checked. Note that this particular fix does not carry over well to categories such as race and age)
> 
> Was this change also included? Which view are you referring to here?
> 

Oh you’re right. This must have been on another branch we didn’t merge properly. I’ll look into cherry picking the changes into this one. 

> For the tests and formatting, I can offer to take those parts over (since especially the test-stuff is a little tricky since it involves stuff you haven't touched at all as you mentioned) and I don't want to bog you down with that. In that case I would put a PR on @juholgado's repo and we can discuss the changes I am proposing there. Other options I can think of are for me to try to identify issues and report them to the two of you to fix them, or we do some kind of remote pair-programming if you prefer that. Also open to any other ideas

You making a PR onto @juholgado’s branch is probably best but I want to figure out what happened with the form change mentioned above first so we don’t get merge conflicts, as well as addressing your other feedback.

EMCain:
Ahhh, we forgot to merge this. I also want to rewrite some of the commit history to make sure my personal account is linked to it properly. I’ll ping you once these changes are made and the feedback you left has been addressed. 

https://github.com/juholgado/OpenOversight/pull/1

OpenOversight/app/utils.py
```diff
@@ -617,3 +617,15 @@ def prompt_yes_no(prompt, default="no"):
...
+def normalize_gender(input_gender):
+    unnormalized_genders = {
+        "M": ('male', 'm', 'man'),
+        "F": ('female', 'f', 'woman'),
+        "Other": ('nonbinary', 'other'),
+    }
+
+    for key, value in unnormalized_genders.items(): 
+        if input_gender.lower().strip() in value:
+            return key
```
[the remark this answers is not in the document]
EMCain: done. And I had it just return None if None is input

OpenOversight/migrations/versions/cd39b33b5360_constrain_officer_gender_options.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = 'cd39b33b5360'
+down_revision = '562bd5f1bc1f'
+branch_labels = None
+depends_on = None
+
+def get_update_statement(normalized, options):
+    return f"""
```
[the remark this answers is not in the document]
EMCain: I changed it to use `.format()` -- I didn't get a chance to test it, I can if you want but I'll have to re-set up my Docker containers. Can you double check that it works now?

EMCain:
update: I fixed the git history (used the email address I want for my own commits), addressed the PR feedback, and rebased against develop. @abandoned-prototype can you look at this again? I can do some more manual testing if you like but I'll have to re-set up my dev environment to use the right Python version. [side note: I don't know why the docker-compose config doesn't force it to use Python 3.5 but that might be a thing to look into in the future]

abandoned-prototype:
Thanks for the update @EMCain! I put up a PR that fixes most of the tests and includes some commens here: https://github.com/juholgado/OpenOversight/pull/2/, once that is merged it should automatically show up on this PR and the tests should re-run. Fortunately the issue we saw last time with tests failing based on different random results was addressed in the meantime it seems.

There is one failing test left, but that test pointed me to something that doesn't seem to work right now and that is adding or editing an officer with unknown gender. When I select 'Not Sure' in the drop-down and submit I receive an error message
`Invalid value, must be one of: M, F, Other, None.` Additionally the form also doesn't seem to pre-load the right value, when the officer's gender is set to `NULL/None`. This is was the failing test seems to detect.

Let me know if you can reproduce that error and if that's something you want to fix (or maybe it's one other commit that wasn't included) or if you want me to take a stab at it.

EMCain:
I am going to look into fixing this error this weekend; I think I know roughly what part of the app you are talking about and I would like to make sure it works properly.

EMCain:
ok, looking like these are the next steps: 

1. review and merge or give feedback on juholgado#2 
2. reproduce and fix this error: 

> There is one failing test left, but that test pointed me to something that doesn't seem to work right now and that is adding or editing an officer with unknown gender. When I select 'Not Sure' in the drop-down and submit I receive an error message
Invalid value, must be one of: M, F, Other, None. Additionally the form also doesn't seem to pre-load the right value, when the officer's gender is set to NULL/None. This is was the failing test seems to detect.

EMCain:
looks like this is the failing test: 

```

    def test_edit_officer_form_coerces_none_race_or_gender_to_not_sure(mockdata, browser):
        # Set NULL race and gender for officer 1
        db.session.execute(
            Officer.__table__.update().where(Officer.id == 1).values(race=None, gender=None))
        db.session.commit()
    
        login_admin(browser)
    
        # Nagivate to edit officer page for officer having NULL race and gender
        browser.get("http://localhost:5000/officer/1/edit")
    
        wait_for_element(browser, By.ID, "gender")
        select = Select(browser.find_element_by_id("gender"))
        selected_option = select.first_selected_option
        selected_text = selected_option.text
>       assert selected_text == 'Not Sure'
E       AssertionError: assert 'Male' == 'Not Sure'
E         - Male
E         + Not Sure

````
Something about the form isn't working as expected, or else I am not using the artificial browser correctly. I'll try and reproduce this by hand before worrying about fixing it in the test.

**Comment by EMCain** [comment]
the instructions in the contributor guide on running a single tests are causing the tests to fail, where they would pass with make_test. 

https://github.com/lucyparsons/OpenOversight/blob/develop/CONTRIB.md#debugging-openoversight---use-pdb-with-a-test

`make test` causes the tests to pass but it means I have to run all of them. Probably worth looking at this next time: 

https://stackoverflow.com/questions/38411619/autotools-how-to-run-a-single-test-using-make-check

**Comment by EMCain** [comment]
ok, after a lot of frustration trying to fix the bug with the EditOfficerForm, here's what I've landed on as a solution for this and related issues. 

[x] Rearrange the GENDER_CHOICES constant to have "Not Sure" at the start. This is sort of a hacky workaround to get null-gender officers to default to "Not Sure" in that form, but arguably other forms that use this list (e.g. the officer lookup forms) should also default to "Not Sure." 
[x] fix the EditOfficer form to both take input properly and to default to the correct value.
[ ] simplify the coerce_gender function into a lambda and use that as the argument to EditOfficerForm 
[ ] update the other forms that have a gender field to use similar arguments to the one in EditOfficerForm 
[ ] fix the officer detail page to render "None" as something like "Information Missing" or "Unknown"
[ ] do the same for the list page when you browse for officers 
[ ] double check that the filtering features still work, both in the wizard for single officer lookup and in the filtering page you get after completing it

File: OpenOversight/app/models.py [hunk 4]
```diff
@@ -135,6 +139,8 @@ def race_label(self):
                 return label
 
     def gender_label(self):
+        if self.gender is None:
+            return 'Data Missing'
```
abandoned-prototype, line 143: Differentiating between missing data on our side and the user not knowing or caring is a great idea! And I think it can make some things much clearer. I like displaying `'Data Missing'`, the other options I see are either leaving it blank or skipping the display in the table altogether.
However, I don't think we need to decide on one way now, so this is definitely fine. Once we did similar work on other demographic data we always have the option of going back and changing it to a different behavior.

OpenOversight/migrations/versions/86eb228e4bc0_refactor_links.py
```diff
@@ -1,7 +1,7 @@
 """Refactor links
 
 Revision ID: 86eb228e4bc0
-Revises: 562bd5f1bc1f
+Revises: 79a14685454f
```
abandoned-prototype, line 4: good catch!

abandoned-prototype (changes requested):
Thanks for all this work @EMCain. Really appreciate that you stuck to this PR even when it grew bigger than we anticipated and encountered some hurdles from the code-base in general. Thanks for also taking care of some documentation and the additional tests!
This is basically ready to go. Just needs the three `flake8` errors fixed and the extra `\` removed

OpenOversight/app/models.py
EMCain: yeah, eventually I'd like to see a unified approach to these issues across demographic categories, and to make sure the UI makes very clear what's going on to the user.

OpenOversight/app/main/forms.py
```diff
@@ -17,6 +17,14 @@
 import datetime
 import re
 
+# Normalizes the "not sure" option to what it needs to be when writing to the database.
+# Note this should only be used for forms which save a record to the DB--not those that
+# are used to look up existing records.
+db_genders = list(GENDER_CHOICES)
+for index, choice in enumerate(db_genders):
+    if choice == ('Not Sure', 'Not Sure'):
+        db_genders[index] = (None, 'Not Sure')  # type: ignore
```
EMCain, line 26: I added the type: ignore to get flake8 to stop complaining about this line with "Incompatible types in assignment" -- we do in fact want None instead of a string for this one particular instance.

abandoned-prototype (approved):
Great work! Thanks again @EMCain

PR #838, opened 2020-10-09
Use ubuntu image for backup script

### LINKED ISSUE #740: backups are way too big
Our backup strategy at this point involves fetching gigabytes of data. We should do at least one of the following:

1. Emit progress output to prevent backups running in circle from failing due to no build output (if there isn't output in a 10 minute window, the job will fail)
2. Prune trash images to reduce backup size: #205
3. Incremental backups
4. Something else?

r4v5:
I _think_ this is broken because we attempt to do the backup in an image
that doesn't have pyenv but just has python 3.6.

@redshiftzero is there a way to test this config.yml before it gets to the main branch? I can't see anywhere in circleci that would make it possible.

## Status

Ready for review

## Description of Changes

Fixes #740 maybe? or at least moves it closer to happening.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,7 +85,7 @@ jobs:
 
   production_backup_only:
     docker:
-      - image: circleci/python:3.6
+      - image: ubuntu-1604
     steps:
       - checkout
       - *setup_dependencies
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,8 +84,9 @@ jobs:
             fab production migrate
 
   production_backup_only:
-    docker:
-      - image: ubuntu-1604
+    machine:
+      enabled: true
+      image: ubuntu-1604:201903-01
     steps:
       - checkout
       - *setup_dependencies


abandoned-prototype:
Not familiar with circleci (just started looking it at now), but could adding it to the workflow in the config.yaml for this branch (temporarily) help with that:
```
  tmp_test_workflow:
    jobs:
      - production_backup:
          filters:
            branches:
              only: backup-attempt-at-fix
```
or whatever the correct command is

redshiftzero:
yep @abandoned-prototype is right, adding a temporary commit to run just the modified job is the best way to test

One caveat, for the production backup the job will fail (also) due to the size of the backup. If there's no build output for a period of time - 10 mins by default (one can modify this but a production backup currently takes at least several hours) - then circleci will fail the build.

r4v5:
for future reference this did successfully back up, the build is https://app.circleci.com/pipelines/github/lucyparsons/OpenOversight/317/workflows/a06b9ad9-5ae5-4b0d-b606-38fe6ec36d52 (but the commit this build was attached to, 
f997b4e, is about to be rebased to drop the other commit testing backups)

PR #845, opened 2020-10-22
Whitelist .webp for image upload

brianmwaters:
## Status

Ready for review

## Description of Changes

Enable uploading .webp images.


abandoned-prototype (approved):
I checked that `.webp` is covered by `imghdr.what` which is used in the crop-image functionality when tagging the image. I don't have AWS set up locally, so we should make sure to test this functionality in staging.

PR #876, opened 2021-09-26
Verify two dependencies

pierwill:
Add checksum verifications for two dependencies downloaded in a Dockerfile. I've computed the hashes locally using `sha256sum`. They will have to be recomputed when updated. This also pulls the node setup script from version control to avoid breaking the verification if the upstream version is updated.

Looks like Mozilla actually does [publish](https://github.com/mozilla/geckodriver/releases) cryptographic signatures for gecko releases. We can use the checksum method in this PR for simplicity, or I'm happy to update this to use the signature.

### DIFF
--- a/dockerfiles/web/Dockerfile
+++ b/dockerfiles/web/Dockerfile
@@ -8,25 +8,36 @@ ENV CURL_FLAGS="--proto =https --tlsv1.2 -sSf -L --max-redirs 1 -O"
 
 ENV DEBIAN-FRONTEND noninteractive
 ENV DISPLAY=:1
+
+# install apt dependencies
 RUN echo "deb http://deb.debian.org/debian stretch-backports main" > /etc/apt/sources.list.d/backports.list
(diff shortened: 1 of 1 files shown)


dismantl:
I ran into an error when building the Docker image:

```
=> ERROR [12/25] RUN npm install -g yarn                                                                                                                                                                                                                0.2s
------
 > [12/25] RUN npm install -g yarn:
#15 0.228 /bin/sh: 1: npm: not found
------
failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c npm install -g yarn]: exit code: 127
make: *** [Makefile:7: build] Error 17
```

pierwill:
Try now? 😇

dismantl (approved):
Works well for me now. I tried changing the versions of node and gecko driver and they both failed to build at the checksum step.

abandoned-prototype:
Great work, thanks for this @pierwill. And thanks for testing @dismantl

PR #877, opened 2021-09-30
867/add markdown incident description

### LINKED ISSUE #867: Enable markdown for Incident Description
Currently the incident description just throws the entered sanitized text between a set of `<p></p>`. For some flexibility it would be nice to maybe allow markdown for that field, so some formatting can be applied. Alternatively we could add different sections to the table, but I think there is a chance that people from different cities want different fields displayed, so markdown would allow us to relatively easily set that up in an import script, or add some structure to a longer description. We would have to make sure to not allow injection of arbitrary html scripts however.

A different, somewhat lower tech solution could be to just allow a certain subset of features (maybe just headers), or to make sure that line breaks in the provided text are shown as line breaks in the result

abandoned-prototype:
## Status

Ready for review 

## Description of Changes

Fixes #867.

 - Incident description accepts markdown and some html while making sure that unsafe html cannot be introduced

## Tests and linting

 - [x] `flake8` checks pass

### DIFF
--- a/OpenOversight/app/__init__.py
+++ b/OpenOversight/app/__init__.py
@@ -3,6 +3,8 @@ import logging
 from logging.handlers import RotatingFileHandler
 import os
 
+import bleach
+from bleach_allowlist import markdown_tags, markdown_attrs
 from flask import Flask, render_template
 from flask_bootstrap import Bootstrap
 from flask_limiter import Limiter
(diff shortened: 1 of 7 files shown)


dismantl (changes requested):
The tests ran successfully but when I tried adding an officer description with markdown it didn't render:
image

dismantl:
Unless the markdown support is only supposed to be for incident descriptions and not officer descriptions? It appears to work for incident descriptions.

abandoned-prototype:
> Unless the markdown support is only supposed to be for incident descriptions and not officer descriptions? It appears to work for incident descriptions.

Yeah, I only implemented it for incidents, but I agree, it makes sense to have the same thing for notes & descriptions for officers as well. Also pretty easy to add with the jinja filter. I am about to update it now

PR #885, opened 2022-05-19
Fix CircleCI automated tasks

r4v5:
Update circleci image to something that's still under development

## Status

Ready for review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,10 +5,10 @@ steps:
     run:
       name: Install dependencies (required for fabric)
       command: |
-        pyenv global 3.5.2
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


r4v5:
Staging successfully backed up in https://app.circleci.com/pipelines/github/lucyparsons/OpenOversight/486/workflows/c95f5d5e-eba6-43b7-9f17-ef40b3e74c9b/jobs/261 ; deployment failed because of the move to docker meaning it can't use the standard docker-compose flow.

That all is going to get reworked as part of the move to poetry and later work; in the meantime, this at least fixes backups again.

abandoned-prototype (approved):
Don't have the full background here, but looks reasonable to me. Thank you very much for this work!

PR #893, opened 2022-09-04
Switch to using docker and poetry

r4v5:
## Status

in progress

## Description of Changes

 - Move us to using Poetry instead of pip for dependencies
 - Add a "production" dockerfile with only runtime deps
 - switch from node-sass to dart sass

## Tests and linting

 - [ ] `flake8` checks pass


michplunkett:
Is this currently in progress? If so, I'd be happy to contribute to it. I've found Poetry to be quite useful.

abandoned-prototype:
> Is this currently in progress? If so, I'd be happy to contribute to it. I've found Poetry to be quite useful.

I would say no, this is no longer worked at, but I know @onemec has been working on setting up Poetry as well. So we should coordinate to make sure we are not duplicating work

michplunkett:
If you could use an extra hand at some point, @onemec, please let me know.

PR #899, opened 2023-01-05
Switch from docker to "machine" for circle-ci

abandoned-prototype:
## Status

Ready for review

## Description of Changes

Hopefully fixes issues with the circle ci deploy script. The script fails due to attempting to run docker-compose inside the circle ci docker image which is not supported. Switching back to a "machine" (with a ubuntu version that is still current) will hopefully make the scripts run successfully.

## Tests and linting

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,8 +46,8 @@ steps:
 
 jobs:
   staging_backup_and_deploy:
-    docker:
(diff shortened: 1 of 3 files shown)


r4v5 (approved):
this'll work; there's also a "set up remote docker" action somewhere we can use, but I'm fine just going for VMs due to our use of compose.

PR #907, opened 2023-03-11
New server, deployment, CI/CD pipeline

abandoned-prototype:
## Status

Ready for review

## Description of Changes

 - Setting up new servers that run OpenOversight as docker image instead of in a python environment.
   - This should allow us to more easily keep the python version updated and quickly move from 3.8 to 3.11
 - Changing the strategy of deploying the code to the server. The overall idea will stay the same: A push to develop will trigger a deployment to the staging server, pushing to main triggers a deployment to production.
 - The way this is executed has been updated.
   - Instead of circleCI I propose to use github actions to do the deployment. The reasons for that are that I think it has a larger number of people using it (which means there are more tutorials, more people creating actions, etc), that the results are better visible to everyone (I think), that no secrets have been stolen recently (https://circleci.com/blog/jan-4-2023-incident-report/) and we are already using it for tests.
   - Speaking of tests, tests will no longer run on push, only on pull requests. As long as every push to either develop or main is done via a PR this should be fine.
   - I also updated Fabric to its new version Fabric 2. We use fabric to relatively easily write scripts that connect from the github action server to our server via ssh and run the appropriate commands to create backups and deploy the code.
 - This PR also includes a necessary change to dart sass, backported from OrcaCollective (Thank you!!)
- I was testing the github action in a separate branch to make sure everything works as intended. Here is the successful run that should be the same as pushing to develop or staging should look like: https://github.com/lucyparsons/OpenOversight/actions/runs/4394148748. The branch is here: https://github.com/lucyparsons/OpenOversight/tree/workflow-test

## Notes for Deployment
Before merging, I actually need to change the name-server settings so that staging.openoversight.com points to the new server.

## Tests and linting


.circleci/config.yml
```diff
@@ -1,116 +1,17 @@
+# unused - remove file once circleCI integration is removed
```
onemec, line 1: I might be mistaken, but should this file now be removed assuming it encompasses a full shift away from CircleCI?

.github/workflows/test_prs.yml
```diff
@@ -24,7 +19,7 @@ jobs:
       FLASK_APP: OpenOversight.app
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8]
+        python-version: ["3.8", "3.9"]
```
onemec, line 22: Your PR vaguely mentioned eventually shifting to Python 3.11. Do we want to either add a bigger test matrix, or a comment saying we'll eventually expand the range here? 

Also, on a similar note, I think it's worthwhile to specify the current Python version we're using in the Dockerhub description or tags or somewhere for issue traceability.

onemec:
Looks good to me overall, though I'll try to review in greater detail later this week.

OpenOversight/app/static/scss/bootstrap/_variables.scss
```diff
@@ -1,3 +1,5 @@
+@use "sass:math";
```
michplunkett, line 1: This plugin is so fresh!

OpenOversight/scripts/entrypoint_dev.sh
```diff
@@ -0,0 +1,3 @@
+yarn build
```
michplunkett, line 1: I'm not sure if it's absolutely necessary, but since you have it in the prod version, you should add `#!/bin/bash` to the top of this file. 

It's also be good to add comments on what it's doing like you did in the prod version.

OpenOversight/scripts/entrypoint_prod.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Compile any assets
```
michplunkett, line 2: Some gaps between functionality and bash commands would make this easier to read.

package.json
```diff
@@ -1,14 +1,12 @@
 {
   "devDependencies": {
-    "jquery": "1.9.1 - 3",
```
michplunkett, line 3: I appreciate that you were able to get rid of a good amount of plugins here, especially `jquery`. It's crazy how long that plugin has stuck around.

dockerfiles/web/Dockerfile-prod
```diff
@@ -0,0 +1,43 @@
+FROM python:3.8-slim-bullseye as base
```
michplunkett, line 1: I think some general comments here would be helpful.

tasks.py
```diff
@@ -0,0 +1,154 @@
+"""
```
michplunkett, line 1: Though many of these functions are quite intuitive, I think it'd be worthwhile to at least have some function level comments on them. You don't need to put parameter level comments, though I do prefer them, but I think we should have at least Class, function, etc. comments in place.

michplunkett (commented):
Generally great with a few nits here and there.

.github/workflows/test_prs.yml
michplunkett: It might be worth it to see how far you can push this matrix. See if `3.10` and `3.11` work in the repository's current state, etc.

OpenOversight/scripts/entrypoint_dev.sh
AetherUnbound: +1 to this!

docker-compose.prod-img.yml
```diff
@@ -0,0 +1,18 @@
+version: "3.4"
+
+services:
+  web:
+    image: ghcr.io/lucyparsons/openoversight:${DB_IMAGE_TAG}
```
AetherUnbound, line 5: This doesn't seem to correspond to a database, should this be `DOCKER_IMAGE_TAG` instead?

tasks.py
```diff
@@ -0,0 +1,154 @@
...
+
+def migrate(c, config):
+    c.run(
+        f"DB_IMAGE_TAG={config.docker_version} docker compose -f docker-compose.prod-img.yml run --rm --no-deps web flask db upgrade",
+        echo=True,
+    )
+
+
+def deploy_(c: Connection, config: HostConfig):
+    print("Start new deployment.")
```
AetherUnbound, line 83: I know it's a little bit more work, but I'd recommend converting these from `print` to `log.info` statements and setting up a `basicConfig` to log at INFO level by default. This would help provide timestamps for when each statement is issued (I've found myself shaking my fist at myself in the past for not having timestamps when using `print` 😅).

tasks.py
```diff
@@ -0,0 +1,154 @@
...
+    with c.cd(config.code_dir):
+        pwd = io.StringIO(config.github_token)
+        print("Get new production docker image.")
+        c.run(
+            f"docker login ghcr.io -u {config.github_user} --password-stdin",
+            in_stream=pwd,
+            echo=True,
+        )
```
AetherUnbound, line 91: Very cool!

AetherUnbound (approved):
This looks great! I have a few comments, nothing blocking though 😄

.github/workflows/test_prs.yml
abandoned-prototype: Yeah, originally I thought I could make it go to 3.11, but both 3.10 and 3.11 failed some tests (https://github.com/lucyparsons/OpenOversight/actions/runs/4391087165). Probably just a library that needs an update version or something, but I decided I shouldn't try to fix that in this PR.

Good point about specifying the currently used version! I will see where I can add that.

.circleci/config.yml
abandoned-prototype: Originally I removed it, but then CircleCI shows up as a failed test with "no config file found" which I thought was distracting (and maybe even blocking the PR, not sure), so now I will remove the file once I removed that integration

OpenOversight/scripts/entrypoint_dev.sh
abandoned-prototype: Good points, thanks! I added `#!/bin/bash` and made executable. Also added comments.

OpenOversight/scripts/entrypoint_prod.sh
abandoned-prototype: You mean just adding a line here, right?

tasks.py
abandoned-prototype: Good suggestion. I added some doc-strings and task help which would show up when using `invoke --help deploy` for example

tasks.py
abandoned-prototype: Thank you, I agree that it is very important to have. Unfortunately python makes logging so weird by default (time is parsed with a comma instead of decimal for microseconds and in local time), so there are some extra lines to make it acceptable.

docker-compose.prod-img.yml
abandoned-prototype: Good find, totally blanked here. Definitely should have `DOCKER` in the name, thanks!

dockerfiles/web/Dockerfile-prod
abandoned-prototype: Agreed! I mostly copied this, but now looking at it again, I think I should be able to remove some of these packages. So I will do that first and then add comments why we need each of the remaining pieces.

OpenOversight/scripts/entrypoint_prod.sh
michplunkett: Yup! That's a much clearer way of putting it.

dockerfiles/web/Dockerfile-prod
abandoned-prototype: Added some comments now and removed the script that installs sqlite which is unused.

.github/workflows/test_prs.yml
abandoned-prototype: I also added your second point now @onemec. The python version is set in the workflow file and will also be present in the tag used for the package as can be seen in the example here from my workflow-test branch: https://github.com/lucyparsons/OpenOversight/pkgs/container/openoversight/versions (dev-py3.9). Thanks for the suggestion!

.github/workflows/build_deploy.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  push:
+    branches:
+      - develop
+      - main
+
+env:
+  image_name: lucyparsons/openoversight
+  # env_mapping: {"develop": "staging", "main": "prod"}
```
michplunkett, line 11: If these bits are gonna be commented out, I'd be game for just deleting them from the file.

OpenOversight/scripts/entrypoint_dev.sh
michplunkett: Noice!

tasks.py
```diff
@@ -0,0 +1,200 @@
...
+* deploy env github_user github_token
+    downloading the most recent image and code for the given environment,
+    then migrating the database if any migrations are necessary,
+    finally restarting the docker containers. The github user and token are
+    used to download the docker image from the github containe registry.
+* cleanup env
+    do cleanup tasks. At the moment this consists of removing unused docker
+    to free up space.
+
+
```
michplunkett, line 21: Nit: I think it'd be worth getting rid of these two empty new lines.

tasks.py
```diff
@@ -0,0 +1,200 @@
...
+    return str(datetime.datetime.utcfromtimestamp(record.created)) + "Z"
+
+
+for h in logging.getLogger().handlers:
+    h.formatter.formatTime = _formatTime
+
+
+@dataclass
+class HostConfig:
+    """HostConfig collects various parameters for each deployment (prod, staging, etc)"""
```
michplunkett, line 51: Nit: Top line documentation comments should look like:
```python
 """
HostConfig collects various parameters for each deployment (prod, staging, etc).
"""
```

It's easier to read and is more standard from my understanding.

tasks.py
```diff
@@ -0,0 +1,200 @@
...
+        """Return ssh connection as used by fabric for this HostConfig."""
+        return Connection(host=self.host_name, user=self.user)
+
+
+def get_configs(
+    env: str, github_user: str = None, github_token: str = None
+) -> HostConfig:
+    """Return a HostConfig based on given environment and github user and token.
+    Github user and token can be omitted if no connection to github is needed.
+    """
```
michplunkett, line 75: Nit:

```python
"""
Return a HostConfig based on given environment and github user and token.
Github user and token can be omitted if no connection to github is needed.
"""
```

michplunkett: The same goes for the rest of the docstrings.

tasks.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def cleanup_(c: Connection, config: HostConfig):
+    """Execute cleanup tasks, usually after deployment"""
+    logging.info("Remove unused docker images")
+    c.run("docker image prune -f")
+
+
+@task(
+    help={"environment": "'staging' or 'prod' indicating which environment to target"}
```
michplunkett, line 194: Nit: It looks like we use this string at least 3 times, so it'd be worth saving the key and value to a constant, imo.

michplunkett (approved):
Some small nits here and there, but otherwise fantastic work!

tasks.py
abandoned-prototype: The pydocstyle checker (part of our pre-commit check) appears to disagree. It's code D200 in this list https://www.pydocstyle.org/en/stable/error_codes.html

tasks.py
abandoned-prototype: Done

tasks.py
abandoned-prototype: Changed this one (because it is multi-line)

tasks.py
abandoned-prototype: I was thinking about that. There is the downside of the help strings potentially not being updated if something changes, but the plus is that it is also a comment to anyone reading the code, so I am leaning to keep it as it is

.github/workflows/build_deploy.yml
abandoned-prototype: I had preferred to keep this in the code (not put it in a variable) but I somehow couldn't make that work. I left this here so it's easier to for people to make sense of it, even if they cannot see the variable settings (which might require admin rights to see, not sure)

tasks.py
michplunkett: Works for me in that case. It looks like it's good to go!

tasks.py
michplunkett: I'm here for that.

abandoned-prototype:
I did a force push to squash the commits from the code review into the original commit to be able to do a merge commit now and maintain the original cherry-picked commit from the OrcaCollectiv's fork without adding multiple tiny commits to the projects history. The "Compare" button (https://github.com/lucyparsons/OpenOversight/compare/2a812ec9acf6c71d21831cd466e04fbda646d7a3..ea618af46af03408159f125153a4a883ffe8d356) verifies the before and after are identical.

tasks.py
abandoned-prototype: Yes, we do have pre-commit hooks in place, including the end-of-file-fixer (see https://github.com/lucyparsons/OpenOversight/blob/develop/.pre-commit-config.yaml) and they ran over this PR: https://github.com/lucyparsons/OpenOversight/actions/runs/4458996732/jobs/7831175755
These extra empty lines were in the middle of a file, not at the end

PR #910, opened 2023-04-02
Backport/follow up

abandoned-prototype:
## Status

Ready for review

## Description of Changes

Quick follow up to the recent backport work.

 - Fix an issue that prevented the placeholder image from being shown in prod environments
   - The problem was that _external used "http" as default scheme, but _external is not needed here I am pretty sure
 - Replace magic numbers (http status numbers) with httpstatus constants.

## Tests and linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -254,9 +254,7 @@ def officer_profile(officer_id):
             face_paths.append(serve_image(face.image.filepath))
         if not face_paths:
             # Add in the placeholder image if no faces are found
-            face_paths = [
-                url_for("static", filename="images/placeholder.png", _external=True)
-            ]
+            face_paths = [url_for("static", filename="images/placeholder.png")]
     except:  # noqa: E722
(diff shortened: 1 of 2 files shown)


OpenOversight/app/main/views.py
```diff
@@ -269,7 +267,6 @@ def officer_profile(officer_id):
             officer.image_url = url_for(
                 "static",
                 filename=faces[0].image.filepath.replace("/static/", ""),
-                _external=True,
```
michplunkett, line 272: I'm guessing these are by default set to `True`?

michplunkett (approved):
This def addresses what I asked for on the previous PR. I'll create a PR later this week addressing the use of ints instead of constants in `auth/views.py`, `main/views.py`, and throughout a lot of the test files.

OpenOversight/app/main/views.py
abandoned-prototype: It should default to False in our use case here. But it is a good point, I probably should have set it to False to make it clear

Issue #909, opened 2023-03-31
Update HTTP response codes with constants


michplunkett:
Within the python 🐍 codebase we have a lot of HTTP response codes that are just ints. They should be replaced with constants from the [`http` package](https://docs.python.org/3/library/http.html).

michplunkett:
https://github.com/lucyparsons/OpenOversight/pull/910#pullrequestreview-1368137413

michplunkett:
It's been addressed by the linked PR.

PR #911, opened 2023-04-05
Use HTTP constants instead of stray ints or strings

michplunkett:
## Status

Done.

## Description of Changes
Replaced stray response code ints with standard library `HTTPStatus` constants. Added `HTTP_METHOD` constants to the `utils.py` and replaced stray `GET`, `POST`, etc. strings with the aforementioned constants. I also removed some redundant syntax in the test files with specific regard to strings and lists.

Addresses #: https://github.com/lucyparsons/OpenOversight/issues/909

## Notes for Deployment
It should not change anything regarding deployment.

## Tests and linting


CONTRIB.md
```diff
@@ -10,7 +10,7 @@ If your git history is not so clean, please do rewrite before you submit your PR
 
 To submit your changes for review you have to fork the repository, push your new branch there and then create a Pull Request with `OpenOversight:develop` as the target.
 
-Use [PULL_REQUEST_TEMPLATE.md](/PULL_REQUEST_TEMPLATE.md) to create the description for your PR! (The template should populate automatically when you go to open the pull request.)
+Use [pull_request_template.md](/.github/pull_request_template.md) to create the description for your PR! (The template should populate automatically when you go to open the pull request.)
```
michplunkett, line 13: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository#adding-a-pull-request-template#adding-a-pull-request-template

OpenOversight/app/utils.py
```diff
@@ -45,6 +45,10 @@
 )
 
 
+# HTTP Method constants
+HTTP_METHOD_GET = "GET"
+HTTP_METHOD_POST = "POST"
```
michplunkett, line 50: These method constants come standard in python 3.11, but we aren't there yet. Whenever we get there, we can remove these.

OpenOversight/tests/routes/test_descriptions.py
```diff
@@ -13,36 +14,36 @@
...
 
 @pytest.mark.parametrize(
     "route",
     [
-        ("officer/1/description/1/edit"),
-        ("officer/1/description/new"),
-        ("officer/1/description/1/delete"),
+        "officer/1/description/1/edit",
+        "officer/1/description/new",
+        "officer/1/description/1/delete",
```
michplunkett, line 32: This sort of parenthesis is redundant, so I removed them.

CONTRIB.md
michplunkett: Adding it to this folder also removes clutter from the base folder while alluding to the role that the `pull_request_template.md` file plays.

OpenOversight/app/models.py
```diff
@@ -551,7 +552,7 @@ def verify_password(self, password):
 
     def generate_confirmation_token(self, expiration=3600):
         payload = {"confirm": self.id}
-        return self._jwt_encode(payload, expiration).decode("utf-8")
+        return self._jwt_encode(payload, expiration).decode(model_encoding)
```
michplunkett, line 555: Used a local constant because using the one in utils would have created a circular dependency.

abandoned-prototype: That makes sense. Looking at the utils file, it is a big mess. We should consider breaking that up by functionality or something to avoid circular dependencies. I will create an issue for that

OpenOversight/app/utils.py
abandoned-prototype: Cool, thanks for calling that out here

OpenOversight/tests/routes/test_descriptions.py
abandoned-prototype: Good call. I was surprised to learn that black doesn't do that already, so I quickly looked into it. Turns out the reason is that black sometimes uses redundant parenthesis to break things up into multiple lines which makes this a little more complicated

CONTRIB.md
abandoned-prototype: This makes sense to me. Thanks!
One general note: I prefer keeping changes that have nothing to do with each other separate, in this case either a separate PR, or rebasing/squashing this PR into just two commits so they can submitted via merge.
This for example makes cherry-picking specific commits easier, and it also makes it less likely that things are changed without any comment why

abandoned-prototype (approved):
Thank you. This looks good!

OpenOversight/app/models.py
michplunkett: Yeah, I was thinking of moving the stuff to a different package, but I felt that would be a much bigger more than necessary. I'll address that issue next.

CONTRIB.md
michplunkett: I can move this to a different PR so that everything is separate. One second!

OpenOversight/tests/routes/test_descriptions.py
michplunkett: Ayyyye, it does that a lot when the lines are longer than whatever the line limit is.

CONTRIB.md
michplunkett: https://github.com/lucyparsons/OpenOversight/pull/912

Issue #913, opened 2023-04-07
Restructure utils.py file to its own package


michplunkett:
## What do?
The `utils.py` file is overly large and needs to be trimmed down. In its current state it is  too interweaved now and is starting to create circular dependency issues. I plan to put it in its own package, have a file for constants, and several files for functions based on what they do.

PR #915, opened 2023-04-12
Restructure `utils.py` to its own package

michplunkett:
## Status
In progress

## Description of Changes
Changed `utils.py` file from a gigantic mono-file to a package with more specialized files.

Fixes: https://github.com/lucyparsons/OpenOversight/issues/913

## Tests and linting
 - [ ] `pre-commit` checks pass

### DIFF
--- a/OpenOversight/app/models.py
+++ b/OpenOversight/app/models.py
@@ -12,11 +12,11 @@ from sqlalchemy.orm import validates
 from werkzeug.security import check_password_hash, generate_password_hash
 
 from . import login_manager
+from .util.constants import ENCODING_UTF_8
 from .validators import state_validator, url_validator
 
 
 db = SQLAlchemy()
-model_encoding = "utf-8"
 jwt = JsonWebToken("HS512")
 
 BaseModel = db.Model  # type: DefaultMeta
@@ -552,7 +552,7 @@ class User(UserMixin, BaseModel):
 
     def generate_confirmation_token(self, expiration=3600):
         payload = {"confirm": self.id}
-        return self._jwt_encode(payload, expiration).decode(model_encoding)
+        return self._jwt_encode(payload, expiration).decode(ENCODING_UTF_8)
 
     def confirm(self, token):
         try:
@@ -572,7 +572,7 @@ class User(UserMixin, BaseModel):
 
     def generate_reset_token(self, expiration=3600):
         payload = {"reset": self.id}
-        return self._jwt_encode(payload, expiration).decode(model_encoding)
+        return self._jwt_encode(payload, expiration).decode(ENCODING_UTF_8)
 
     def reset_password(self, token, new_password):
         try:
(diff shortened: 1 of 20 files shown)


OpenOversight/app/models.py
```diff
@@ -12,11 +12,11 @@
 from werkzeug.security import check_password_hash, generate_password_hash
 
 from . import login_manager
+from .util.constants import ENCODING_UTF_8
 from .validators import state_validator, url_validator
 
 
 db = SQLAlchemy()
-model_encoding = "utf-8"
```
michplunkett, line 19: We could remove this now that it isn't a circular reference.

michplunkett:
Gonna run the tests on my local sometime tomorrow.

Makefile
```diff
@@ -89,5 +89,6 @@ help: ## Print this message and exit
 		| sort \
 		| column -s ':' -t
 
+.PHONY: attach
```
michplunkett, line 92: Added a missing `PHONY`.

abandoned-prototype (commented):
Firstly, Thank you very much for doing this!
One nit: Was it intentional to change `utils.py` to `util`, getting rid of the plural? It appears that generally we are leaning towards the plural in the naming here.

Finally, I did remember an issue that we ran into in the past, that prevented us from using absolute imports. I tested it locally and I believe it is still an issue, so I created https://github.com/lucyparsons/OpenOversight/pull/916 to change that. So in short, I #916 should be merged first and then this one should be fine :)

michplunkett:
That's good to know! I'll make the change from `util` to `utils` now!

OpenOversight/app/util/general.py
```diff
@@ -0,0 +1,164 @@
...
+
+def allowed_file(filename):
+    return (
+        "." in filename
+        and filename.rsplit(".", 1)[1].lower()
+        in current_app.config["ALLOWED_EXTENSIONS"]
+    )
+
+
+def get_or_create(session, model, defaults=None, **kwargs):
```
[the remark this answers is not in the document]
michplunkett: If we're good now, I think we can keep it where it's at. I'll add a note to the top of it tho signaling that it could be something we move to its own file should a circular import arise.

.gitignore
```diff
@@ -93,6 +93,9 @@ vagrant/puppet/.tmp
 .vscode/**
 *.code-workspace
 
+# Operating system files
+**/.DS_Store
```
michplunkett, line 97: I was getting this file if I opened up anything in finder on my comp.

OpenOversight/migrations/versions/59e9993c169c_change_faces_to_thumbnails.py
```diff
@@ -15,7 +15,7 @@
 sys.path.insert(0, os.path.dirname(current_app.root_path))
 
 from app.models import Face, db  # noqa: E402
-from app.utils import crop_image  # noqa: E402
+from app.utils.cloud import crop_image  # noqa: E402
```
michplunkett, line 18: I felt weird changing a migration file, but it seems it was necessary.

PR #921, opened 2023-05-04
Packages/test performance

abandoned-prototype:
## Status

Ready for review

## Description of Changes

 - For some reason the mockdata used in the tests not only contains a lot of data, but it was freshly inserted into the database for every test that used it.
   - Now we will load the mockdata into the in-memory database only once (via the fixture with scope session). This makes tests significantly faster.
   - Adding an additional department to the mockdata broke some tests which had to be fixed.
   - Changing anything in the mockdata brakes tests because some of the data is assigned pseudo-randomly (random with fixed seed), so a test might fail because an officer used to not have a link in the profile but does after the change. I fixed the failing tests in a way that this does not happen again.

## Notes for Deployment
 Changes tests only

## Tests and linting

### DIFF
--- a/OpenOversight/tests/conftest.py
+++ b/OpenOversight/tests/conftest.py
@@ -48,6 +48,10 @@ RANK_CHOICES_2 = [
     "Chief",
 ]
 
+DEPARTMENT_NAME = "Springfield Police Department"
+DEPARTMENT_TWO_NAME = "Chicago Police Department"
+DEPARTMENT_EMPTY_NAME = "Empty Police Department"
+
 
 AC_DEPT = 1
 
@@ -116,7 +120,7 @@ def pick_salary():
     return random.randint(100, 100000000) / 100
 
 
-def generate_officer():
+def generate_officer(department):
     year_born = pick_birth_date()
     f_name, m_initial, l_name = pick_name()
     return models.Officer(
@@ -127,7 +131,7 @@ def generate_officer():
         gender=pick_gender(),
         birth_year=year_born,
         employment_date=datetime.datetime(year_born + 20, 4, 4, 1, 1, 1),
-        department_id=pick_department().id,
+        department_id=department.id,
         unique_internal_identifier=pick_uid(),
     )
 
@@ -216,6 +220,13 @@ def db(app):
     with app.app_context():
         _db.app = app
         _db.create_all()
+        connection = _db.engine.connect()
+        session = scoped_session(session_factory=sessionmaker(bind=connection))
+        _db.session = session
+        add_mockdata(session)
+        session.commit()
+        connection.close()
+        session.remove()
 
         yield _db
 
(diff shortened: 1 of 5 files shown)


OpenOversight/tests/conftest.py
```diff
@@ -48,6 +49,10 @@
     "Chief",
 ]
 
+DEPARTMENT_NAME = "Springfield Police Department"
+DEPARTMENT_TWO_NAME = "Chicago Police Department"
+DEPARTMENT_EMPTY_NAME = "Empty Police Department"
```
michplunkett, line 54: 🔬 I think `DEPARTMENT_NAME_XXXX` makes more sense as a naming motif.

`DEPARTMENT_NAME` -> `DEPARTMENT_NAME_SPRINGFIELD`
`DEPARTMENT_TWO_NAME` -> `DEPARTMENT_NAME_CHICAGO`
`DEPARTMENT_EMPTY_NAME` -> `DEPARTMENT_NAME_EMPTY`.

michplunkett, line 54: When I see "empty" in the name, I presume it's an empty string.

OpenOversight/tests/conftest.py
```diff
@@ -310,15 +339,18 @@ def add_mockdata(session):
 
     test_images = [
         models.Image(
-            filepath="/static/images/test_cop{}.png".format(x + 1), department_id=1
+            filepath="/static/images/test_cop{}.png".format(x + 1),
```
michplunkett, line 342: 🔬 An f-string feels like a more intuitive way to write this, imo.

OpenOversight/tests/conftest.py
```diff
@@ -348,15 +381,23 @@ def add_mockdata(session):
...
+        for officer in officers_dept1[
+            : math.ceil(len(officers_dept1) * assignment_ratio)
+        ]
     ]
     assignments_dept2 = [
-        build_assignment(officer, test_units, jobs_dept2) for officer in officers_dept2
+        build_assignment(officer, test_units, jobs_dept2)
+        for officer in officers_dept2[
+            : math.ceil(len(officers_dept2) * assignment_ratio)
+        ]
```
michplunkett, line 400: 🔬  The syntax here doesn't feel incredibly intuitive.

OpenOversight/tests/conftest.py
michplunkett, line 54: Since these are paired to specific short names, I think it may be valuable to put them in a dictionary.

```python
DEPARTMENT_KEY_SPD = "SPD"
DEPARTMENT_KEY_CPD = "CPD"
DEPARTMENT_KEY_EPD = "EPD"

DEPARTMENTS = {
    DEPARTMENT_KEY_SPD: "Springfield Police Department",
    DEPARTMENT_KEY_CPD: "Empty Police Department",
    DEPARTMENT_KEY_EPD: "Chicago Police Department"
}
```

You can use the key as the short name (lower or upper case, whateva) and the value as the long name.

michplunkett (approved):
Looks good! I just have some formatting suggestions, but they aren't blockers.

OpenOversight/tests/conftest.py
michplunkett: On second thought, tuples would work as well.
```python
DEPARTMENT_SPD = ("SPD", "Springfield Police Department")
DEPARTMENT_CPD = ("CPD", "Chicago Police Department")
DEPARTMENT_EPD = ("EPD", "Empty Police Department")
```

abandoned-prototype: Changed it slightly, hopefully make it a little more sense. But I want to make it clearer which department is considered the default one.

abandoned-prototype: Good point. Changed it to `DEPARTMENT_WITHOUT_OFFICERS`.

OpenOversight/tests/conftest.py
abandoned-prototype: Strongly agree, this is from old code. Changed this one. And maybe we can use https://github.com/ikamensh/flynt or something next to get rid of all of the `.format`.

OpenOversight/tests/conftest.py
abandoned-prototype: Tried to make it a little better, by extracting "the math" to a variable and adding a comment.

OpenOversight/tests/conftest.py
abandoned-prototype: The department short name does not seem to matter much for tests (they are not used anywhere other than the initial creation) so I decided against giving each their own variable.

Issue #914, opened 2023-04-07
Fix COVERALLS link 


michplunkett:
## What do?
The COVERALLS link for this repo has been broken since `May 7, 2021.` If we are going to keep it posted on our `README.md`, it should be working.

michplunkett:
Seems like it has already been referenced: https://github.com/lucyparsons/OpenOversight/issues/730

michplunkett:
I re-synced Coveralls with the repo, but I'm not sure if that did much.

PR #933, opened 2023-06-26
We are only using versions 3.8 and 3.9 in our builds

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/914

## Description of Changes
Coveralls has not been working the past couple of years and it appears the same can be said for Travis and potentially CircleCI. Gonna update things so that Coveralls works again and we delete any files that no longer need to be there.

## Changes proposed in this pull request:
 - Fix [Coveralls](https://coveralls.io/github/lucyparsons/OpenOversight) for OO
 - Remove CircleCI and/or Travis integration if they are no longer being used.

## Tests and linting
 - [ ] `pytest` passes on my local development environment.
 - [ ] `pre-commit` passes on my local development environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,8 @@ language: python
 dist: xenial
 sudo: required
 python:
-  - "3.5"
-  - "3.6"
-  - "3.7"
   - "3.8"
+  - "3.9"
 services:
   - docker
 env:


.travis.yml
```diff
@@ -2,10 +2,8 @@ language: python
 dist: xenial
 sudo: required
 python:
-  - "3.5"
-  - "3.6"
-  - "3.7"
   - "3.8"
+  - "3.9"
```
michplunkett, line 6: We are only using `3.8` and `3.9` in our builds, so there's no reason to include the older versions for Travis.

.travis.yml
```diff
@@ -1,27 +0,0 @@
-language: python
```
michplunkett, line 1: From what I'm seeing, we have not completed a Travis build in the last two years, so we should be able to get rid of this file and presumably the integration altogether.
Screenshot 2023-06-26 at 11 34 39 AM

docker-compose.prod-img.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.4"
-
```
michplunkett, line 2: Screenshot 2023-06-26 at 11 44 27 AM

.travis.yml
michplunkett, line 1: We are no longer using this integration, so it's scrapped.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,41 @@
+name: Test Coverage
```
michplunkett, line 1: Created using [this service](https://akx.github.io/travis-to-github-actions/) and using the deleted `.travis.yml` file as the template.

Issue #239, opened 2017-04-10
Add HTML formatting and linting to `pre-commit`


redshiftzero:
Once we find a good linter for HTML, we should add it to CI. 

I was using [bootlint](https://github.com/twbs/bootlint) and [html-linter](https://pypi.python.org/pypi/html-linter/) using the `--disable=optional_tag,protocol` options to clean up some of the HTML in #238.

If there's something better out there please comment and let us know! ✨

jrabbit:
https://github.com/motet-a/jinjalint exists but doesn't seem too robust

abandoned-prototype:
Now that we use pre-commit adding an html template linter should be relatively easy and pretty helpful

michplunkett:
It doesn't look like jinjalint is being mainted, but djlint is: https://www.djlint.com/docs/languages/jinja/

We could likely implement this pretty easily.

michplunkett:
This is gonna be split up into two PRs, one formatting and one for linting. Since the changes to the HTML, JS, and CSS will be so vast for just the formatting, no reviewer will actually be able to see the code that is changed for formatting as opposed to linting.

michplunkett:
The HTML formatting part is done, but the linting is not.

PR #966, opened 2023-07-10
Add HTML formatting to `pre-commit`

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/239

## Description of Changes
Added [`djLint`](https://www.djlint.com/docs/languages/jinja/) to the `pre-commit` process and formatted files accordingly.

## Tests and linting

`pre-commit` output:
```console
(openoversight) xxxxxx@MacBook-Air-5 OpenOversight % make lint
pre-commit run --all-files
trim trailing whitespace.................................................Passed
check docstring is first.................................................Passed
check that executables have shebangs.....................................Passed
check json...............................................................Passed
check for case conflicts.................................................Passed
check toml...........................................(no files to check)Skipped
check for merge conflicts................................................Passed
check xml................................................................Passed
check yaml...............................................................Passed
fix end of files.........................................................Passed
check for broken symlinks............................(no files to check)Skipped
mixed line ending........................................................Passed
fix python encoding pragma...............................................Passed
pretty format json.......................................................Passed
fix requirements.txt.....................................................Passed
check blanket noqa.......................................................Passed
check for not-real mock methods..........................................Passed
check for eval().........................................................Passed
use logger.warning(......................................................Passed
Run isort to sort imports................................................Passed
Run pydocstyle...........................................................Passed
Do not use shebangs in non-executable files..............................Passed
flake8...................................................................Passed
black....................................................................Passed
djLint formatting........................................................Passed
(openoversight) xxxxxx@MacBook-Air-5 OpenOversight % 
```

### DIFF
--- a/OpenOversight/tests/routes/test_officer_and_department.py
+++ b/OpenOversight/tests/routes/test_officer_and_department.py
@@ -1674,18 +1674,24 @@ def test_browse_filtering_allows_good(client, mockdata, session):
         )
 
         filter_list = rv.data.decode(ENCODING_UTF_8).split("<dt>Rank</dt>")[1:]
-        assert any(
-            "<dd>{}</dd>".format(job.job_title) in token for token in filter_list
-        )
+        assert any("{}".format(job.job_title) in token for token in filter_list)
+        assert any("<dd>" in token for token in filter_list)
+        assert any("</dd>" in token for token in filter_list)
 
         filter_list = rv.data.decode(ENCODING_UTF_8).split("<dt>Unit</dt>")[1:]
-        assert any("<dd>{}</dd>".format(unit.descrip) in token for token in filter_list)
+        assert any("{}".format(unit.descrip) in token for token in filter_list)
+        assert any("<dd>" in token for token in filter_list)
+        assert any("</dd>" in token for token in filter_list)
 
         filter_list = rv.data.decode(ENCODING_UTF_8).split("<dt>Race</dt>")[1:]
-        assert any("<dd>White</dd>" in token for token in filter_list)
+        assert any("White" in token for token in filter_list)
+        assert any("<dd>" in token for token in filter_list)
+        assert any("</dd>" in token for token in filter_list)
 
         filter_list = rv.data.decode(ENCODING_UTF_8).split("<dt>Gender</dt>")[1:]
-        assert any("<dd>Male</dd>" in token for token in filter_list)
+        assert any("Male" in token for token in filter_list)
+        assert any("<dd>" in token for token in filter_list)
+        assert any("</dd>" in token for token in filter_list)
 
 
 def test_find_officer_redirect(client, mockdata, session):
--- a/OpenOversight/tests/routes/test_officer_and_department.py
+++ b/OpenOversight/tests/routes/test_officer_and_department.py
@@ -3,6 +3,7 @@ import copy
 import csv
 import json
 import random
+import re
 from datetime import date, datetime
 from http import HTTPStatus
 from io import BytesIO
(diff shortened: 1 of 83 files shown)


OpenOversight/tests/routes/test_officer_and_department.py
```diff
@@ -1674,18 +1674,24 @@ def test_browse_filtering_allows_good(client, mockdata, session):
...
-        )
+        assert any("{}".format(job.job_title) in token for token in filter_list)
+        assert any("<dd>" in token for token in filter_list)
+        assert any("</dd>" in token for token in filter_list)
 
         filter_list = rv.data.decode(ENCODING_UTF_8).split("<dt>Unit</dt>")[1:]
-        assert any("<dd>{}</dd>".format(unit.descrip) in token for token in filter_list)
+        assert any("{}".format(unit.descrip) in token for token in filter_list)
+        assert any("<dd>" in token for token in filter_list)
+        assert any("</dd>" in token for token in filter_list)
```
AetherUnbound, line 1684: The changes to these assertions maybe feel like we aren't testing what we previously were (since order no longer matters with the new tests). Is it possible to change the assertions so they test the same thing even if there is whitespace in between?

AetherUnbound (commented):
What a neat find! Being able to lint the template files is a huge win. I have one question about some changed assertions, I'm curious if we could still preserve the check for content order.

OpenOversight/tests/routes/test_officer_and_department.py
michplunkett: I'll give it a shot, yup! It's replicated in the next PR (the image filtering one), so I'll make sure to make whatever change I make on this PR on that one as well.

AetherUnbound: Awesome, thank you! I wonder if we could do something like `combined = "".join([token.strip() for token in filter_list])`, then maybe the assertion can remain the same?

michplunkett: That sounds much simpler than the regex solution I was thinking of trying. I'll give it a shot tomorrow and respond back here.

michplunkett: I tried your solution and my overly simple solution of `filter_list = [token.replace(" ", "").replace("\n", "") for token in filter_list` to no avail.

The problem was that between the `<dd>` elements you had a starting line break and A LOT of spaces. To address that, I created this function and created all of the `filter_list`s with it.
```python
def normalize_tokens_for_comparison(html_str: str, split_str: str):
  """Remove new lines, leading, and closing spaces between <dd> elements in
  formatted HTML."""
  parsed_list = html_str.data.decode(ENCODING_UTF_8).split(split_str)[1:]
  parsed_list = [re.sub(r"<dd>\n\s+", "<dd>", token) for token in parsed_list]
  parsed_list = [re.sub(r"\n\s+</dd>", "</dd>", token) for token in parsed_list]
  return parsed_list
```

I've since added it to both branches.

OpenOversight/tests/routes/test_officer_and_department.py
```diff
@@ -1673,18 +1674,28 @@ def test_browse_filtering_allows_good(client, mockdata, session):
...
-        filter_list = rv.data.decode(ENCODING_UTF_8).split("<dt>Rank</dt>")[1:]
+        def normalize_tokens_for_comparison(html_str: str, split_str: str):
+            """Remove new lines, leading, and closing spaces between <dd> elements in
+            formatted HTML."""
+            parsed_list = html_str.data.decode(ENCODING_UTF_8).split(split_str)[1:]
+            parsed_list = [re.sub(r"<dd>\n\s+", "<dd>", token) for token in parsed_list]
+            parsed_list = [
+                re.sub(r"\n\s+</dd>", "</dd>", token) for token in parsed_list
+            ]
+            return parsed_list
```
michplunkett, line 1685: This addresses the problem that with the newly formatted HTML have a starting line break and A LOT of spaces between the `<dd>` elements.

AetherUnbound (approved):
Fantastic!

Issue #797, opened 2020-08-13
Add states to browse department page or separate department lists by states


ssempervirens:
As of now, our browse department page lists the names of departments, but as we continue to add new departments in new cities, I'm worried this will get confusing. For example, I didn't know that the Chittenden County Sheriff's Office was in Vermont until I googled it. 

The easiest solution here would be to add the state name in parentheses or something after the name — so it would read Chittenden County Sheriff's Office (VT) or Chittenden County Sheriff's Office (Vermont). A perhaps better solution in the longer term would be to have departments organized by federal, and then states, so that each department is listed under a state heading. I'm fine with either but I think it would be useful to add states to the department names.
<img width="757" alt="Screen Shot 2020-08-13 at 4 35 37 PM" src="https://user-images.githubusercontent.com/20425838/90197688-343c2580-dd84-11ea-91ba-d8cb83647c3d.png">

fritzdavenport:
See also (closed) https://github.com/lucyparsons/OpenOversight/issues/792
and https://github.com/lucyparsons/OpenOversight/issues/364

abandoned-prototype:
Bumping this issue since it is still a problem and becomes bigger the more departments we are adding.
In my opinion in the longer term we should think about UI improvements, so people can find the departments they are interested in quicker, some of this is also discussed in #364.

For the short term (covered in this issue), I think we should add a state column to the department and make sure it is populated in all paths that create departments (via UI by administrator and flask command directly on the server).
Then for /browse we should organize the list by state (in alphabetic order) and add a header line with an `id`, so that /browse#NewYork brings you directly to the New York departments for example. Within each state the departments should also be ordered alphabetically 

For other places where we use a dropdown (e.g. /submit, /find) I think using the [optgroup tag](https://www.w3schools.com/TAgs/tag_optgroup.asp) makes sense, not sure how much work it would be to include that (ideally it's only one component we change) everywhere. If this is not feasible we could add the state abbreviation in brackets in front and then order alphabetically (e.g. `[IL] Chicago Police Department`)

michplunkett:
The states have been added to the UI and in the DB. The next step is to meet with a UX person and come up with how to do filtering for the departments.

michplunkett:
Gonna create another ticket for UX things regarding department searching as this ticket already has a bit too much work associated with it.

PR #981, opened 2023-07-21
Add `state` to `departments` table

michplunkett:
## Fixes issue
Partially addresses this: https://github.com/lucyparsons/OpenOversight/issues/797

## Description of Changes
Add `state` column to the `departments` table and update all relevant tests and features. This is the first step to addressing the Department state searching feature.

While taking care of the `.format` in `models/database.py` for `Department`, I took care of the other ones as well.

## Tests and linting

<details><summary>DB Migration Output</summary>

```console
$ flask db upgrade
[2023-07-26 15:52:20,432] INFO in __init__: OpenOversight startup
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 1931b987ce0d -> 18f43ac4622f, add state column to departments
$ flask db downgrade
[2023-07-26 15:52:25,802] INFO in __init__: OpenOversight startup
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running downgrade 18f43ac4622f -> 1931b987ce0d, add state column to departments
$ flask db upgrade
[2023-07-26 15:52:30,441] INFO in __init__: OpenOversight startup
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 1931b987ce0d -> 18f43ac4622f, add state column to departments
$ 
```
</details>


OpenOversight/app/utils/constants.py
```diff
@@ -1,13 +1,71 @@
 import os
 
 
+# Data Constants
+US_STATE_ABBREVIATIONS = [
```
abandoned-prototype, line 5: We have that `us` package that should provide these abbreviations

OpenOversight/tests/conftest.py
```diff
@@ -300,13 +300,20 @@ def add_mockdata(session):
     department = Department(
         name=DEPARTMENT_NAME,
         short_name="SPD",
+        state=random.sample(US_STATE_ABBREVIATIONS, 1)[0],
```
abandoned-prototype, line 303: This should be [`random.choice`](https://docs.python.org/3.11/library/random.html#random.choice). Although, I would prefer just picking one state

OpenOversight/app/utils/constants.py
michplunkett: Done!

OpenOversight/tests/conftest.py
michplunkett: What's the reasoning for just picking one? The way I've understood most testing is that it tends to benefit from random inputs (presuming they are inline with the standards of the column), as you're testing more than just one scenario.

b-meson:
Everything else looks okay except that one question I had.

OpenOversight/tests/routes/test_incidents.py
```diff
@@ -213,8 +215,8 @@ def test_admins_can_edit_incident_links_and_licenses(mockdata, client, session):
         old_links_forms = [
             LinkForm(url=link.url, link_type=link.link_type).data for link in inc.links
         ]
-        new_url = "http://rachel.com"
-        link_form = LinkForm(url="http://rachel.com", link_type="video")
+        new_url = "https://rachel.com"
```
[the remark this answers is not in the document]
michplunkett: No idea, tbh. I could switch it out with a `faker` URL if that'd be better. I just changed it from `http` to `https`.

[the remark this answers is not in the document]
b-meson: Right, I don't know who Rachel is or what her website is lol. I'm generally not a fan of using other people's infrastructure to test our code.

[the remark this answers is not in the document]
michplunkett: Give me 2-3 minutes.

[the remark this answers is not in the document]
michplunkett: Fixed!

OpenOversight/app/models/database.py
```diff
@@ -41,20 +41,22 @@ class Department(BaseModel):
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String(255), index=True, unique=True, nullable=False)
     short_name = db.Column(db.String(100), unique=False, nullable=False)
+    state = db.Column(db.String(2), index=True)
```
abandoned-prototype, line 44: As in my comment on the other PR, we don't need an index on state (The existing index on name also doesn't make much sense to me)

OpenOversight/tests/test_models.py
```diff
@@ -22,8 +22,8 @@
 
 def test_department_repr(mockdata):
     department = Department.query.first()
-    assert department.__repr__() == "<Department ID {}: {}>".format(
-        department.id, department.name
+    assert department.__repr__() == "<Department ID {}: {} {}>".format(
```
abandoned-prototype, line 25: `repr(department)` should work here

abandoned-prototype (approved):

michplunkett:
I need to update documentation and some a couple of tests cases, and I'll merge it tomorrow.

OpenOversight/tests/test_utils.py
```diff
@@ -19,13 +22,24 @@
...
 
 # Utils tests
-
 upload_s3_patch = patch(
     "OpenOversight.app.utils.cloud.upload_obj_to_s3",
     MagicMock(return_value="https://s3-some-bucket/someaddress.jpg"),
 )
 
 
+class PoliceDepartment:
```
abandoned-prototype, line 31: What will this be used for? It seems to be basically the same as the class in database.py -  all the same attributes

OpenOversight/tests/test_utils.py
```diff
@@ -19,13 +22,24 @@
...
+    """Base Police Department class."""
+
+    def __init__(self, name, short_name, state=random.choice(us.STATES).abbr):
```
abandoned-prototype, line 34: This will assign the same state to all departments created with this constructor which might not be intended?
Context: https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

OpenOversight/tests/conftest.py
abandoned-prototype: Problem with random is that it can lead to flaky tests (sometimes passing, sometimes not) which are much worse to fix than a test that always fails

OpenOversight/app/models/database.py
```diff
@@ -39,22 +39,26 @@
 class Department(BaseModel):
     __tablename__ = "departments"
     id = db.Column(db.Integer, primary_key=True)
-    name = db.Column(db.String(255), index=True, unique=True, nullable=False)
+    name = db.Column(db.String(255), index=False, unique=False, nullable=False)
     short_name = db.Column(db.String(100), unique=False, nullable=False)
+    state = db.Column(db.String(2), default="")
```
abandoned-prototype, line 44: Same comment as in the other PR: We should also decide on whether the default is the empty string (and then make the column non-null) or leave null as the default (combining both should be avoided). Either works for me.
Between default and server_default I think server_default should be preferred. So I think it should be either `db.Column(db.String(2))` or `db.Column(db.String(2), server_default="", nullable=False)`. Let me know if you think I am wrong here, definitely a possibility

OpenOversight/tests/test_utils.py
michplunkett: It's used an easier and more standard alternative to code like this:
```python
name = "Added Police Department"
short_name = "APD"
unique_internal_identifier = "30ad0au239eas939asdj"
```

OpenOversight/tests/test_utils.py
michplunkett: Gooooood call! I can make that change.

OpenOversight/app/models/database.py
michplunkett: `server_default` is definitely the right choice here. There are still some modifications that I'm making to this. `default` does not seem to do much here.

OpenOversight/tests/conftest.py
michplunkett: It CAN lead to flakey tests presuming you are not using your random assignments properly or using bad data in your random examples. Regarding states, that won't be the case (it is ALWAYS one of the 50 state abbreviations).  Regarding other values, it's important you choose the correct [provider](https://faker.readthedocs.io/en/master/providers.html). The use of fake data within testing specifically is something used fairly widely and helps us have data within our application that is closer to the real thing.

Should we use an incorrect testing provider, we'll definitely get some errors, but I haven't seen the case thus far.

OpenOversight/app/commands.py
```diff
@@ -565,7 +566,11 @@ def bulk_add_officers(filename, no_create, update_by_name, update_static_fields)
 
 
 @click.command()
-@click.argument("department-name")
+@click.argument("department-name", required=True)
+@click.argument(
+    "department-state",
+    type=click.Choice([state.abbr for state in us.STATES]),
+)
```
michplunkett, line 573: In the next PR, once we can create and edit Departments from the front-end, we can make this required.

OpenOversight/app/models/database.py

OpenOversight/app/models/database.py
```diff
@@ -39,22 +39,26 @@
 class Department(BaseModel):
     __tablename__ = "departments"
     id = db.Column(db.Integer, primary_key=True)
-    name = db.Column(db.String(255), index=True, unique=True, nullable=False)
+    name = db.Column(db.String(255), index=False, unique=False, nullable=False)
     short_name = db.Column(db.String(100), unique=False, nullable=False)
+    state = db.Column(db.String(2), server_default="", nullable=False)
```
michplunkett, line 44: It's been set to `db.Column(db.String(2), server_default="", nullable=False)`, @abandoned-prototype.

Issue #980, opened 2023-07-21
Convert `time` and `date` to `timestamptz` in `incidents` table


michplunkett:
## What issue are you seeing?
We are currently saving the time an incident happened to two separate fields in the `incident` table. They are also timezone agnostic so looking at them doesn't necessarily tell you when, as we cannot glean who put them there or what their timezone was when they input that data.

PR #992, opened 2023-07-27
Add `occurred_at` column

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/980

## Description of Changes
When both the `time` and `date` fields are filled in, we migrate them to the `occurred_at` column which is stored in UTC.

## Screenshots
Before migration:
Screenshot 2023-07-27 at 3 01 41 PM

After initial `upgrade`:
Screenshot 2023-07-27 at 3 20 29 PM

After `downgrade`:
Screenshot 2023-07-27 at 3 36 43 PM

Final `upgrade`:
Screenshot 2023-07-27 at 3 37 02 PM

Incident listing:
Screenshot 2023-07-27 at 3 51 41 PM

## Tests and linting

<details><summary>DB Migration Output</summary>

```console
$ flask db upgrade
[2023-07-27 20:20:06,782] INFO in __init__: OpenOversight startup
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 18f43ac4622f -> 9fa948bcea25, change incident date and time to timestamptz
$ flask db downgrade
[2023-07-27 20:36:20,895] INFO in __init__: OpenOversight startup
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running downgrade 9fa948bcea25 -> 18f43ac4622f, change incident date and time to timestamptz
$ flask db upgrade
[2023-07-27 20:36:49,910] INFO in __init__: OpenOversight startup
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade 18f43ac4622f -> 9fa948bcea25, change incident date and time to timestamptz
$ 
```
</details>


OpenOversight/app/templates/incident_detail.html
```diff
@@ -41,24 +41,26 @@
...
+                {% with detail=True %}
+                  {% include "partials/incident_fields.html" %}
+                {% endwith %}
+              </tbody>
+            </table>
+          </div>
+        </div>
+        <div class="col-sm-12 col-md-6">
+          <h1>Incident Description</h1>
+          {{ incident.description | markdown }}
```
michplunkett, line 62: Before change:
Screenshot 2023-07-27 at 4 03 29 PM

After:
Screenshot 2023-07-27 at 4 03 40 PM

michplunkett:
Just getting through the last three tests and this will be good to go.

OpenOversight/app/models/database.py
```diff
@@ -468,8 +468,7 @@ class Incident(BaseModel):
     __tablename__ = "incidents"
 
     id = db.Column(db.Integer, primary_key=True)
-    date = db.Column(db.Date, unique=False, index=True)
-    time = db.Column(db.Time, unique=False, index=True)
+    occurred_at = db.Column(db.DateTime(timezone=True), unique=False, nullable=False)
```
abandoned-prototype, line 471: This should be a nullable field. Not every incident has a date or time associated with it.
Actually there is a bigger problem, how can we decide between an incident for which no time is known and an incident that happened at midnight. I think that is actually the reason these were two different fields. Sorry, I wasn't really thinking about this issue before. Maybe an additional field of `time_unknown` could help? Or more interestingly, an enum that allows us to set which part of the time we know, e.g. if only the year and month is known, we only display that.
So I think I am fine with merging this, but I will think about following it up with that enum idea

abandoned-prototype (commented):
I am not sure about the timezones here. Can you make that a little more explicit? Ideally we either ask the user for the right timezone, or we use the user's local timezone. Using the server timezone doesn't seem right to me

michplunkett:
Based on my understanding, and it seems to be holding up in testing, it takes in the time with the user's timezone and then saves it in UTC. From there it is rendered back at the user in whatever time they have. Unless I am forgetting something, we should not have to worry about timezones here since they are converted to UTC and then displayed in the user's timezone.

OpenOversight/app/models/database.py
michplunkett: I'd be happy to make that update, if you'd like. It should require too much work and it's not like it's a feature anyone is waiting on.

abandoned-prototype:
> Based on my understanding, and it seems to be holding up in testing, it takes in the time with the user's timezone and then saves it in UTC. From there it is rendered back at the user in whatever time they have. Unless I am forgetting something, we should not have to worry about timezones here since they are converted to UTC and then displayed in the user's timezone.

How did you test it? When I run this locally and then create an incident the time I enter is stored in the DB in UTC (without conversion) and then displayed in local time. So when I say 3:22pm, it stores 3:22pm UTC in the db, and then displays 10:22am in the UI

OpenOversight/app/models/database.py
michplunkett: I'll modify it so that we either have a `date` or `occurred_at` column.

michplunkett:
I think I had a much simpler understanding of this and was thinking of this as if we were handing the backend a UTC timestamp. I am modifying this PR a good amount. I'll reach out when it's good.

michplunkett:
I'm gonna knock this one out on Monday when my mind is a little clearer.

michplunkett:
I'm gonna restart this feature in the following PR: https://github.com/lucyparsons/OpenOversight/pull/1024

PR #995, opened 2023-07-28
Sort `Department`s and add stats and cacheing

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/797

## Description of Changes
Added sorting by `state` and `name` columns for department listing , added baseline `Department` statistics for each department, and added cacheing for those `Department` statistics.

Before change:
Screenshot 2023-07-28 at 3 24 50 PM

After change:
Screenshot 2023-07-28 at 3 27 03 PM

## Tests and linting

### DIFF
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -61,6 +61,24 @@ class Department(BaseModel):
             "unique_internal_identifier_label": self.unique_internal_identifier_label,
         }
 
+    def total_documented_officers(self):
+        return (
+            db.session.query(Officer).filter(Officer.department_id == self.id).count()
+        )
+
+    def total_documented_assignments(self):
+        return (
+            db.session.query(Assignment.id)
+            .join(Officer, Assignment.officer_id == Officer.id)
+            .filter(Officer.department_id == self.id)
+            .count()
+        )
+
+    def total_documented_incidents(self):
+        return (
+            db.session.query(Incident).filter(Incident.department_id == self.id).count()
+        )
+
 
 class Job(BaseModel):
     __tablename__ = "jobs"
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -3,6 +3,8 @@ import time
 from datetime import date
 
 from authlib.jose import JoseError, JsonWebToken
+from cachetools import TTLCache, cached
+from cachetools.keys import hashkey
 from flask import current_app
 from flask_login import UserMixin
 from flask_sqlalchemy import SQLAlchemy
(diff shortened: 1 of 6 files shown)


OpenOversight/app/templates/browse.html
```diff
@@ -14,7 +14,7 @@ <h1>
       </h1>
     </div>
     <div class="text-center frontpage-leads">
-      {% for department in departments %}
+      {% for department in departments | sort(attribute="state,name") %}
```
michplunkett, line 17: Sort.

OpenOversight/app/templates/browse.html
```diff
@@ -26,6 +26,13 @@ <h2>
                 </a>
               {% endif %}
             </h2>
+            <div>
+              <b class="dept-{{ department.id }}">Officers Documented:</b> {{ department.total_documented_officers() }}
+              <br>
+              <b class="dept-{{ department.id }}">Assignments Documented:</b> {{ department.total_documented_assignments() }}
+              <br>
+              <b class="dept-{{ department.id }}">Incidents Documented:</b> {{ department.total_documented_incidents() }}
+            </div>
```
michplunkett, line 35: Informational blurbs.

OpenOversight/app/templates/browse.html
abandoned-prototype: Great! I would prefer to do the sorting in the view and not the template (I was hoping to at some point also move the existing logic out of the templates)

OpenOversight/app/models/database.py
```diff
@@ -61,6 +61,24 @@ def to_custom_dict(self):
...
+            .join(Officer, Assignment.officer_id == Officer.id)
+            .filter(Officer.department_id == self.id)
+            .count()
+        )
+
+    def total_documented_incidents(self):
+        return (
+            db.session.query(Incident).filter(Incident.department_id == self.id).count()
+        )
+
```
abandoned-prototype, line 81: Nice addition! Only concern I have is performance, since for `/browse` we have to do this for every department, so `3 * num departments` queries. Maybe a good time to bring in the query caching you had told me about. How could we enable something like that?

abandoned-prototype (approved):

OpenOversight/app/templates/browse.html
michplunkett: I've generally been told that it's a better idea to do the sorting in the HTML, presuming it only matters there and the amount of information isn't HUGE, as it's only a client concern. If it was something we needed on the server side, I could see doing it there, but I think it's fine to do it here.

All that being said, I don't have any qualms doing it one way or the other. Thumbs up this message and I can take care of it on the server side.

OpenOversight/app/models/database.py
michplunkett: Most of this came from inspiration I saw based on @AetherUnbound's work:
https://github.com/OrcaCollective/OpenOversight/blob/main/OpenOversight/app/models.py#L88C1-L95

I don't see why we couldn't implement a fairly straight-forward cache here similar to what they accomplished on their branch.

michplunkett: Most of this came from inspiration I saw based on @AetherUnbound's work:
https://github.com/OrcaCollective/OpenOversight/blob/b8bf6394d8f357bd23f0d03adc2b1b573cf5d329/OpenOversight/app/models.py#L88C1-L95

I don't see why we couldn't implement a fairly straight-forward cache here similar to what they accomplished on their branch.

michplunkett: https://github.com/lucyparsons/OpenOversight/issues/997

OpenOversight/app/templates/browse.html
michplunkett: Done!

OpenOversight/app/models/database.py

PR #1001, opened 2023-07-31
Delete `setup.py`

michplunkett:
## Description of Changes
I looked around the application and we don't actually use the file at all, so I wiped it.

## Tests and linting


setup.py
```diff
@@ -3,16 +3,16 @@
...
     classifiers=(
         "Natural Language :: English",
-        "Programming Language :: Python :: 3.5",
-        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: 3 :: Only",
     ),
-    python_requires=">=3.5",
+    python_requires=">=3.11",
     install_requires=["flask", "werkzeug", "Flask-WTF", "psycopg2", "sqlalchemy"],
```
michplunkett, line 17: These are VERY much out of sync, but I'm a little unsure of how to update them other than just reading from the `requirements.txt` file.

michplunkett: Since it's so out of date and we only use the `requirements.txt` file, I'm tempted to delete the `install_requires` line: https://github.com/lucyparsons/OpenOversight/blob/9bef4baedb08365df53c603340a7547c711b5b1d/DEPLOY.md?plain=1#L5-L9

setup.py
```diff
@@ -3,16 +3,15 @@
...
     author_email="jen@redshiftzero.com",
     classifiers=(
         "Natural Language :: English",
-        "Programming Language :: Python :: 3.5",
-        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: 3 :: Only",
     ),
-    python_requires=">=3.5",
-    install_requires=["flask", "werkzeug", "Flask-WTF", "psycopg2", "sqlalchemy"],
```
michplunkett, line 17: These are VERY much out of sync and we do not reference using the `setup.py` requirements anywhere in our repository. To prevent any confusion, I am going to delete the line.

https://github.com/lucyparsons/OpenOversight/blob/9bef4baedb08365df53c603340a7547c711b5b1d/DEPLOY.md?plain=1#L5-L9

null-space-0 (approved):
Yeah if it's not called let's just get rid of it, it's just making things messy

dockerfiles/web/Dockerfile-prod
```diff
@@ -43,8 +43,6 @@ RUN pip3 install -r requirements.txt
 # Packages to install with yarn.
 COPY package.json /usr/src/app/
 RUN yarn
-# Copy python script to initialize database.
-COPY /usr/src/app/
```
michplunkett, line 47: This deletion addresses the issue in this build: https://github.com/lucyparsons/OpenOversight/actions/runs/5732169896/job/15534595950

PR #1006, opened 2023-08-03
Fix adding duplicate officers to incidents

sea-kelp:
## Description of Changes
Merging fix https://github.com/OrcaCollective/OpenOversight/pull/347 upstream.

**Original PR description**:
Only add new officers to incidents.

When a user edits an incident currently, we retrieve the Incident from the database and append all officers in the form to that Incident. This led to sqlalchemy trying to insert duplicate rows into the officer_incidents association table.

As far as I can tell, this was silently ignored by sqlalchemy in the past and no unexpected rows were created. With the upgrade to sqlalchemy 2.0.19 (https://github.com/OrcaCollective/OpenOversight/pull/344), the library appears to have started raising exceptions related to this behavior.

This change checks to see whether an officer already exists in Incident.officers before adding them so as to not add duplicate officers.

## Notes for Deployment
N/A

N/A

## Tests and linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -1626,7 +1626,7 @@ class IncidentApi(ModelView):
                     except ValueError:
                         our_id = officer["oo_id"].split('value="')[1][:-2]
                         of = Officer.query.filter_by(id=int(our_id)).first()
-                    if of:
+                    if of and of not in obj.officers:
                         obj.officers.append(of)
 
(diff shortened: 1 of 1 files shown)


michplunkett:
Looks good to me! If possible, it'd be nice to have a migration that removes any duplicates, though I'd also be happy to write that after this is merged.

michplunkett (approved):

PR #1024, opened 2023-08-16
Add `occurred_at` column to `incidents`

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/980

## Description of Changes
Add `occurred_at` column to `incidents` table so we can have UTC times for incidents, should we have all the needed information.

## Tests and linting


OpenOversight/migrations/versions/2023-08-16-2126_729ab5beaa6c_add_occurred_at_column.py
```diff
@@ -0,0 +1,78 @@
+"""add occurred_at column
```
michplunkett, line 1:
```bash
/usr/src/app$ flask db downgrade
...
INFO  [alembic.runtime.migration] Running downgrade 729ab5beaa6c -> b38c133bed3c, add occurred_at column
I have no name!@4219f96b41a8:/usr/src/app$ flask db upgrade
...
INFO  [alembic.runtime.migration] Running upgrade b38c133bed3c -> 729ab5beaa6c, add occurred_at column
/usr/src/app$ 
```

OpenOversight/tests/test_models.py
```diff
@@ -79,19 +80,45 @@ def test_department_total_documented_incidents(mockdata):
...
 
 
-def test_officer_repr(mockdata):
-    officer = Officer.query.first()
-    if officer.unique_internal_identifier:
-        assert (
-            repr(officer) == f"<Officer ID {officer.id}: {officer.first_name} "
-            f"{officer.middle_initial} {officer.last_name} {officer.suffix} "
-            f"({officer.unique_internal_identifier})>"
+def test_officer_repr(session):
```
michplunkett, line 83: Added additional test coverage for Officers.

OpenOversight/app/models/database.py
```diff
@@ -312,13 +312,10 @@ def currently_on_force(self):
...
-                f"<Officer ID {self.id}: {self.first_name} {self.middle_initial} "
-                + f"{self.last_name} {self.suffix} ({self.unique_internal_identifier})>"
+                f"<Officer ID {self.id}: {self.full_name()} "
+                f"({self.unique_internal_identifier})>"
             )
-        return (
-            f"<Officer ID {self.id}: {self.first_name} {self.middle_initial} "
-            + f"{self.last_name} {self.suffix}>"
-        )
+        return f"<Officer ID {self.id}: {self.full_name()}>"
```
michplunkett, line 318: These changes will be masked by: https://github.com/lucyparsons/OpenOversight/pull/1037

OpenOversight/app/templates/partials/officer_incidents.html
```diff
@@ -2,7 +2,7 @@ <h3>Incidents</h3>
 {% if officer.incidents %}
   <table class="table table-hover table-responsive">
     <tbody>
-      {% for incident in officer.incidents | sort(attribute='date') | reverse %}
+      {% for incident in officer.incidents %}
```
michplunkett, line 5: It is redundant with the controller sort.

michplunkett:
**Note:** I am still manually testing this PR, but I THINK it's just about done. Let me know if y'all see any issue, etc.

OpenOversight/app/templates/partials/incident_fields.html
```diff
@@ -1,15 +1,33 @@
...
+      <strong>Date</strong>
+    </td>
+    <td>{{ incident.occurred_at | local_date }}</td>
+  </tr>
   <tr>
     <td>
       <strong>Time</strong>
     </td>
-    <td>{{ incident.time | display_time }}</td>
+    <td>{{ incident.occurred_at | local_time }}</td>
```
sea-kelp, line 30: If I'm understanding this correctly, we are localizing the datetime to the user's own timezone here, right? I think it would make more sense to display the datetime in the department's timezone instead.

For example, if I were to look up an NYPD incident, I would want to know that the incident occurred around noon in NY, rather than 9am in Seattle.

michplunkett: They translate accordingly and I think there's value in having them be discrete points in time, as opposed to them being relative points. Having them be discrete points in time also makes it so that its worth entering times for data collection things downstream from just this page.

Issue #1008, opened 2023-08-03
Finish auditing fields for non-`users` tables


michplunkett:
## What issue are you seeing?
The following tables need `last_updated_by` and `last_updated_at` columns added to them and to have their logic updated accordingly: `assignments`, `departments`, `descriptions`*, `faces`, `jobs`, `license_plates`, `links`, `locations`, `notes`*,  `officers`, `raw_images`, `salaries`, `unit_types`.

This all presumes these rows can be edited, which I believe is the case. If they cannot, then it is fine to not give them those two fields. There was talk of updating the `BaseModel` to add these, but I'm not sure if that would take too much time. Either solution would be great.

* Has `updated_at`, but the name should be changed and the values should be carried over

michplunkett:
@sea-kelp is gonna take care of this one! 🥳 🥳 🥳 🥳

PR #1032, opened 2023-08-21
Add create and last_update columns to models

sea-kelp:
## Fixes issue
#1008 

## Description of Changes
Add creation and last update columns to track when and by whom models were updated.

## Notes for Deployment
Run alembic migration

N/A

## Tests and linting

### DIFF
--- a/OpenOversight/app/main/model_view.py
+++ b/OpenOversight/app/main/model_view.py
@@ -80,8 +80,13 @@ class ModelView(MethodView):
 
         if form.validate_on_submit():
             new_obj = self.create_function(form, current_user)
+            if hasattr(new_obj, "created_by"):
+                new_obj.created_by = current_user.get_id()
+            if hasattr(new_obj, "last_updated_by"):
+                new_obj.last_updated_by = current_user.get_id()
             db.session.add(new_obj)
             db.session.commit()
+
             match self.model.__name__:
                 case Incident.__name__:
                     Department(id=new_obj.department_id).remove_database_cache_entries(
@@ -117,31 +122,6 @@ class ModelView(MethodView):
 
         if not form:
             form = self.get_edit_form(obj)
-            now = datetime.datetime.now()
-            # if the object doesn't have a creator id set it to current user
-            if (
-                hasattr(obj, "created_by")
-                and hasattr(form, "created_by")
-                and getattr(obj, "created_by")
-            ):
-                form.created_by.data = obj.created_by
-            elif hasattr(form, "created_by"):
-                form.created_by.data = current_user.get_id()
-
-            # TODO: Add created_at field to forms
-            if (
-                hasattr(obj, "created_at")
-                and hasattr(form, "created_at")
-                and getattr(obj, "created_at")
-            ):
-                form.created_at.data = obj.created_at
-            elif hasattr(form, "created_at"):
-                form.created_at.data = now
-
-            # if the object keeps track of who updated it last, set to current user
-            if hasattr(obj, "last_updated_by") and hasattr(form, "last_updated_by"):
-                form.last_updated_by.data = current_user.get_id()
-                form.last_updated_at.data = now
 
         if hasattr(form, "department"):
             add_department_query(form, current_user)
@@ -236,8 +216,15 @@ class ModelView(MethodView):
 
     def populate_obj(self, form, obj):
         form.populate_obj(obj)
+
+        # if the object doesn't have a creator id set it to current user
+        if hasattr(obj, "created_by") and not getattr(obj, "created_by"):
+            obj.created_by = current_user.get_id()
+        # if the object keeps track of who updated it last, set to current user
         if hasattr(obj, "last_updated_at"):
             obj.last_updated_at = datetime.datetime.now()
+            obj.last_updated_by = current_user.get_id()
+
         db.session.add(obj)
         db.session.commit()
 
--- a/OpenOversight/app/main/model_view.py
+++ b/OpenOversight/app/main/model_view.py
@@ -117,6 +117,7 @@ class ModelView(MethodView):
 
         if not form:
             form = self.get_edit_form(obj)
+            now = datetime.datetime.now()
             # if the object doesn't have a creator id set it to current user
             if (
                 hasattr(obj, "created_by")
@@ -127,10 +128,20 @@ class ModelView(MethodView):
             elif hasattr(form, "created_by"):
                 form.created_by.data = current_user.get_id()
 
+            # TODO: Add created_at field to forms
+            if (
+                hasattr(obj, "created_at")
+                and hasattr(form, "created_at")
+                and getattr(obj, "created_at")
+            ):
+                form.created_at.data = obj.created_at
+            elif hasattr(form, "created_at"):
+                form.created_at.data = now
+
             # if the object keeps track of who updated it last, set to current user
             if hasattr(obj, "last_updated_by") and hasattr(form, "last_updated_by"):
                 form.last_updated_by.data = current_user.get_id()
-                form.last_updated_at.data = datetime.datetime.now()
+                form.last_updated_at.data = now
 
         if hasattr(form, "department"):
             add_department_query(form, current_user)
@@ -225,8 +236,8 @@ class ModelView(MethodView):
 
     def populate_obj(self, form, obj):
         form.populate_obj(obj)
-        if hasattr(obj, "updated_at"):
-            obj.updated_at = datetime.datetime.now()
+        if hasattr(obj, "last_updated_at"):
+            obj.last_updated_at = datetime.datetime.now()
         db.session.add(obj)
         db.session.commit()
 
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -9,8 +9,7 @@ from flask import current_app
 from flask_login import UserMixin
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import CheckConstraint, UniqueConstraint, func
-from sqlalchemy.ext.declarative import DeclarativeMeta
-from sqlalchemy.orm import validates
+from sqlalchemy.orm import DeclarativeMeta, declarative_mixin, declared_attr, validates
 from sqlalchemy.sql import func as sql_func
 from werkzeug.security import check_password_hash, generate_password_hash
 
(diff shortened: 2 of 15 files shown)


OpenOversight/app/main/model_view.py
```diff
@@ -77,11 +77,16 @@ def new(self, form=None):
...
+            now = datetime.datetime.now()
             if hasattr(form, "created_by") and not form.created_by.data:
                 form.created_by.data = current_user.get_id()
+            if hasattr(form, "created_at") and not form.created_at.data:
+                # TODO: Add created_at field to forms
+                form.created_at.data = now
             if hasattr(form, "last_updated_by"):
                 form.last_updated_by.data = current_user.get_id()
-                form.last_updated_at.data = datetime.datetime.now()
+                form.last_updated_at.data = now
```
sea-kelp, line 89: Should `created_by` and `last_updated_by` be implemented as hidden form fields? Users are able to arbitrarily modify the fields using devtools

michplunkett: They shouldn't be, nope! I'm removing those in a PR now.

OpenOversight/app/main/model_view.py
```diff
@@ -230,8 +246,8 @@ def get_department_id(self, obj):
 
     def populate_obj(self, form, obj):
         form.populate_obj(obj)
-        if hasattr(obj, "updated_at"):
-            obj.updated_at = datetime.datetime.now()
+        if hasattr(obj, "last_updated_at"):
+            obj.last_updated_at = datetime.datetime.now()
```
michplunkett, line 250: Noice.

OpenOversight/migrations/versions/2023-08-21-0300_a35aa1a114fa_add_create_and_last_update_columns.py
```diff
@@ -0,0 +1,289 @@
...
+            ["id"],
+            ondelete="SET NULL",
+        )
+
+    with op.batch_alter_table("notes", schema=None) as batch_op:
+        batch_op.alter_column("updated_at", new_column_name="last_updated_at")
+        # batch_op.add_column(sa.Column('last_updated_at', sa.DateTime(timezone=True), nullable=True))
+        batch_op.add_column(sa.Column("last_updated_by", sa.Integer(), nullable=True))
+        # batch_op.create_foreign_key(None, 'users', ['last_updated_by'], ['id'], ondelete='SET NULL')
+        # batch_op.drop_column('updated_at')
```
michplunkett, line 132: Can you remove the commented alembic lines?

OpenOversight/tests/routes/test_notes.py
```diff
@@ -280,7 +280,7 @@ def test_acs_can_delete_their_notes_in_their_department(mockdata, client, sessio
             officer_id=officer.id,
             created_by=user.id,
             created_at=datetime.now(),
-            updated_at=datetime.now(),
+            last_updated_at=datetime.now(),
```
michplunkett, line 283: It's smöl, but it may be worth it to create a variable for `datetime.now()` so we aren't calling it twice.

OpenOversight/app/models/database.py
```diff
@@ -63,22 +62,48 @@
...
+
+    @declared_attr
+    def last_updated_by(cls):
+        return db.Column(
+            db.Integer, db.ForeignKey("users.id", ondelete="SET NULL"), unique=False
+        )
+
+    @declared_attr
+    def creator(cls):
+        return db.relationship("User", foreign_keys=[cls.created_by])
```
michplunkett, line 97: Fresssssh!

OpenOversight/app/models/database.py
```diff
@@ -63,22 +62,48 @@
...
         nullable=False,
         server_default=sql_func.now(),
         unique=False,
     )
-    created_by = db.Column(
-        db.Integer, db.ForeignKey("users.id", ondelete="SET NULL"), unique=False
+    last_updated_at = db.Column(
+        db.DateTime(timezone=True),
+        nullable=True,
+        unique=False,
```
michplunkett, line 80: Since we are always adding a value to `last_updated_at` when we create it, we could add make it have the same setup as `created_at` (non-nullable, defaults to `sql_func.now()`).

OpenOversight/app/models/database.py
```diff
@@ -710,8 +584,27 @@ class User(UserMixin, BaseModel):
     is_disabled = db.Column(db.Boolean, default=False)
     dept_pref = db.Column(db.Integer, db.ForeignKey("departments.id"))
     dept_pref_rel = db.relationship("Department", foreign_keys=[dept_pref])
-    classifications = db.relationship("Image", back_populates="user")
-    tags = db.relationship("Face", backref="user")
+
+    # creator backlinks
+    classifications = db.relationship(
+        "Image", back_populates="creator", foreign_keys="[Image.created_by]"
```
michplunkett, line 590: 🔬:
- You could make the `creator` string a consant.
- Are the brackets needed syntax?

michplunkett (changes requested):
Some questions are asks, but at the very least I'd like to see `last_updated_at` be defaulted the same as `created_by` and the comments removed from the alembic file. Overall though, this is fantastic!

OpenOversight/app/models/database.py
sea-kelp: Updated!

OpenOversight/app/models/database.py
sea-kelp: Added `creator` as a constant and removed the brackets

OpenOversight/migrations/versions/2023-08-21-0300_a35aa1a114fa_add_create_and_last_update_columns.py
sea-kelp: Removed commented lines!

OpenOversight/app/utils/forms.py
```diff
@@ -101,7 +101,7 @@ def add_officer_profile(form: AddOfficerForm, current_user: User):
                     user_id=current_user.get_id(),
                     officer=officer,
                     created_at=datetime.datetime.now(),
```
michplunkett, line 103: The `last_updated_by` field should be added to these models.

Nit: You could delete the `created_at` fields since they auto-populate. The same would be fine for the `last_updated_at` field.

OpenOversight/app/models/database.py
```diff
@@ -63,22 +63,49 @@
...
-    created_by = db.Column(
-        db.Integer, db.ForeignKey("users.id", ondelete="SET NULL"), unique=False
+    last_updated_at = db.Column(
+        db.DateTime(timezone=True),
+        nullable=False,
+        server_default=sql_func.now(),
+        unique=False,
     )
 
+    @declared_attr
```
michplunkett, line 85: Do we need this to be under the `@declared_attr`? Couldn't this just be added as a regular column like it was in the `Job` model?

OpenOversight/tests/routes/test_descriptions.py
```diff
@@ -124,7 +124,8 @@ def test_admins_can_edit_descriptions(mockdata, client, session):
         description = Description(
             text_contents=old_description,
             officer_id=officer.id,
-            updated_at=original_date,
+            created_at=original_date,
+            last_updated_at=original_date,
```
michplunkett, line 128: I don't think we need the `last_updated_at` spot filled in unless it is an editing action, but we do need the `last_updated_by` column filled in. `created_at` doesn't seem necessary in general unless you'd like to use it in a assertion statement within the test, imo.

OpenOversight/app/main/model_view.py
```diff
@@ -127,10 +128,20 @@ def edit(self, obj_id, form=None):
...
 
+            # TODO: Add created_at field to forms
+            if (
+                hasattr(obj, "created_at")
+                and hasattr(form, "created_at")
+                and getattr(obj, "created_at")
+            ):
+                form.created_at.data = obj.created_at
+            elif hasattr(form, "created_at"):
+                form.created_at.data = now
```
michplunkett, line 139: The last merged PR removed the `created_by` fields from the forms, and I think it'd make sense if we didn't have the `created_at` fields on there either. The field should be staying the same in the DB as long as we don't add update the `created_at` field, yeah?

OpenOversight/app/models/database.py
michplunkett: Ahhhh, that makes sense! I appreciate the rationale/reasoning.

OpenOversight/migrations/versions/2023-08-21-0300_a35aa1a114fa_add_create_and_last_update_columns.py
```diff
@@ -0,0 +1,332 @@
...
+            sa.Column(
+                "last_updated_at",
+                sa.DateTime(timezone=True),
+                server_default=sa.text("now()"),
+                nullable=True,
+            )
+        )
+        batch_op.add_column(sa.Column("last_updated_by", sa.Integer(), nullable=True))
+        batch_op.create_foreign_key(
+            "fk_assignments_last_updated_by_users",
```
michplunkett, line 32: LAST THING! We have a lot of scenarios where there weren't names used for FKs and they are being input as `[table]_[column_name]_fkey`. Could you update the foreign key names to that motif?
Screenshot 2023-08-22 at 3 31 50 PM

michplunkett: I'm addressing all of the previous migrations in this PR: https://github.com/lucyparsons/OpenOversight/pull/1036

OpenOversight/app/main/model_view.py
```diff
@@ -117,20 +122,6 @@ def edit(self, obj_id, form=None):
...
-                and getattr(obj, "created_by")
-            ):
-                form.created_by.data = obj.created_by
-            elif hasattr(form, "created_by"):
-                form.created_by.data = current_user.get_id()
-
-            # if the object keeps track of who updated it last, set to current user
-            if hasattr(obj, "last_updated_by") and hasattr(form, "last_updated_by"):
-                form.last_updated_by.data = current_user.get_id()
-                form.last_updated_at.data = datetime.datetime.now()
```
michplunkett, line 133: I think it would make sense to keep this here. That way we are updating the two columns for every edit.

```python
if hasattr(obj, "last_updated_by"):
    form.last_updated_by.data = current_user.get_id()
    form.last_updated_at.data = datetime.datetime.now()
```

OpenOversight/app/main/model_view.py
```diff
@@ -225,8 +216,15 @@ def get_department_id(self, obj):
...
-        if hasattr(obj, "updated_at"):
-            obj.updated_at = datetime.datetime.now()
+
+        # if the object doesn't have a creator id set it to current user
+        if hasattr(obj, "created_by") and not getattr(obj, "created_by"):
+            obj.created_by = current_user.get_id()
+        # if the object keeps track of who updated it last, set to current user
+        if hasattr(obj, "last_updated_at"):
+            obj.last_updated_at = datetime.datetime.now()
+            obj.last_updated_by = current_user.get_id()
```
michplunkett, line 226: Does this trigger on edits or just new objects?

sea-kelp: This code only triggers on new objects but I've also updated `populate_obj` which updates the last_updated on edits

OpenOversight/app/main/model_view.py
```diff
@@ -80,8 +80,13 @@ def new(self, form=None):
 
         if form.validate_on_submit():
             new_obj = self.create_function(form, current_user)
+            if hasattr(new_obj, "created_by"):
+                new_obj.created_by = current_user.get_id()
+            if hasattr(new_obj, "last_updated_by"):
+                new_obj.last_updated_by = current_user.get_id()
```
sea-kelp, line 86: The created_by and last_updated_by for new objects are updated here

OpenOversight/app/main/model_view.py
sea-kelp: I've replaced this with the change in `populate_obj` which will update the `last_updated_by` and `last_updated_at` when editing objects

OpenOversight/migrations/versions/2023-08-21-0300_a35aa1a114fa_add_create_and_last_update_columns.py
sea-kelp: Good callout! I've updated the foreign key names

michplunkett (approved):

Issue #1023, opened 2023-08-16
Address SQLAlchemy warning


michplunkett:
## What issue are you seeing?
SQLAlchemy is sending a warning every time we create the DB which may create a problem in the future if we don't address it soon.

```zsh
/usr/src/app/OpenOversight/app/../migrations/env.py:75: SAWarning: Cannot correctly sort tables; there are unresolvable cycles between tables "departments, users", which is usually caused by mutually dependent foreign key constraints.  Foreign key constraints involving these tables will not be considered; this warning may raise an error in a future release.
  context.run_migrations()
```

michplunkett:
Useful links:
- https://stackoverflow.com/a/63054555
- https://stackoverflow.com/a/72677284

PR #1036, opened 2023-08-22
Address sql alchemy warning

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/1023

## Description of Changes
Address the warnings that pop up when populating rows in the DB by giving the foreign key constraints names. The warning no longer shows up when running `make dev`. I named the foreign keys according to the motif that they were given by default:
Screenshot 2023-08-22 at 3 31 50 PM

```zsh
/usr/src/app/OpenOversight/app/../migrations/env.py:75: SAWarning: Cannot correctly sort tables; there are unresolvable cycles between tables "departments, users", which is usually caused by mutually dependent foreign key constraints.  Foreign key constraints involving these tables will not be considered; this warning may raise an error in a future release.
  context.run_migrations()
```

## Tests and linting
 - [x] Ran `make dev` to create tables and populate data to validate changes to migrations.


b-meson (approved):
looks great. Thanks Michael.

PR #1048, opened 2023-08-30
Address edit `Department` error

michplunkett:
## Description of Changes
I removed the created_by field from the from on editing a `Department`.

## Tests and linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -1,7 +1,7 @@
-import datetime
 import os
 import re
 import sys
+from datetime import datetime
 from http import HTTPMethod, HTTPStatus
 from traceback import format_exc
 
@@ -675,7 +675,6 @@ def redirect_add_department():
 @admin_required
 def add_department():
     form = DepartmentForm()
-    form.created_by = current_user.get_id()
     if form.validate_on_submit():
         department_does_not_exist = (
             Department.query.filter_by(
@@ -690,6 +689,7 @@ def add_department():
                 short_name=form.short_name.data,
                 state=form.state.data,
                 created_by=current_user.get_id(),
+                last_updated_by=current_user.get_id(),
             )
             db.session.add(department)
             db.session.flush()
@@ -769,6 +769,8 @@ def edit_department(department_id: int):
         department.name = form.name.data
         department.short_name = form.short_name.data
         department.state = form.state.data
+        department.last_updated_by = current_user.get_id()
+        department.last_updated_at = datetime.now()
         db.session.flush()
         if form.jobs.data:
             new_ranks = []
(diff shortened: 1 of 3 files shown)


OpenOversight/app/main/views.py
```diff
@@ -675,7 +675,6 @@ def redirect_add_department():
 @admin_required
 def add_department():
     form = DepartmentForm()
-    form.created_by = current_user.get_id()
```
michplunkett, line 678: This was in `add_department`, but it certainly doesn't belong there as the field no longer exists.

Screenshot 2023-08-30 at 1 31 33 PM

OpenOversight/tests/routes/test_officer_and_department.py
```diff
@@ -526,6 +526,8 @@ def test_admin_can_add_police_department(mockdata, client, session):
         department = Department.query.filter_by(name=TestPD.name).one()
         assert department.short_name == TestPD.short_name
         assert department.state == TestPD.state
+        assert department.created_by == user.id
+        assert department.last_updated_by == user.id
```
michplunkett, line 530: Assert both fields are populated upon creation.

OpenOversight/tests/routes/test_officer_and_department.py
```diff
@@ -692,6 +695,8 @@ def test_admin_can_edit_police_department(mockdata, client, session):
         edit_state_department = Department.query.filter_by(name=CorrectedPD.name).one()
         assert edit_state_department.short_name == CorrectedPD.short_name
         assert edit_state_department.state == CorrectedPD.state
+        assert edit_state_department.last_updated_by == user.id
+        assert edit_state_department.last_updated_at > edit_state_department.created_at
```
michplunkett, line 699: Assert that the last_updated_at value is greater than the created_at value.

OpenOversight/app/main/views.py
```diff
@@ -690,6 +689,7 @@ def add_department():
                 short_name=form.short_name.data,
                 state=form.state.data,
                 created_by=current_user.get_id(),
+                last_updated_by=current_user.get_id(),
```
michplunkett, line 692: Since departments are not going through the Mixin-derived creation method, I made sure to add both fields.

OpenOversight/app/main/views.py
```diff
@@ -769,6 +769,8 @@ def edit_department(department_id: int):
         department.name = form.name.data
         department.short_name = form.short_name.data
         department.state = form.state.data
+        department.last_updated_by = int(current_user.get_id())
+        department.last_updated_at = datetime.now()
```
michplunkett, line 773: Update values upon edit.

OpenOversight/app/main/views.py
```diff
@@ -769,6 +769,8 @@ def edit_department(department_id: int):
         department.name = form.name.data
         department.short_name = form.short_name.data
         department.state = form.state.data
+        department.last_updated_by = int(current_user.get_id())
```
michplunkett, line 772: During testing, this kept being returned as a string, so I had to cast it as an int. My hope is that it's at worst redundant.

sea-kelp: This is really strange! Would we want to dig into why that's happening?

sea-kelp:
Would you mind adding a test for this? I would have expected one of our tests to have picked this up.

michplunkett:
Which error specifically?

OpenOversight/app/main/views.py
sea-kelp: I just took a look and it looks like this is because we are using `current_user.get_id()`, which is a Flask-Login function which is required to return a string. You should be able to use `current_user.id` to get an int id instead

https://flask-login.readthedocs.io/en/latest/#your-user-class

michplunkett:
Can you recheck this PR when you get a chance, @sea-kelp?

OpenOversight/app/main/views.py
michplunkett: Excuse me, I didn't see this question. I'll make the update now.

michplunkett: Fixed!

sea-kelp (approved):
Looks good!

PR #1056, opened 2023-09-10
Syntax changes and model update changes

michplunkett:
## Description of Changes
A lot of small syntax changes along with adding missing field inclusions.

## Tests and linting

### DIFF
--- a/OpenOversight/app/models/database_imports.py
+++ b/OpenOversight/app/models/database_imports.py
@@ -74,6 +74,7 @@ def parse_str(value: Optional[str], default: Optional[str] = "") -> Optional[str
 
 def create_officer_from_dict(data: Dict[str, Any], force_id: bool = False) -> Officer:
     admin_user = User.query.filter_by(is_administrator=True).first()
+
     officer = Officer(
         department_id=int(data["department_id"]),
         last_name=parse_str(data.get("last_name", "")),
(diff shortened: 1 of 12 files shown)


OpenOversight/app/templates/partials/officer_incidents.html
```diff
@@ -2,7 +2,7 @@ <h3>Incidents</h3>
 {% if officer.incidents %}
   <table class="table table-hover table-responsive">
     <tbody>
-      {% for incident in officer.incidents | sort(attribute='date') | reverse %}
+      {% for incident in officer.incidents %}
```
michplunkett, line 5: This is now covered in the view.

OpenOversight/app/models/database_imports.py
```diff
@@ -223,12 +240,12 @@ def update_link_from_dict(data: Dict[str, Any], link: Link) -> Link:
         link.description = parse_str(data.get("description"), None)
     if "author" in data:
         link.author = parse_str(data.get("author"), None)
-    if "created_by" in data:
-        link.created_by = parse_int(data.get("created_by"))
```
michplunkett, line 227: The `created_by` field shouldn't be updated.

sea-kelp (approved):
Should this also be replaced with a constant?
https://github.com/lucyparsons/OpenOversight/pull/1056/files#diff-0ea02af7be8529b5f1eb21efd1af13f1d633060c50ee2529a1064ddba1616c93L76-L80

michplunkett:
@sea-kelp, at the very least it should be an f-string. We can change that on the next code change.

PR #1062, opened 2023-09-19
Use `onupdate` for `last_updated_at`

### LINKED ISSUE #1061: Use `onupdate` for `last_updated_at` fields
## What issue are you seeing?
Instead of manually updating the `last_updated_at` values, we should be using the `onupdate=func.now()` syntax.

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/1061

## Description of Changes
Instead of manually updating `last_updated_at`, we should let SQL Alchemy take care of it.

## Tests and linting

### DIFF
--- a/OpenOversight/app/main/model_view.py
+++ b/OpenOversight/app/main/model_view.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 from http import HTTPMethod
 from typing import Callable, Union
 
@@ -222,7 +221,6 @@ class ModelView(MethodView):
             obj.created_by = current_user.id
         # if the object keeps track of who updated it last, set to current user
         if hasattr(obj, "last_updated_at"):
-            obj.last_updated_at = datetime.now()
             obj.last_updated_by = current_user.id
 
         db.session.add(obj)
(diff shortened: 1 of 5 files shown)


OpenOversight/app/main/model_view.py
```diff
@@ -222,7 +221,6 @@ def populate_obj(self, form, obj):
             obj.created_by = current_user.id
         # if the object keeps track of who updated it last, set to current user
         if hasattr(obj, "last_updated_at"):
```
sea-kelp: Can you change this to "last_updated_by"?

OpenOversight/app/models/database_imports.py
```diff
@@ -338,12 +334,15 @@ def update_incident_from_dict(data: Dict[str, Any], incident: Incident) -> Incid
             )
         )
     if "last_updated_by" in data:
-        incident.last_updated_by = parse_int(data.get("last_updated_by"))
+        incident.last_updated_by = parse_int(
+            data.get(
+                "last_updated_by", User.query.filter_by(is_administrator=True).first()
```
sea-kelp, line 339: Would it be better to just leave it as null?

sea-kelp (commented):
Looks good! A couple comments:

OpenOversight/app/models/database_imports.py
michplunkett: We're doing this for the rest of the last_updated_by fields, so I'd at least prefer it be the same for all updates. We could modify it to be either empty or this some other time, imo.

OpenOversight/app/main/model_view.py
michplunkett: Done!

sea-kelp (approved):

PR #1064, opened 2023-09-22
Deploy to prod on tag creation

### LINKED ISSUE #1030: Create a simpler deployment process
## What issue are you seeing?
We currently put all changes to the `develop` branch, merge the changes we want into `main`, create a tag on `main`, and then merge that back into `develop`. Since we are making use of squash-commits, each commit to `develop` is a discrete feature, and ALL commits are approved by at least one person, I'd propose that we just create tags off of `develop`. This way we could make our process one step as oppose to four highly redundant steps.

abandoned-prototype:
## Fixes issue
#1030

## Description of Changes
Previously a push on the main branch triggered a deploy to prod. In the future, creating a release will trigger that push.
The change is already built in a way that we can delete the main branch and rename the develop branch to main as a next step (once we confirmed everything works).

## Notes for Deployment
I was only able to partially test this, so we will see how it goes.

## Tests and linting

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -1,17 +1,21 @@
 name: Publish Docker image and deploy it
 
 on:
+  release:
+    types: [published]
   push:
     branches:
-      - develop
       - main
+      - develop
 
 env:
   image_name: lucyparsons/openoversight
-  # env_mapping: {"develop": "staging", "main": "prod"}
-  env_name: ${{ fromJson(vars.env_mapping)[github.ref_name] }}
-  # docker_mapping: {"develop": "latest", "main": "stable"}
-  docker_name: ${{ fromJson(vars.docker_mapping)[github.ref_name] }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_deploy.yml
```diff
@@ -1,17 +1,21 @@
 name: Publish Docker image and deploy it
 
 on:
+  release:
+    types: [published]
   push:
     branches:
-      - develop
       - main
```
michplunkett, line 8: I think it may be worth taking off `main` if we are going to scrap it at some point.

michplunkett (approved):
I think there'd be some value to removing the `main` branch mention, but it's not worth holding back on the merge.

Otherwise, I'm very excited for this to happen!

abandoned-prototype:
Leaving both main and develop for now allows us to remove the main branch and rename develop to main if we want at a later point. (And then clean up this workflow)

PR #1073, opened 2023-11-19
Replace inline styles with external css

### LINKED ISSUE #969: Remove `H021` from `pre-commit` ignore list and make CSS classes/IDs
## Related Issue(s):
https://github.com/lucyparsons/OpenOversight/issues/239

## What issue are you seeing?
We currently have A LOT of inline styles and they will take some care to fix. This will be a good first issue for someone to contribute to the repository and would help us get our HTML in fantastic shape.

## Steps
1. Remove `H021` from the `ignore` list in the `.pre-commit-config.yaml` file.
Screenshot 2023-07-12 at 12 59 00 PM

2. Make the necessary changes.

brrusselburg:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/969

## Description of Changes
Removes the H021 tag and does the relevant clean-up. In this case, inline styles are refactored into external CSS via pre-existing stylesheets.

Unsure of the CSS class names, very open to changes. One in particular I really disliked, but left a note-to-self. Historically more of a back-end dev, so apologies if I missed something obvious with these changes!

## Tests and linting
  
(`pre-commit` technically fails on some .py files and migrations, but none of the new changes it seems)

### DIFF
--- a/OpenOversight/app/static/css/openoversight.css
+++ b/OpenOversight/app/static/css/openoversight.css
@@ -654,3 +654,28 @@ tr:hover .row-actions {
     visibility: hidden;
     width: auto
 }
+
+.red {
+  color: #a00;
+}
+
+.no-box-shadow {
+  box-shadow: none;
+}
+
+.no-border-top {
+  border-top: none;
+}
+
+.no-display {
+  display: none;
+}
+
+.slightly-above {
+  position: relative;
+  top: -0.8em;
+}
+
+.bottom-margin {
+  margin-bottom: 2rem;
+}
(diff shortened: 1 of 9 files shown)


michplunkett (approved):
Noice! Well done and thank you for the quick contribution.

OpenOversight/app/static/css/openoversight.css
```diff
@@ -654,3 +654,28 @@ tr:hover .row-actions {
...
+
+.no-border-top {
+  border-top: none;
+}
+
+.no-display {
+  display: none;
+}
+
+.slightly-above {
```
michplunkett, line 674: This is def kind of an odd name, but CSS IDs/classes always tend to be a bit odd. We're a far way away from having a standard naming convention. These names all work great.

PR #1078, opened 2023-12-27
Fix `Departments` pagination automatic filter application

### LINKED ISSUE #1077: Departments Pagination Automatically Applies Filters
## What issue are you seeing?

When I am browsing Officers of a Police Department and click the next/previous buttons to go to another page, the next page I visit automatically applies filters (outside of the `page` filter), which removes Officers from the list being browsed through.

Spraynard:
Implementing behavior that will allow use of pagination without automatically applying filters

## Fixes issue
 
Fixes #1077 

## Description of Changes
Changes check if there are "filters" being applied to the list given in the `departments` view. If there are no filters applied then we generate a "next" link that does not automatically apply filters.

## Tests and linting


sea-kelp:
Hello, thank you for reporting this bug! I took a look and I think the bug you're experiencing might be related to OpenOversight automatically adding the `require_photo` parameter to the url when you click to the next/previous page. Can you please try removing the "&require_photo=True" url parameter to see if that yields your expected search results?

If that works, we should be able to fix this issue by having the [redirect_list_officer](https://github.com/lucyparsons/OpenOversight/blob/76815426303da3bed2d24d2b2ed6329046a8216e/OpenOversight/app/main/views.py#L845-L860) and [list_officer](https://github.com/lucyparsons/OpenOversight/blob/76815426303da3bed2d24d2b2ed6329046a8216e/OpenOversight/app/main/views.py#L885-L900) routes default the `require_photo` parameter to `None` instead of `False` :slightly_smiling_face:

Spraynard:
Hello,

Yes from looking at the bug it seems that removing the `require_photo=FALSE` parameter from the URL gave me the next page with the "correct" amount of Officer records in the total list.

So that means the bug isn't due to filters being applied but because of that `require_photo` parameter having a value that gives us unintended consequences.

This pull request should now have changes that provide a fix to this buggy behavior

sea-kelp (approved):
This looks good to me! Would you please update the type annotation to `Optional[bool]`?

Spraynard:
Sure thing. Added the optional typings

sea-kelp:
Thank you for making those changes! Tagging in @michplunkett and @abandoned-prototype to take a second look

michplunkett:
Taking a look!

michplunkett:
Is there a ticket this is based off of, @Spraynard?

michplunkett (commented):
This is a great find! Could you add a validation test similar to the `test_officer_browse_pagination` function in `test_functional.py`?

sea-kelp:
@michplunkett Added a quick functional test - would you mind taking a look?

michplunkett (approved):
Excellent work, @Spraynard! Thank you for making the test addition, @sea-kelp, I appreciate it.

PR #1087, opened 2024-01-12
Add validation to Incident form

### LINKED ISSUE #839: Incidents can be linked to officers that don't belong to the incident department
Desired outcome:
- If a user attempts to create an incident for a department, and then selects officers from another department as being involved in that incident, the user will see an error message

Current outcome:
- If a user attempts to create an incident for a department, and then selects officers from another department as being involved in that incident, the incident is created and persisted to the database 

Steps to reproduce:
1) Log in as admin and navigate to `/incidents`
2) Create an incident
3) Choose a department for the incident
4) Select officers that don't work for that department, and link them to the incident

Relevant areas of the codebase: 
- `IncidentForm` in `main/forms.py`
- `IncidentApi` in `main/views.py`

Spraynard:
## Fixes issue

Fixes #839 

## Description of Changes

Adds inline validator to `IncidentForm` that checks to see if the officer entities linked to the instance are in the same department as the incident.

Screen Shot 2024-01-11 at 2 25 45 AM

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/forms.py
+++ b/OpenOversight/app/main/forms.py
@@ -516,12 +516,19 @@ class IncidentForm(DateFieldForm):
     # In line validator for the officers field. Check to see if the officers are within the selected department
     def validate_officers(form, field):
         department = form.department.data
+
+        if not department:
+            return
+        
         department_id = department.id
         department_name = department.name
 
         for field_data in field.data:
             officer = Officer.query.get(field_data['oo_id'])
 
+            if not officer:
+                return
+            
             if officer.department.id is not department_id:
                 raise ValidationError(f"Officer {officer.full_name()} (OpenOversight ID: {field_data['oo_id']}) is not in {department_name}")
         
--- a/OpenOversight/app/main/forms.py
+++ b/OpenOversight/app/main/forms.py
@@ -515,7 +515,7 @@ class OOIdForm(Form):
 
 class IncidentForm(DateFieldForm):
     # In line validator for the officers field. Check to see if the officers are within the selected department
-    def validate_officers(self, field: FieldList) -> None: 
+    def validate_officers(self, field: FieldList) -> None:
         department = self.department.data
 
         if not department:
--- a/OpenOversight/app/main/forms.py
+++ b/OpenOversight/app/main/forms.py
@@ -512,6 +512,7 @@ def validate_oo_id(self, oo_id):
 class OOIdForm(Form):
     oo_id = StringField("OO Officer ID", validators=[validate_oo_id])
 
+
 class IncidentForm(DateFieldForm):
     # In line validator for the officers field. Check to see if the officers are within the selected department
     def validate_officers(form, field):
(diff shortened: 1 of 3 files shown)


OpenOversight/app/main/forms.py
```diff
@@ -514,6 +514,27 @@ class OOIdForm(Form):
 
 
 class IncidentForm(DateFieldForm):
+    # In line validator for the officers field. Check to see if the officers are within the selected department
+    def validate_officers(form, field):
```
michplunkett: Add a return type to the `validate_officers` function.

michplunkett (changes requested):
Hey @Spraynard, a couple of things:
- Could you change the title so that it mentions the feature fix in the present imperative tense as per the PR template?
- Could you add a small test to validate these changes?

Spraynard:
@michplunkett Sure thing! I'll add in a test when I'm next able. Thanks for looking at this

OpenOversight/app/main/forms.py
Spraynard: Please let me know if the additional changes are satisfactory! Thanks!

michplunkett: It looks good! It looks like the tests are failing and it'd be ideal if you could use the `faker` package on the tests, similar to what you'll see on the others.

OpenOversight/app/main/forms.py
```diff
@@ -514,6 +514,27 @@ class OOIdForm(Form):
...
+        department_id = department.id
+        department_name = department.name
+
+        for field_data in field.data:
+            officer = Officer.query.get(field_data["oo_id"])
+
+            if not officer:
+                return None
+
+            if officer.department.id is not department_id:
```
sea-kelp, line 533: `==` should be used for value comparison

OpenOversight/app/main/forms.py
```diff
@@ -514,6 +514,27 @@ class OOIdForm(Form):
...
+            return
+
+        department_id = department.id
+        department_name = department.name
+
+        for field_data in field.data:
+            officer = Officer.query.get(field_data["oo_id"])
+
+            if not officer:
+                return None
```
sea-kelp, line 531: I think we should raise an error here instead of failing silently

sea-kelp (commented):
Thank you for working on this!

IMO there have been cases where multiple departments (or taskforces involving officers from multiple departments) have been involved in an incident. [This incident](https://www.npr.org/2020/09/04/909515885/protester-suspected-in-portland-shooting-death-killed-by-law-enforcement) comes to mind as one example:

> Four officers fired their weapons, according to Brady. They're from the Pierce County Sheriff's Department, Lakewood Police Department and the Washington State Department of Corrections, all assigned to the fugitive task force acting in their role as U.S. marshals, he said.

In cases like these, I feel that adding this restriction would limit the scope of incidents that could be expressed in OpenOversight.

The Seattle fork implemented [a feature](https://github.com/OrcaCollective/OpenOversight/pull/254) a while back to identify officers by department + serial number while identifying officers; I wonder if we could do something similar here to prevent input errors.

@michplunkett: Thoughts?

Spraynard:
@sea-kelp I honestly feel the same way, that it is _technically_ possible for another department's officer to be involved in an incident! I'll wait to see what other people's thoughts are about what you said before I continue on this!

b-meson:
Hi all,

I agree with @sea-kelp, sometimes there are incidents that officers take part of that are outside of their jurisdiction but I'm worried that if we allow police to be tagged in incidents outside of their departments we will have unhappy secondary effects. 

I wonder if the way to do this properly is to 'untether' incidents from police departments. While you can have an officer and a department linked to an incident, should incidents be "assigned" to a particular PD? (I think not necessary?)

sea-kelp:
I think that would make sense as long as we are still able to search incidents by involved officer departments (for example, "all incidents that involved a Chicago PD officer and a CBP officer between 2020-01-01 and 2021-01-01").

b-meson:
@sea-kelp That's what I was thinking as well.

Spraynard:
Alright I will close this pull request and I will put some context on the issue

b-meson:
Thanks for the PR @Spraynard and all your work on this! 

Just so I understand and your hard work doesn't go unused, is the new plan to move incidents to be more generic (away from a specific PD) and removes the need for the validation check? But in the future, we still want to revisit the validations and tests for "incidents"? Is that right?

PR #1094, opened 2024-04-06
Remove `jsonify` and use `Response`

michplunkett:
## Description of Changes
Flask automatically renders the `Response` as JSON, so the `jsonify` function is redundant. Added missing `HTTPMethods` as well.

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -376,10 +376,7 @@ def redirect_add_assignment(officer_id: int):
     )
 
 
-@main.route(
-    "/officers/<int:officer_id>/assignments/new",
-    methods=[HTTPMethod.POST],
-)
+@main.route("/officers/<int:officer_id>/assignments/new", methods=[HTTPMethod.POST])
 @ac_or_admin_required
 def add_assignment(officer_id: int):
     form = AssignmentForm()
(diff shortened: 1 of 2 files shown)


OpenOversight/app/main/views.py
```diff
@@ -368,7 +366,7 @@ def sitemap_officers():
 
 @main.route(
     "/officer/<int:officer_id>/assignment/new",
-    methods=[HTTPMethod.GET, HTTPMethod.POST],
+    methods=[HTTPMethod.POST],
```
michplunkett: Validated by the `test_route_post_only` test.

sea-kelp:
This seems to break some things, particularly where we are returning string where we used to return JSON errors
```bash
# Before
$ curl -X GET -H 'Accept: application/json' http://localhost:3000/asdf
{
  "error": "Not Found"
}

# After
$ curl -X GET -H 'Accept: application/json' http://localhost:3000/asdf
Not Found
```

michplunkett:
Oof, good find. It seems to work in terms of application logic, but not so much for external calls to the API.

michplunkett:
I may just get rid of the `Response` classes in that case and put the `jsonify` functions back in place.

PR #968, opened 2023-07-12
Add linting to `pre-commit`

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/239

## Description of Changes
I added `djlint` linting to the `pre-commit` script and made the corresponding changes.

## Tests and linting
 - [x] This branch is up-to-date with the `develop` branch.
 - [x] `pytest` passes on my local development environment.
 - [ ] `pre-commit` passes on my local development environment.

PR #1101, opened 2024-06-03
Replace `flake8` with `ruff`

michplunkett:
## Description of Changes
Replacing `flake8` with `ruff`, which has approximate parity and is more than 10x faster: https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-flake8

## Tests and Linting

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -76,14 +76,16 @@ repos:
           args:
             - --quiet
 
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.0.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.7
     hooks:
-        - id: flake8
-          args:
-            - --ignore=E203,E402,E501,E800,W503,W391,E261,W504
-            - --select=B,C,E,F,W,T4,B9
-            - --exclude=./OpenOversight/app/db_repository/versions/
+      - id: ruff
+        types_or: [python, pyi, jupyter]
+        args:
+          - --fix
+          - --select=B,C,E,F,W
+          - --ignore=E203,E402,E501,W391,E261
+          - --exclude=./OpenOversight/app/db_repository/versions/
 
   - repo: https://github.com/psf/black
     rev: 23.7.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,7 +84,7 @@ repos:
         args:
           - --fix
           - --select=B,C,E,F,W
-          - --ignore=E203,E402,E501,W391,E261
+          - --ignore=C901,E203,E402,E501,W391,E261
           - --exclude=./OpenOversight/app/db_repository/versions/
 
   - repo: https://github.com/psf/black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -85,7 +85,6 @@ repos:
           - --fix
           - --select=B,C,E,F,W
           - --ignore=C901,E203,E402,E501,W391,E261
-          - --exclude=./OpenOversight/app/db_repository/versions/
 
   - repo: https://github.com/psf/black
     rev: 23.7.0
--- a/OpenOversight/app/commands.py
+++ b/OpenOversight/app/commands.py
@@ -352,7 +352,9 @@ def process_assignment(row, officer, compare=False):
                         current
                         and field_name in row
                         and is_equal(row[field_name], current)
-                    ) or (not current and (field_name not in row or not row[field_name])):
+                    ) or (
+                        not current and (field_name not in row or not row[field_name])
+                    ):
                         i += 1
                 if i == len(assignment_field_names):
                     job_title = job.job_title
--- a/OpenOversight/app/commands.py
+++ b/OpenOversight/app/commands.py
@@ -230,7 +230,7 @@ def update_officer_from_row(row, officer, update_static_fields=False):
                     new_value = parse(row[field_name])
                     if isinstance(old_value, date):
                         new_value = new_value.date()
-                except Exception as e:
+                except Exception as err:
                     msg = (
                         'Field {} is a date-type, but "{}" was specified for '
                         "Officer {} {} and cannot be parsed as a date-type.\nError "
@@ -239,10 +239,10 @@ def update_officer_from_row(row, officer, update_static_fields=False):
                             row[field_name],
                             officer.first_name,
                             officer.last_name,
-                            e,
+                            err,
                         )
                     )
-                    raise Exception(msg) from e
+                    raise Exception(msg) from err
             else:
                 new_value = row[field_name]
             if old_value is None:
--- a/OpenOversight/app/commands.py
+++ b/OpenOversight/app/commands.py
@@ -230,7 +230,7 @@ def update_officer_from_row(row, officer, update_static_fields=False):
                     new_value = parse(row[field_name])
                     if isinstance(old_value, date):
                         new_value = new_value.date()
-                except ValueError as err:
+                except ValueError as e:
                     msg = (
                         'Field {} is a date-type, but "{}" was specified for '
                         "Officer {} {} and cannot be parsed as a date-type.\nError "
@@ -242,7 +242,7 @@ def update_officer_from_row(row, officer, update_static_fields=False):
                             e,
                         )
                     )
-                    raise Exception(msg) from err
+                    raise Exception(msg) from e
             else:
                 new_value = row[field_name]
             if old_value is None:
--- a/OpenOversight/app/commands.py
+++ b/OpenOversight/app/commands.py
@@ -230,7 +230,7 @@ def update_officer_from_row(row, officer, update_static_fields=False):
                     new_value = parse(row[field_name])
                     if isinstance(old_value, date):
                         new_value = new_value.date()
-                except ValueError as e:
+                except ValueError as err:
                     msg = (
                         'Field {} is a date-type, but "{}" was specified for '
                         "Officer {} {} and cannot be parsed as a date-type.\nError "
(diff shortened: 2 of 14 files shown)


OpenOversight/tests/test_models.py
```diff
@@ -173,8 +173,7 @@ def test_salary_repr(mockdata):
 
 def test_password_not_printed(mockdata):
     user = User(password="bacon")
-    with raises(AttributeError):
-        user.password
```
michplunkett, line 177: Not entirely sure what this was testing.

OpenOversight/app/formfields.py
```diff
@@ -19,9 +19,9 @@ def _value(self):
         else:
             return self.data and self.data.strftime(self.format) or ""
 
-    def process_formdata(self, valuelist):
```
michplunkett, line 22: `valuelist` is an un-pythonic name.

.pre-commit-config.yaml
```diff
@@ -76,14 +76,16 @@ repos:
...
-          args:
-            - --ignore=E203,E402,E501,E800,W503,W391,E261,W504
-            - --select=B,C,E,F,W,T4,B9
-            - --exclude=./OpenOversight/app/db_repository/versions/
+      - id: ruff
+        types_or: [python, pyi, jupyter]
+        args:
+          - --fix
+          - --select=B,C,E,F,W
+          - --ignore=C901,E203,E402,E501,W391,E261
```
michplunkett, line 87: Ignored `C901`: https://www.flake8rules.com/rules/C901.html

OpenOversight/app/utils/forms.py
```diff
@@ -86,46 +86,46 @@ def add_officer_profile(form: AddOfficerForm, user: User) -> Officer:
     )
     db.session.add(assignment)
     if form.links.data:
-        for link in form.data["links"]:
-            if link["url"]:
-                li = get_or_create_link_from_form(link, user)
-                officer.links.append(li)
+        form_links = [link for link in form.data["links"] if link["url"]]
+        for link in form_links:
```
abandoned-prototype, line 90: Not sure I would agree with the changes in this file, but also don't feel strongly about it. Was this something ruff suggested?

abandoned-prototype (approved):
This looks like a nice improvement. Thanks!

OpenOversight/app/utils/forms.py
michplunkett: The changes were made due to [this rule](https://www.flake8rules.com/rules/C901.html). It called out 10+ other functions that had more than 10 if/else statements, and I changed the ones that could be done with the least amount of re-working. 

If you'd like to see the other functions, you could remove `C901` from the ignore list and rerun the linter.

michplunkett: I think we could reimplement `C901` in another PR, but it would require a lot of changes:
```console
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

OpenOversight/app/commands.py:183:5: C901 `update_officer_from_row` is too complex (14 > 10)
OpenOversight/app/commands.py:473:5: C901 `bulk_add_officers` is too complex (14 > 10)
OpenOversight/app/csv_imports.py:156:5: C901 `_handle_assignments_csv` is too complex (18 > 10)
OpenOversight/app/main/model_view.py:113:9: C901 `edit` is too complex (11 > 10)
OpenOversight/app/main/views.py:745:5: C901 `edit_department` is too complex (15 > 10)
OpenOversight/app/main/views.py:883:5: C901 `list_officer` is too complex (18 > 10)
OpenOversight/app/main/views.py:1374:5: C901 `label_data` is too complex (11 > 10)
OpenOversight/app/models/database_imports.py:102:5: C901 `update_officer_from_dict` is too complex (11 > 10)
OpenOversight/app/utils/forms.py:273:5: C901 `filter_by_form` is too complex (21 > 10)
Found 9 errors.

black....................................................................Passed
mypy.....................................................................Passed
```

.pre-commit-config.yaml
```diff
@@ -76,14 +76,15 @@ repos:
...
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.0.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.7
     hooks:
-        - id: flake8
-          args:
-            - --ignore=E203,E402,E501,E800,W503,W391,E261,W504
-            - --select=B,C,E,F,W,T4,B9
-            - --exclude=./OpenOversight/app/db_repository/versions/
```
michplunkett, line 86: This folder no longer exists, so I removed the `exclude`.

OpenOversight/tests/test_models.py
sea-kelp: I think this is testing that you can set but not recover the plaintext password through user.password
https://github.com/lucyparsons/OpenOversight/blob/5e0e0d711e287e15e664c99dde1fd364443dd247/OpenOversight/app/models/database.py#L752-L761

sea-kelp (commented):
I haven't had time to review all the changes in this PR but I want to push back against large refactoring changes like this without good reason. 

The previous [djlint](https://github.com/lucyparsons/OpenOversight/pull/968/files) change introduced several minor bugs which weren't surfaced until the changes were live in production due to the scale of the refactoring. Comparing the build times, this [ruff change](https://github.com/lucyparsons/OpenOversight/actions/runs/9371416615) built in 5m 26s while an [earlier build](https://github.com/lucyparsons/OpenOversight/actions/runs/8935754940) took 4m 30s.

I mention this because our fork will need to pull in changes like this to keep in sync with the upstream. I definitely don't mean to block adding important guardrails but wanted to get a temp check since we expect ruff to have rough parity with our current setup and aren't seeing the performance enhancements.

OpenOversight/tests/test_models.py
michplunkett: I can change the assert in that case!

michplunkett: Done!

michplunkett:
@sea-kelp, I definitely get the hesitancy. I BELIEVE most every part of the code modified here is covered in test cases. Would you feel more comfortable with these changes if I were to verify that every line change was covered by tests?

I also think there's a difference between changes in HTML versus Python syntax changes. The Python changes can be verified through a handful of mechanisms that are generally unavailable to HTML changes. Most of the changes here are also syntax changes that are more in line with Python best practices, which I think is of value as well (e.g., `isinstance` versus type comparisons, removal of redundant `map` function calls, etc.).

If there's a specific change you're curious or have questions about, I'd certainly be willing to go more into it.

OpenOversight/tests/test_models.py
sea-kelp: I think this test would still pass if `user.password` returns the password. Is there a reason we don't want to use the original `pytest.raises`?

OpenOversight/tests/routes/route_helpers.py
```diff
@@ -75,22 +75,21 @@ def process_form_data(form_dict: dict) -> dict:
...
+                for idx, item in enumerate(value):
+                    new_dict[f"{key}-{idx}"] = item
+            else:
+                raise ValueError(
+                    "Lists must contain dicts, strings or ints. {} submitted".format(
+                        type(value[0])
                     )
-        elif type(value) == dict:
+                )
+        elif isinstance(value, dict):
```
sea-kelp: Previously, if `value` is empty list (`[]`), it would run the `if type(value) == list` branch and not set a value in `new_dict`.

With the new change, empty list will cause an error on line 78 when trying to access `value[0]`.

OpenOversight/app/utils/forms.py
```diff
@@ -170,35 +170,32 @@ def create_incident(self, form: IncidentForm, user: User) -> Incident:
...
-        for link in form.data["links"]:
-            if link["url"]:
-                li = get_or_create_link_from_form(link, user)
-                links.append(li)
+    form_officers = [o for o in form.data["officers"] if o["oo_id"]]
+    for officer in form_officers:
+        of = Officer.query.filter_by(id=int(officer["oo_id"])).one()
+        if of:
+            officers.append(of)
+
```
sea-kelp: Why did we remove the check for `"officers" in form.data`?

OpenOversight/app/commands.py
```diff
@@ -230,7 +230,7 @@ def update_officer_field(officer_field_name):
                     new_value = parse(row[field_name])
                     if isinstance(old_value, date):
                         new_value = new_value.date()
-                except Exception as e:
+                except ValueError as e:
```
sea-kelp: Do we know how was this subclass was chosen?

OpenOversight/app/utils/forms.py
michplunkett: It was due to this: https://github.com/lucyparsons/OpenOversight/pull/1101#discussion_r1626305791

michplunkett: I BELIEVE we should be good in that space, but I could also check with a test to confirm we are, should one not already be in place.

OpenOversight/tests/test_models.py
michplunkett: I modified it because we had a hanging `user.password` statement, but I could modify it so that it validates both conditions.

OpenOversight/tests/routes/route_helpers.py
michplunkett: I'll take another look!

OpenOversight/app/commands.py
michplunkett: I chose it through a manual lookup.

michplunkett:
Thanks for taking a look, @sea-kelp! I'll let you know when the changes are made.

OpenOversight/tests/test_models.py
michplunkett: Done done.

OpenOversight/app/commands.py
michplunkett: I just defaulted back to `Exception` to be safe.

PR #1105, opened 2024-06-17
Upgrade to Bootstrap 5

sea-kelp:
## Description of Changes
*This change is cherry-picked from https://github.com/OrcaCollective/OpenOversight/pull/458*
Upgrade to Bootstrap 5 in preparation for upcoming changes. Everything looks slightly different due to changes in the bootstrap framework.

## Notes for Deployment
None

## Tests and Linting
Manually checked that all pages look roughly reasonably similar to the previous look

### DIFF
--- a/OpenOversight/app/templates/about.html
+++ b/OpenOversight/app/templates/about.html
@@ -11,47 +11,59 @@
     <div class="hero-section no-sub">
       <div class="hero">About OpenOversight</div>
     </div>
-    <div class="vertical-padding">
-      <p class="lead">
-        <small>
-          OpenOversight is a <a href="https://lucyparsonslabs.com">Lucy Parsons Labs</a> project that aims to improve law enforcement
-          visibility and transparency using public and crowdsourced data. We maintain databases, digital galleries, and profiles of individual law enforcement officers from departments
-          across the United States that consolidate information including names, birthdates, mentions in news articles, salaries, and photographs.
-        </small>
-      </p>
-      <div class="text-center pb-50">
-        <a href="{{ url_for("main.get_officer") }}"
-           class="btn btn-lg btn-info"
-           id="try-it">Try it</a>
-      </div>
-      <div class="row font-weight-300 text-center display-flex">
-        <div class="col-lg-4 col-half-border">
-          <span class="fa fa-institution fa-2x"></span>
-          <p>
-            This project is a response to the lack of transparency and justice in policing. The public should have the right to know which officers are patrolling their neighborhoods and watching their communities. When officers abuse their positions of power, they should be able to be easily identified and held accountable.
-          </p>
-        </div>
-        <div class="col-lg-4 col-half-border">
-          <span class="fa fa-map-marker fa-2x"></span>
-          <p>
-            It is the first project of its kind in the United States, and was first implemented in Chicago in October 2016.
-            OpenOversight launched in the East Bay of the San Francisco Bay Area in fall 2017 and in New York City in 2018. A
-            Baltimore instance was launched in 2019 at <a href="https://bpdwatch.com">BPDWatch.com</a>.
-          </p>
-        </div>
-        <div class="col-lg-4 col-half-border">
-          <span class="fa-brands fa-github fa-2x"></span>
-          <p>
-            OpenOversight is released as free and open source software so others can launch similar law enforcement accountability
-            projects in their own cities. The software is available for download and collaborative development on
-            <a href="https://github.com/lucyparsons/OpenOversight">GitHub</a>.
-          </p>
+    <div class="vertical-padding horizontal-padding text-center">
+      <div class="col-lg-6 text-start mx-auto">
+        <p class="lead">
+          <small>
+            OpenOversight is a <a href="https://lucyparsonslabs.com">Lucy Parsons Labs</a> project that aims to improve law enforcement
+            visibility and transparency using public and crowdsourced data. We maintain databases, digital galleries, and profiles of individual law enforcement officers from departments
+            across the United States that consolidate information including names, birthdates, mentions in news articles, salaries, and photographs.
+          </small>
+        </p>
+        <div class="text-center pb-50">
+          <a href="{{ url_for("main.get_officer") }}"
+             class="btn btn-lg btn-info"
+             id="try-it">Try it</a>
         </div>
       </div>
     </div>
-    <div class="container-fluid">
-      <div class="row display-flex  horizontal-padding vertical-padding pb-50">
-        <div class="col-lg-6 col-half-no-border bg-light-blue">
+    <div class="vertical-padding horizontal-padding">
+      <div class="row font-weight-300 text-center display-flex sink-row">
+        <div class="col-lg-4">
+          <div class="col-half-border">
+            <span class="fa fa-institution fa-2x"></span>
+            <p>
+              This project is a response to the lack of transparency and justice in policing. The public should have the right to know which officers are patrolling their neighborhoods and watching their communities. When officers abuse their positions of power, they should be able to be easily identified and held accountable.
+            </p>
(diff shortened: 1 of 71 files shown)


AetherUnbound (approved):
Confirmed the changes look good, thanks so much for this huge effort!

OpenOversight/app/templates/about.html
```diff
@@ -11,47 +11,59 @@
...
-          visibility and transparency using public and crowdsourced data. We maintain databases, digital galleries, and profiles of individual law enforcement officers from departments
-          across the United States that consolidate information including names, birthdates, mentions in news articles, salaries, and photographs.
-        </small>
-      </p>
-      <div class="text-center pb-50">
-        <a href="{{ url_for("main.get_officer") }}"
-           class="btn btn-lg btn-info"
-           id="try-it">Try it</a>
+    <div class="vertical-padding horizontal-padding text-center">
+      <div class="col-lg-6 text-start mx-auto">
```
michplunkett, line 15: Should this be `col-xs-12 col-lg-6` or does it default to 12 if you don't have anything before a certain frame size?

OpenOversight/app/templates/base.html
```diff
@@ -41,142 +39,142 @@
...
-              {% endif %}
-            </ul>
-          </div>
+            {% endif %}
+          </ul>
         </div>
       </div>
     </nav>
     {% with messages = get_flashed_messages() %}
       {% if messages %}
```
michplunkett, line 123: 🔬 Do we need the `if` here?

OpenOversight/app/templates/edit_officer.html
```diff
@@ -1,28 +1,28 @@
 {% extends "base.html" %}
-{% import "bootstrap/wtf.html" as wtf %}
```
michplunkett, line 2: Didn't see any usages of `import "bootstrap/wtf.html"` in the code. ✅

OpenOversight/app/templates/link_delete.html
```diff
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 {% block content %}
-  <div class="container theme-showcase" role="main">
+  <div class="container theme-showcase py-5" role="main">
```
michplunkett, line 3: 🔬  Should we add `py-5` as an addition to all of the `container` classes? Are they always used together?

OpenOversight/app/templates/list_officer.html
```diff
@@ -12,16 +12,18 @@
...
+  <div class="container pb-5" role="main">
+    <div class="frontpage-leads">
+      <h1>{{ department.display_name }} Officers</h1>
+    </div>
     <div class="row">
       <div class="filter-sidebar col-sm-3">
         <h3 class="sidebar-title">Filter officers</h3>
         <form class="form" method="get" role="form">
           <div class="panel">
-            <div class="panel-heading{% if not form_data['first_name'] and not form_data['last_name'] %} collapsed{% endif %}"
```
michplunkett, line 22: Should we keep the `if` here like we did below?

michplunkett (approved):
Great work! I do have a couple of questions, but there's nothing major.

OpenOversight/app/templates/link_delete.html
sea-kelp: `py-5` adds padding in the outer-most container of content blocks in all templates (for example: https://spd.watch/browse) with the exception of templates with grey "hero section" elements (for example: https://spd.watch/). I've checked all pages manually to make sure they look right but let me know if there are any that look broken.

OpenOversight/app/templates/edit_officer.html
sea-kelp: I'm seeing that `wtf` is still used in this file, we're just replacing it with `bootstrap5/form.html`

OpenOversight/app/templates/about.html
sea-kelp: `col-12` adds the style `width: 100%`, which this element defaults to, but I can add that class to avoid confusion

michplunkett: If it's already defaulted to that, I have no qualms with leaving it out.

OpenOversight/app/templates/link_delete.html
michplunkett: Since it's just the outer-most container, this seems good to me as is!

OpenOversight/app/templates/about.html
sea-kelp: Actually, it looks like this is what we already use today so keeping as is
https://github.com/lucyparsons/OpenOversight/blob/d989873c3640b7deed1d9bd528cacc3bc807148f/OpenOversight/app/templates/about.html#L28-L49

OpenOversight/app/templates/base.html
sea-kelp: Do you mean the `if messages` block? It's needed because the outer container still adds margins and other styles, even if there are no messages

OpenOversight/app/templates/list_officer.html
sea-kelp: Good catch! I think this is an artifact of cherry-picking from our instance since we always show the name panel

OpenOversight/app/templates/auth/user.html
```diff
@@ -1,19 +1,23 @@
...
-  <div class="container theme-showcase" role="main">
+  <div class="container theme-showcase py-5" role="main">
     <div class="page-header">
       <h1>
         Update {{ user.username }}
         <small>{{ user.email }}</small>
       </h1>
     </div>
-    <div class="col-md-6 col-xs-12">
+    <div class="col-md-6 col-12">
```
sea-kelp, line 14: Replaced `col-xs` classes with `col-` to follow the change in Bootstrap 4
https://getbootstrap.com/docs/4.0/migration/#grid-system

OpenOversight/app/templates/base.html
michplunkett: That's what I was talking about, but it seems we're good!

OpenOversight/app/templates/about.html
michplunkett: I can make a ticket to standardize how we use bootstrap later on. Keeping it as is feels right to me.

OpenOversight/tests/test_functional.py
```diff
@@ -73,41 +73,28 @@ def wait_for_element_to_be_visible(browser, locator, text, timeout=10):
...
-        browser.find_element_by_id("cpd").click()
+    element = browser.find_element(By.ID, "cpd")
+    scroll_to_element(browser, element)
+    element.click()
 
     page_text = browser.find_element_by_tag_name("body").text
     assert "Find an Officer" in page_text
 
 
-def test_user_can_use_form_to_get_to_browse(mockdata, browser, server_port):
```
abandoned-prototype, line 88: Can you elaborate why this test had to be removed? Was this functionality removed, or did the test no longer work with the updated framework?

abandoned-prototype (approved):
Thank you for making this upgrade, and especially for also bringing it into this repo! Excited for the upcoming changes!

OpenOversight/tests/test_functional.py
sea-kelp: Yes, thanks for the question!

This javascript step-by-step dialogue was removed in this PR in favor of having all fields laid out on the page, in preparation for the upcoming layout changes. See https://spd.watch/find for an example of what it will look like.

Because of that change, it didn't make sense to keep this test around since the form submission functionality is covered by [this test](https://github.com/lucyparsons/OpenOversight/blob/bfd066a771b415915e1d368af1a00776f3cd3635/OpenOversight/tests/routes/test_officer_and_department.py#L1979-L2028)

PR #1106, opened 2024-06-23
Remove `black` and use `ruff` formatter

michplunkett:
## Description of Changes
Removed the black formatter put the ruff formatter in its place: https://docs.astral.sh/ruff/formatter/

## Tests and Linting

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -77,16 +77,16 @@ repos:
             - --quiet
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.7
+    rev: v0.4.10
     hooks:
       - id: ruff
-        types_or: [python, pyi, jupyter]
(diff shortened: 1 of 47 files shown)


.pre-commit-config.yaml
```diff
@@ -77,22 +77,18 @@ repos:
             - --quiet
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.4.7
+    rev: v0.4.10
     hooks:
       - id: ruff
-        types_or: [python, pyi, jupyter]
```
michplunkett, line 83: Ruff isn't great at modifying Jupyter notebook files (as validated in my other repos), so I removed it from our list. If we end up wanting that functionality, I think the [nbQA](https://github.com/nbQA-dev/nbQA) package would be a far better choice.

b-meson (approved):
thanks for this PR. it looks great.

PR #1118, opened 2024-07-29
Update pre-commit libraries

michplunkett:
## Description of Changes
Updating and removing unnecessary pre-commit packages.

## Tests and Linting

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,7 @@ repos:
       - id: mypy
 
   - repo: https://github.com/Riverside-Healthcare/djLint
-    rev: v1.32.1
+    rev: v1.34.1
     hooks:
       - id: djlint-reformat
         pass_filenames: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
         files: \.py$
 
   - repo: https://github.com/ikamensh/flynt/
-    rev: '1.0.0'
+    rev: 1.0.1
     hooks:
         - id: flynt
           exclude: ^build/.*$|^.tox/.*$|^venv/.*$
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -42,20 +42,6 @@ repos:
           - --lines-after-imports=2
           - --profile=black
 
-  - repo: https://github.com/pycqa/pydocstyle
```
michplunkett, line 45: Screenshot 2024-07-29 at 11 53 53 AM

https://github.com/pycqa/pydocstyle

OpenOversight/tests/test_utils.py
```diff
@@ -233,7 +233,7 @@ def test_save_image_to_s3_and_db_returns_existing_image(
     test_png_bytes_io.close = lambda: None
     firstUpload = save_image_to_s3_and_db(test_png_bytes_io, 1, 1)
     secondUpload = save_image_to_s3_and_db(test_png_bytes_io, 1, 1)
-    assert type(secondUpload) == Image
+    assert isinstance(secondUpload, Image)
```
michplunkett, line 236: The only meaningful changes are the use of `isinstance` instead of `type`, which can be validated by this statement.

b-meson:
The PR looks okay but why are we making changes to the `forms.py` in this branch?

michplunkett:
It's for this [rule](https://docs.astral.sh/ruff/rules/type-comparison/), @b-meson.

It looks like something like this could also work instead if you'd prefer:
```python
isinstance(secondUpload, Image)
type(secondUpload) is Image
```

OpenOversight/app/main/forms.py
```diff
@@ -441,7 +441,7 @@ class DateFieldForm(Form):
     time_field = TimeField("Time", validators=[Optional()])
 
     def validate_time_field(self, field):
-        if not type(field.data) == time:
+        if type(field.data) is not time:
```
michplunkett, line 444: This feels like a more apt change.

b-meson (approved):

PR #1119, opened 2024-07-31
Add user profile tests

### LINKED ISSUE #436: Add user profile tests
- [x] Test that the route is a redirect or 404 when user is not logged in
- [x] Test that users can see their own profiles
- [x] Test that users cannot see other users profiles
- [x] Test that acs cannot see other users profiles
- [x] Test that admins can see other users profiles(?)
- ~~Test that acs cannot see the `users/` and `users/1` routes~~

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/436

## Description of Changes
Added tests to validate `/user/` route logic and correct profile logic to match pre-specified tests.

Screenshot 2024-07-31 at 5 27 37 PM

There is not a `/users/` route, so I marked it out.

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -288,7 +288,10 @@ def get_tutorial():
 @login_required
 def profile(username: str):
     if re.search("^[A-Za-z][A-Za-z0-9_.]*$", username):
-        user = User.by_username(username).one()
+        try:
+            user = User.by_username(username).one()
+        except NoResultFound:
+            abort(HTTPStatus.NOT_FOUND)
     else:
         abort(HTTPStatus.NOT_FOUND)
 
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -295,6 +295,9 @@ def profile(username: str):
     else:
         abort(HTTPStatus.NOT_FOUND)
 
+    if current_user.username != user.username and not current_user.is_administrator:
+        abort(HTTPStatus.FORBIDDEN)
+
     pref = db.session.get(User, current_user.id).dept_pref
     department = db.session.get(Department, pref).name if pref else None
     return render_template("profile.html", user=user, department=department)
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -295,7 +295,7 @@ def profile(username: str):
     else:
         abort(HTTPStatus.NOT_FOUND)
 
-    if current_user.username != user.username and not current_user.is_administrator:
+    if current_user.username != user.username:
         abort(HTTPStatus.FORBIDDEN)
 
     pref = db.session.get(User, current_user.id).dept_pref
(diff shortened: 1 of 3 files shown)


OpenOversight/app/main/views.py
```diff
@@ -288,10 +288,16 @@ def get_tutorial():
 @login_required
 def profile(username: str):
     if re.search("^[A-Za-z][A-Za-z0-9_.]*$", username):
-        user = User.by_username(username).one()
+        try:
+            user = User.by_username(username).one()
+        except NoResultFound:
+            abort(HTTPStatus.NOT_FOUND)
```
michplunkett, line 294: Without this error check the route would error out upon entering an invalid `username`.

OpenOversight/app/main/views.py
```diff
@@ -288,10 +288,16 @@ def get_tutorial():
...
     else:
         abort(HTTPStatus.NOT_FOUND)
 
+    if current_user.username != user.username:
+        abort(HTTPStatus.FORBIDDEN)
```
michplunkett: Make sure users can only see their own profiles.

OpenOversight/tests/routes/test_user_api.py
```diff
@@ -25,36 +25,24 @@
...
 @pytest.mark.parametrize("route,methods", routes_methods)
 def test_user_api_login_required(route, methods, client, mockdata):
-    if HTTPMethod.GET in methods:
-        rv = client.get(route)
-        assert rv.status_code == HTTPStatus.FORBIDDEN
-    if HTTPMethod.POST in methods:
-        rv = client.post(route)
-        assert rv.status_code == HTTPStatus.FORBIDDEN
+    rv = client.get(route)
+    assert rv.status_code == HTTPStatus.FORBIDDEN
```
michplunkett: These `if` statements are unnecessary.

OpenOversight/app/main/views.py
sea-kelp: I think users are expected to be able to see other users' profiles since the leaderboard links to user profiles:
https://github.com/lucyparsons/OpenOversight/blob/43f3ac9dbd89cb33aeb957f0523580e733e941e5/OpenOversight/app/templates/leaderboard.html#L29

Admins can also use it to audit user classifications

OpenOversight/app/main/views.py
sea-kelp: Good catch!

OpenOversight/tests/routes/test_user_api.py
sea-kelp: Is there a reason we're not testing POST routes anymore?

michplunkett: Ahhh, I see it now! I'll revert the changes.

OpenOversight/app/main/views.py
michplunkett: Dope. I was running off the previous ticket. I'll update the logic and tests.

sea-kelp (approved):
These look great! Thanks for adding these tests

PR #1120, opened 2024-08-01
Fix database fixture and password unit test

sea-kelp:
## Description of Changes
1. Simplify database fixtures
2. Fix password unit test - This test currently doesn't fail if `user.password` returns a value. Using `pytest.raises` fails the test if AttributeError is not raised.

## Notes for Deployment
None!

N/A

## Tests and Linting

### DIFF
--- a/OpenOversight/tests/conftest.py
+++ b/OpenOversight/tests/conftest.py
@@ -311,15 +311,14 @@ def db(app):
 @pytest.fixture(scope="function")
 def session(db):
     """Creates a new database session for a test."""
-    with db.engine.connect() as connection:
-        with connection.begin() as tx:
-            db.session = scoped_session(sessionmaker(bind=connection))
+    with db.engine.connect() as connection, connection.begin() as tx:
+        db.session = scoped_session(sessionmaker(bind=connection))
 
-            try:
-                yield db.session
-            finally:
-                db.session.remove()
-                tx.rollback()
+        try:
+            yield db.session
+        finally:
+            db.session.remove()
+            tx.rollback()
 
 
 @pytest.fixture
--- a/OpenOversight/tests/conftest.py
+++ b/OpenOversight/tests/conftest.py
@@ -300,13 +300,10 @@ def db(app):
     with app.app_context():
         _db.app = app
         _db.create_all()
-        connection = _db.engine.connect()
-        session = scoped_session(session_factory=sessionmaker(bind=connection))
-        _db.session = session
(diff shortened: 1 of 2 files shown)


OpenOversight/tests/conftest.py
```diff
@@ -300,30 +300,27 @@ def db(app):
...
+    with db.engine.connect() as connection, connection.begin() as tx:
+        db.session = scoped_session(sessionmaker(bind=connection))
+
+        try:
+            yield db.session
+        finally:
+            db.session.remove()
+            # If an error was raised during test, tx is already rolled back
+            if tx.is_active:
+                tx.rollback()
```
sea-kelp, line 323: We were getting the warning `SAWarning: transaction already deassociated from connection` from 2 unit tests which intentionally violated database constraints, causing the transaction to be automatically rolled back when the commit failed. Then, when the fixture was being cleaned up, it was trying to roll back the transaction a second time.

Instead, this change checks whether the transaction is still active and rolls back if needed.

OpenOversight/tests/conftest.py
```diff
@@ -300,30 +300,27 @@ def db(app):
...
         _db.app = app
         _db.create_all()
-        connection = _db.engine.connect()
-        session = scoped_session(session_factory=sessionmaker(bind=connection))
-        _db.session = session
-        add_mockdata(session)
-        session.commit()
-        connection.close()
-        session.remove()
+        with _db.engine.begin() as connection:
```
michplunkett, line 303: Much more succinct. 🤌🏻

OpenOversight/tests/test_models.py
```diff
@@ -173,11 +173,8 @@ def test_salary_repr(mockdata):
 def test_password_not_printed(mockdata):
     """Validate that password fields cannot be directly accessed."""
     user = User(password="bacon")
-    try:
+    with raises(AttributeError):
```
michplunkett, line 176: Mixed feelings on this one, but it's more of a style thing than anything else.

michplunkett (approved):
LGTM!

OpenOversight/tests/test_models.py
sea-kelp: If you change this line to `return "hello world"`, the previous test logic passes while this test now fails
https://github.com/lucyparsons/OpenOversight/blob/43f3ac9dbd89cb33aeb957f0523580e733e941e5/OpenOversight/app/models/database.py#L796

PR #1124, opened 2024-09-27
Add invalid ID tests for `main.views`

michplunkett:
## Description of Changes
There were a handful of routes where we did not validate the initial ID parameters. Additionally, I standardized how object querying was done within the `main.views` file.

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -321,11 +321,6 @@ def officer_profile(officer_id: int):
         officer = Officer.query.filter_by(id=officer_id).one()
     except NoResultFound:
         abort(HTTPStatus.NOT_FOUND)
-    except:  # noqa: E722
-        exception_type, value, full_traceback = sys.exc_info()
-        error_str = " ".join([str(exception_type), str(value), format_exc()])
-        current_app.logger.error(f"Error finding officer: {error_str}")
-        abort(HTTPStatus.NOT_FOUND)
     form.job_title.query = (
         Job.query.filter_by(department_id=officer.department_id)
         .order_by(Job.order.asc())
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -263,9 +263,14 @@ def redirect_sort_images(department_id: int):
 )
 @login_required
 def sort_images(department_id: int):
+    try:
+        department = Department.query.filter_by(id=department_id).one()
+    except NoResultFound:
+        abort(HTTPStatus.NOT_FOUND)
+
     # Select a random unsorted image from the database
     image_query = Image.query.filter_by(contains_cops=None).filter_by(
-        department_id=department_id
+        department_id=department.id
     )
     image = get_random_image(image_query)
 
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -618,9 +618,10 @@ def redirect_display_tag(tag_id: int):
 def display_tag(tag_id: int):
     try:
         tag = db.session.get(Face, tag_id)
-        proper_path = serve_image(tag.original_image.filepath)
     except NoResultFound:
         abort(HTTPStatus.NOT_FOUND)
+
+    proper_path = serve_image(tag.original_image.filepath)
     return render_template(
         "tag.html", face=tag, path=proper_path, jsloads=["js/tag.js"]
     )
@@ -903,6 +904,10 @@ def list_officer(
     current_job=None,
     require_photo: Optional[bool] = None,
 ):
+    department = db.session.get(Department, department_id)
+    if not department:
+        abort(HTTPStatus.NOT_FOUND)
+
     form = BrowseForm()
     form.rank.query = (
         Job.query.filter_by(department_id=department_id, is_sworn_officer=True)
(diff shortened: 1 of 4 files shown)


OpenOversight/app/main/views.py
```diff
@@ -320,6 +325,7 @@ def officer_profile(officer_id: int):
         exception_type, value, full_traceback = sys.exc_info()
         error_str = " ".join([str(exception_type), str(value), format_exc()])
         current_app.logger.error(f"Error finding officer: {error_str}")
+        abort(HTTPStatus.NOT_FOUND)
```
michplunkett: I'm a little unsure of how to test this. Any advice would be appreciated, @sea-kelp @abandoned-prototype.

sea-kelp: I see what you're saying - it doesn't seem like this code is reachable since (as far as I can tell) the only other exception that could be thrown in the try block is [MultipleResultsFound](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one), but officer id is configured as a [primary key](https://github.com/lucyparsons/OpenOversight/blob/8b12b4b954d379b82cd818b7830acd38e5546241/OpenOversight/app/models/database.py#L228). I wonder if we should just remove it.

If you do want to test it, you could patch `Query.one()` with a side effect to raise MultipleResultsFound

michplunkett: Given that we are searching by the primary key, I am comfortable removing this exception section. Thanks for the 👁️ , @sea-kelp!

OpenOversight/app/main/views.py
```diff
@@ -263,9 +263,14 @@ def redirect_sort_images(department_id: int):
 )
 @login_required
 def sort_images(department_id: int):
+    try:
+        department = Department.query.filter_by(id=department_id).one()
+    except NoResultFound:
+        abort(HTTPStatus.NOT_FOUND)
```
michplunkett, line 269: Though it is equivalent logic, but I feel more comfortable using the `id` field of the `Department` object to query the `Image`s downstream.

OpenOversight/app/main/views.py
```diff
@@ -316,10 +321,6 @@ def officer_profile(officer_id: int):
         officer = Officer.query.filter_by(id=officer_id).one()
     except NoResultFound:
         abort(HTTPStatus.NOT_FOUND)
-    except:  # noqa: E722
-        exception_type, value, full_traceback = sys.exc_info()
-        error_str = " ".join([str(exception_type), str(value), format_exc()])
-        current_app.logger.error(f"Error finding officer: {error_str}")
```
michplunkett, line 322: This logic was unreachable.

OpenOversight/app/main/views.py
```diff
@@ -896,6 +898,10 @@ def list_officer(
     current_job=None,
     require_photo: Optional[bool] = None,
 ):
+    department = db.session.get(Department, department_id)
```
michplunkett: If there isn't a valid `Department` there is no need to perform any of the other queries.

OpenOversight/app/main/views.py
```diff
@@ -380,14 +381,15 @@ def redirect_add_assignment(officer_id: int):
 def add_assignment(officer_id: int):
     form = AssignmentForm()
     officer = db.session.get(Officer, officer_id)
+    if not officer:
+        flash("Officer not found")
+        abort(HTTPStatus.NOT_FOUND)
```
michplunkett: Similar logic to: https://github.com/lucyparsons/OpenOversight/pull/1124/files#r1779369649

OpenOversight/app/main/views.py
```diff
@@ -609,11 +611,11 @@ def redirect_display_tag(tag_id: int):
 
 @main.route("/tags/<int:tag_id>")
 def display_tag(tag_id: int):
-    try:
-        tag = db.session.get(Face, tag_id)
-        proper_path = serve_image(tag.original_image.filepath)
-    except NoResultFound:
+    tag = db.session.get(Face, tag_id)
+    if not tag:
         abort(HTTPStatus.NOT_FOUND)
```
michplunkett, line 616: It does not error if you give it an invalid ID, that is validated by the `test_invalid_tag_id_view_tag` test function.

OpenOversight/app/main/views.py
```diff
@@ -379,15 +380,17 @@ def redirect_add_assignment(officer_id: int):
 @ac_or_admin_required
 def add_assignment(officer_id: int):
     form = AssignmentForm()
-    officer = db.session.get(Officer, officer_id)
+    try:
```
michplunkett, line 383: Made object queries more uniform to the rest of the application.

OpenOversight/app/main/views.py
abandoned-prototype: I am assuming this will lead to one additional query so not exactly equivalent, but makes sense to me

OpenOversight/app/main/views.py
```diff
@@ -263,9 +263,14 @@ def redirect_sort_images(department_id: int):
...
 def sort_images(department_id: int):
+    try:
+        department = Department.query.filter_by(id=department_id).one()
+    except NoResultFound:
+        abort(HTTPStatus.NOT_FOUND)
+
     # Select a random unsorted image from the database
     image_query = Image.query.filter_by(contains_cops=None).filter_by(
-        department_id=department_id
+        department_id=department.id
```
abandoned-prototype, line 273: I would either not change this line, or also update the `department_id` on line 282

abandoned-prototype (approved):
Looks good. Thanks!

OpenOversight/app/main/views.py
michplunkett: True, one more additional query. I think there's more value to the sureness around it than the minor performance hit.

OpenOversight/app/main/views.py
michplunkett: I'll update line 282, good 👁️!

PR #1131, opened 2024-10-23
Add `allegations` table

### LINKED ISSUE #983: Allegations table addition for Incidents
## What issue are you seeing?
Add `allegations` for `incidents`.

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/983
This PR is the first step for the ticket.

## Description of Changes
Adding allegations table.

## Tests and Linting


OpenOversight/app/models/database.py
```diff
@@ -705,6 +705,28 @@ class Incident(BaseModel, TrackUpdates):
     )
 
 
+class Allegation(BaseModel, TrackUpdates):
```
michplunkett, line 857: For some reason, I cannot get alembic to generate a migration for this table addition. Is there anything that looks out of place, @sea-kelp?

sea-kelp: I tried a number of things with the LPL OO alembic but couldn't get it to generate the revision. I was however able to cherry pick the commits (613489b9f14c9d8e8e79430ae5a04a939e060793..f2977e4575a3edd7663467ec9a692d5ccb4a8c2c) into our OpenOversight repo and generate the revision there. I'm wondering if there's maybe a change that we haven't ported over yet that's causing the issue?
https://github.com/OrcaCollective/OpenOversight/commit/cadea7114e4c723d411dc3fbd7e91df708cc5652

Obviously, our models might be slightly different from yours so I wouldn't recommend directly using this generated migration

sea-kelp (commented):
At a higher-level, I think we were discussing having multiple incident-types (lawsuits, internal investigations), each of which would have different types of additional fields.

I was thinking that allegations would belong to an internal investigation-type incident but wouldn't really apply to lawsuits, though I could see an argument for lawsuit causes of action as allegations.

OpenOversight/app/models/database.py
michplunkett: I remember seeing some comments that each individual models needs to be imported into Alembic's `env.py` file. I'll take a look at the Orca one and see what the differences are.

michplunkett: It's super hacky, but I added the table without it creating a migration and then removed the table and it created a migration. I'm just going to reverse the code on upgrade and downgrade, and take that result. I'm not super happy with how it ended up working, but I'll take it at this point.

michplunkett: If I downgrade and then upgrade the flask db command, it will error since the table will already be created when the application starts out.

I then continually used the flask db downgrade command and we get errors for other downgrades as well. I think this means we are consistent, but there are certainly some difficulties between the ORM and flask to maybe look into at some point.

michplunkett:
Would you prefer to have a quick video conversation about this at some point, @sea-kelp?

OpenOversight/app/models/database.py
michplunkett: It looks like this PR may take care of the issue: https://github.com/lucyparsons/OpenOversight/pull/1152

michplunkett:
I think this PR may have gotten lost in discussion purgatory. We can revive it if need be.

Issue #385, opened 2018-02-17
Build API


magpiewings:
Several groups and individuals have already asked me for our data, we should have an API so this is easy!

abandoned-prototype:
To make this a little more concrete, we should implement GET endpoints for the following:

- list departments (`/api/v1/departments`)
- list officers of department + filter (`/api/v1/officers` or `/api/v1/departments/<department_id>/officers` not sure which is better)
- officer details (`/api/v1/officers/<id>` or `/api/v1/departments/<department_id>/officers/<officer_id>`)
- incident detail view (`/api/v1/incidents/<id>`) (maybe also list for department)

Not exactly sure which fields we should return, it probably makes sense to start with whatever we offer in the UI.
Ideally we document the API with something like [swagger](https://editor.swagger.io/).

PR #1138, opened 2024-12-18
Add API endpoints for departments

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/385

## Description of Changes
Add the baseline features for a data-focused API.

## Tests and Linting


OpenOversight/app/api/v1/api.py
```diff
@@ -0,0 +1,36 @@
...
+    put_database_cache_entry,
+)
+from OpenOversight.app.utils.constants import KEY_DEPT_ALL_OFFICERS
+
+
+v1 = Blueprint("v1", __name__, url_prefix="/v1")
+
+
+@v1.route("/departments/<int:department_id>/officers", methods=[HTTPMethod.GET])
+def download_dept_officers(department_id: int):
```
michplunkett: Using one of the formats here: https://github.com/lucyparsons/OpenOversight/issues/385#issuecomment-1606526071

Thoughts, @sea-kelp @abandoned-prototype @b-meson @AetherUnbound?

OpenOversight/app/main/__init__.py
```diff
@@ -1,6 +0,0 @@
-from flask import Blueprint
-
-
-main = Blueprint("main", __name__)
-
-from OpenOversight.app.main import views  # noqa: E402,F401
```
michplunkett, line 6: Moved this to `./main/views.py`.

OpenOversight/app/auth/__init__.py
```diff
@@ -1,6 +0,0 @@
-from flask import Blueprint
-
-
-auth = Blueprint("auth", __name__)
-
-from . import views  # noqa: E402,F401
```
michplunkett, line 6: Moved this to `./auth/views.py`.

OpenOversight/tests/test_models.py
```diff
@@ -158,7 +163,7 @@ def test_face_repr(mockdata):
 
 def test_unit_repr(mockdata):
     unit = Unit.query.first()
-    assert repr(unit) == f"Unit: {unit.description}"
+    assert repr(unit) == f"<Unit: {unit.description}>"
```
michplunkett, line 167: Standardized repr functions.

OpenOversight/app/main/views.py
```diff
@@ -1676,7 +1681,17 @@ def download_incidents_csv(department_id: int):
     cache_params = (Department(id=department_id), KEY_DEPT_ALL_INCIDENTS)
     incidents = get_database_cache_entry(*cache_params)
     if incidents is None:
-        incidents = Incident.query.filter_by(department_id=department_id).all()
+        incidents = (
```
michplunkett, line 1684: Addresses failed test due to lazy loading issue.

OpenOversight/app/models/database.py
```diff
@@ -136,23 +175,14 @@ class Department(BaseModel, TrackUpdates):
     __table_args__ = (UniqueConstraint("name", "state", name="departments_name_state"),)
 
     def __repr__(self):
-        return f"<Department ID {self.id}: {self.name} {self.state}>"
-
-    def to_custom_dict(self):
```
michplunkett, line 141: This isn't needed anymore with the addition of the `Serializable` class.

OpenOversight/app/api/v1.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if notes is None:
+        notes = (
+            db.session.query(Description)
+            .join(Description.officer)
+            .filter(Officer.department_id == department_id)
+            .options(contains_eager(Description.officer))
+            .all()
+        )
+        put_database_cache_entry(*cache_params, notes)
```
michplunkett, line 145: It might be worth moving these to static functions on each of the objects since they are used both here and in views.

michplunkett:
Things that might be needed, are documentation for these endpoints. I'm not sure what a good protocol is for that rn, though I'm welcome to suggestions.

OpenOversight/tests/routes/test_api_v1.py
```diff
@@ -0,0 +1,77 @@
...
+            .options(contains_eager(Assignment.base_officer))
+            .options(joinedload(Assignment.unit))
+            .options(joinedload(Assignment.job))
+            .all()
+        )
+
+        resp_assignments = client.get(
+            url_for("v1.get_dept_assignments", department_id=department_id),
+            follow_redirects=True,
+        )
```
michplunkett, line 41: I could probably get rid of `follow_redirects`, but I think it might be worth keeping in case we modify the endpoint at some point.

OpenOversight/app/models/database.py
```diff
@@ -706,7 +818,9 @@ class Incident(BaseModel, TrackUpdates):
         db.Integer, db.ForeignKey("locations.id", name="incidents_address_id_fkey")
     )
     address = db.relationship(
-        "Location", backref=db.backref("incidents", cascade_backrefs=False)
+        "Location",
+        backref=db.backref("incidents", cascade_backrefs=False),
+        lazy="joined",
```
michplunkett: Addresses this periodic error: `sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Incident at 0xffff731ba510> is not bound to a Session; lazy load operation of attribute 'address' cannot proceed (Background on this error at: h...`

OpenOversight/app/models/database.py
```diff
@@ -264,10 +377,10 @@ class Note(BaseModel, TrackUpdates):
 class Description(BaseModel, TrackUpdates):
     __tablename__ = "descriptions"
 
-    officer = db.relationship("Officer", back_populates="descriptions")
     id = db.Column(db.Integer, primary_key=True)
     text_contents = db.Column(db.Text())
     officer_id = db.Column(db.Integer, db.ForeignKey("officers.id", ondelete="CASCADE"))
+    officer = db.relationship("Officer", back_populates="descriptions")
```
michplunkett, line 241: Ordering for ease of reading.

OpenOversight/app/models/database.py
```diff
@@ -199,11 +216,109 @@ class Department(BaseModel, TrackUpdates):
     __table_args__ = (UniqueConstraint("name", "state", name="departments_name_state"),)
 
     @property
-    def display_name(self):
+    def display_name(self) -> str:
         return self.name if not self.state else f"[{self.state}] {self.name}"
 
+    @staticmethod
+    def get_assignments(department_id: int) -> Query:
```
abandoned-prototype: The return type for all of these is `list`

abandoned-prototype (commented):
Thank you, this is great work. I like the sharing of methods between csv download and API, that's great.  

However, I do think that json output does not have the same limitations that csv has and we should make use of that. What I have in mind here is a response to `/departments/<id>/officers` that looks something like this:

```
[
  {
    "birth_year": 1967,
    "department_id": 3,
    "employment_date": "1993-04-12",
    "first_name": "MICHAEL",
    "gender": "M",
    "id": 32892,
    "last_name": "TYMOCH",
    "middle_initial": "K",
    "race": "WHITE",
    "suffix": null,
    "unique_internal_identifier": null,
    "assignments": [
      {
        "id": 15001,
        "job_id": 20,
        "officer_id": 32892,
        "resign_date": null,
        "star_no": "",
        "start_date": null,
        "unit_id": null
      },
      {
        "id": 88142,
        "job_id": 20,
        "officer_id": 32892,
        "resign_date": null,
        "star_no": "P3662",
        "start_date": null,
        "unit_id": 15
      },
      {
        "id": 88143,
        "job_id": 20,
        "officer_id": 32892,
        "resign_date": null,
        "star_no": "P2107",
        "start_date": null,
        "unit_id": 14
      }
    ],
    "salaries": [
      {
        "id": 15001,
        "is_fiscal_year": true,
        "officer_id": 32892,
        "overtime_pay": null,
        "salary": 63342,
        "year": 2020
      },
      {
        "id": 15002,
        "is_fiscal_year": true,
        "officer_id": 32892,
        "overtime_pay": null,
        "salary": 60042,
        "year": 2020
      },
      {
        "id": 15003,
        "is_fiscal_year": true,
        "officer_id": 32892,
        "overtime_pay": null,
        "salary": 50141,
        "year": 2020
      }
    ]
  },
  {
    "birth_year": 1995,
    "department_id": 3,
    "employment_date": "2018-06-18",
    "first_name": "BRITTANY",
    "gender": "F",
    "id": 32763,
    "last_name": "SANDS",
    "middle_initial": "E",
    "race": "WHITE",
    "suffix": null,
    "unique_internal_identifier": null,
    "assignments": [],
    "salaries": []
  },
  {
    "birth_year": 1958,
    "department_id": 3,
    "employment_date": "2014-08-25",
    "first_name": "ANDREW",
    "gender": "M",
    "id": 32607,
    "last_name": "NEWMARK",
    "middle_initial": "D",
    "race": "WHITE",
    "suffix": null,
    "unique_internal_identifier": null,
    "assignments": [],
    "salaries": []
  }
]
```

Basically all department data combined and reducing the number of endpoints to `/officers` and `/incidents`. I think that makes working with the data much easier, as otherwise one would have to call multiple different endpoints and then merge all the datasets.

It might make sense to omit null value entries, not sure.

I am also not sure how computationally expensive creating this output would be, so it might be useful (or necessary) to also include pagination to the endpoint output to alleviate that issue.
Similarly eventually we also want to support filters, either of these points could be a follow up.

In terms of documentation, I suggested [swagger](https://swagger.io/) in the original issue. I haven't used it in a while so there might be better solutions out there by now

OpenOversight/app/models/database.py
michplunkett: Updated.

michplunkett:
I think the Swagger point could be addressed now, but I'd like to address the nesting part in another PR. It's already quite large and the nesting part is a non-trivial addition, imo.

OpenOversight/app/models/database.py
```diff
@@ -620,6 +591,211 @@ class Image(BaseModel, TrackUpdates):
 )
 
 
+class Incident(BaseModel, TrackUpdates):
```
michplunkett, line 594: The classes were reordered so that I could accurately label return values for the `Department` class.

OpenOversight/app/models/database.py
```diff
@@ -678,68 +854,9 @@ def validate_state(self, key, state):
...
-        db.Integer, db.ForeignKey("departments.id", name="incidents_department_id_fkey")
-    )
-    department = db.relationship(
-        "Department", backref=db.backref("incidents", cascade_backrefs=False), lazy=True
-    )
-
-
 class User(UserMixin, BaseModel):
     __tablename__ = "users"
+
```
michplunkett, line 859: This spacing is consistent with every other `BaseModel` class.

michplunkett:
Tbh, I think it'd be worth adding the Swagger documentation in another PR and then we can deploy this once that's done. I'd rather make things more succinct than otherwise.

b-meson:
Thanks for your hard work on this @michplunkett ! One concern I had about deploying an API endpoint is having it hammered with requests. Even with a limiter for 5 requests like we have and the caching within the DB, I'm not sure if this sufficient.  However, I also don't want to have to deploy an auth mechanism for the API. 

Everything else looks good.

OpenOversight/app/api/v1.py
```diff
@@ -0,0 +1,56 @@
...
+
+def objs_to_dicts_jsonify(obj_list: Query) -> Response:
+    return jsonify([o.to_dict() for o in obj_list])
+
+
+@v1.route("/departments/<int:department_id>/assignments", methods=[HTTPMethod.GET])
+@limiter.limit("5/minute")
+def get_dept_assignments(department_id: int) -> Response:
+    assignments = Department.get_assignments(department_id)
+    return objs_to_dicts_jsonify(assignments)
```
sea-kelp, line 21: I know we already vend a full copy our department data for CSV exports but I think we may want to consider adding pagination and filtering to reduce the amount of data we vend out per request.

OpenOversight/app/models/database.py
```diff
@@ -620,6 +591,211 @@ class Image(BaseModel, TrackUpdates):
...
+                .join(Assignment.base_officer)
+                .filter(Officer.department_id == department_id)
+                .options(contains_eager(Assignment.base_officer))
+                .options(joinedload(Assignment.unit))
+                .options(joinedload(Assignment.job))
+                .all()
+            )
+            put_database_cache_entry(*cache_params, assignments)
+
+        return assignments
```
sea-kelp, line 692: Something like `Department.by_id(1).get_assignments()` or `Assignments.by_department(Department.by_id(1))` might make more sense from an OOP perspective

OpenOversight/app/api/v1.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@v1.route("/departments/<int:department_id>/descriptions", methods=[HTTPMethod.GET])
+@limiter.limit("5/minute")
+def get_dept_descriptions(department_id: int) -> Response:
+    descriptions = Department.get_descriptions(department_id)
+    return objs_to_dicts_jsonify(descriptions)
```
sea-kelp, line 28: It looks like the Description contains the creator id and creation time (This is also happening with the CSV exports). Do we have any reason to vend those?

sea-kelp (commented):
Thanks for doing this work!

I'm wondering what benefits we see with this API over our current CSV exports if we don't add filtering parameters since the API appears to just vend a full export of our data.

Also, do we want to call this version `v1` already if this is still an early iteration?

OpenOversight/app/api/v1.py
michplunkett: Nah, we shouldn't be putting those out.

michplunkett:
Thanks for all of the comments, everyone! It looks like we're not entirely certain on the requirements for the API or if we even want an API given that we have the CSV functionality. I moved the static data methods to this PR (https://github.com/lucyparsons/OpenOversight/pull/1150) and will close this one until we decide how we'd like to move forward.

PR #1139, opened 2024-12-18
Remove `Flask-Sitemap` package

michplunkett:
## Description of Changes
AFAIK, we do not seem to be using Sitemap within the application to any meaningful extent. There is no mention of it in the documentation, and it seems to only be loosely used throughout the application. If we are looking to flesh out the API further, I think it's worth removing it and potentially adding it back in a more explicit manner so that it doesn't seem like it's just legacy code.

Documentation: [link](https://flask-sitemap.readthedocs.io/en/latest/).

Screenshot 2024-12-18 at 5 34 22 PM

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -24,7 +24,7 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import contains_eager, joinedload, selectinload
 from sqlalchemy.orm.exc import NoResultFound
 
-from OpenOversight.app import limiter, sitemap
+from OpenOversight.app import limiter
 from OpenOversight.app.auth.forms import LoginForm
 from OpenOversight.app.main import main
 from OpenOversight.app.main.downloads import (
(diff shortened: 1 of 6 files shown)


OpenOversight/app/utils/constants.py
```diff
@@ -44,7 +41,6 @@
 ENCODING_UTF_8 = "utf-8"
 FILE_TYPE_HTML = "html"
 FILE_TYPE_PLAIN = "plain"
-SAVED_UMASK = os.umask(0o077)  # Ensure the file is read/write by the creator only
```
michplunkett, line 47: This constant is unused.

OpenOversight/app/main/views.py
```diff
@@ -124,23 +123,6 @@
 )
 
 
-# Ensure the file is read/write by the creator only
-SAVED_UMASK = os.umask(0o077)
```
michplunkett, line 128: This constant is unused.

michplunkett:
Sitemap is failing due to the large amount of data we have. It's good to keep it, but we may need to adjust how it's used to actually make it work. Thanks for the extra info, @sea-kelp!

PR #1140, opened 2024-12-20
Standardize model `__repr__` functions

michplunkett:
## Description of Changes
Addressed inconsistencies with the model `__repr__` functions.

Mostly making this because this PR got a bit out of control: https://github.com/lucyparsons/OpenOversight/pull/1138

## Tests and Linting

### DIFF
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -406,7 +406,7 @@ class Assignment(BaseModel, TrackUpdates):
     resign_date = db.Column(db.Date, index=True, unique=False, nullable=True)
 
     def __repr__(self):
-        return f"<Assignment: ID {self.officer_id} : {self.star_no}>"
+        return f"<Assignment ID: {self.officer_id} : {self.star_no}>"
 
     @property
     def start_date_or_min(self):
@@ -656,12 +656,12 @@ class Link(BaseModel, TrackUpdates):
     author = db.Column(db.String(255), nullable=True)
     has_content_warning = db.Column(db.Boolean, nullable=False, default=False)
 
-    def __repr__(self):
-        return f"<Link ID: {self.id} : {self.title}>"
-
     @validates("url")
     def validate_url(self, key, url):
         return url_validator(url)
+
+    def __repr__(self):
+        return f"<Link ID: {self.id} : {self.title}>"
 
 
 class Incident(BaseModel, TrackUpdates):
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -212,6 +212,9 @@ class Note(BaseModel, TrackUpdates):
     officer_id = db.Column(db.Integer, db.ForeignKey("officers.id", ondelete="CASCADE"))
     officer = db.relationship("Officer", back_populates="notes")
 
+    def __repr__(self):
+        return f"<Note ID: {self.id} : {self.text_contents}>"
+
 
 class Description(BaseModel, TrackUpdates):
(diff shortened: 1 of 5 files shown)


OpenOversight/app/models/database.py
```diff
@@ -279,6 +285,14 @@ class Officer(BaseModel, TrackUpdates):
         CheckConstraint("gender in ('M', 'F', 'Other')", name="gender_options"),
     )
 
+    def __repr__(self):
+        if self.unique_internal_identifier:
+            return (
+                f"<Officer ID {self.id}: {self.full_name()} "
```
AetherUnbound: Shouldn't this be...?
```suggestion
                f"<Officer ID: {self.id} : {self.full_name()} "
```

OpenOversight/app/models/database.py
```diff
@@ -279,6 +285,14 @@ class Officer(BaseModel, TrackUpdates):
...
+                f"({self.unique_internal_identifier})>"
+            )
+        return f"<Officer ID {self.id}: {self.full_name()}>"
```
AetherUnbound: (Same as above):
```suggestion
        return f"<Officer ID: {self.id} : {self.full_name()}>"
```

OpenOversight/tests/test_models.py
```diff
@@ -137,28 +139,33 @@ def test_assignment_repr(mockdata):
...
 
 
+def test_incident_repr(mockdata):
+    incident = Incident.query.first()
+    assert repr(incident) == f"<Incident ID: {incident.id} : {incident.report_number}>"
+
+
 def test_job_repr(mockdata):
     job = Job.query.first()
     assert repr(job) == f"<Job ID {job.id}: {job.job_title}>"
```
AetherUnbound: Should the format for Job ID also change? Since this looks more like the other format.

AetherUnbound (commented):
Looks good but there are a few more inconsistencies that might be good to address I think!

michplunkett:
Thanks for the eye, @AetherUnbound! I'll address those tomorrow and get back to you when they're done.

b-meson:
I don't know enough about Python to feel super comfortable reviewing it so I'll let the rest of the reviewers weigh in!

sea-kelp (approved):
Other than the inconsistencies Madison pointed out, this looks good! If we're being pedantic, the Python spec technically says you should be able to instantiate the object by eval-ing the string returned by  `__repr__`
https://docs.python.org/3/library/functions.html#repr

I also wonder if we could get these for free by just making our model classes dataclasses: https://docs.sqlalchemy.org/en/20/orm/dataclasses.html

michplunkett:
These are all good comments, thank you!

Please excuse the delay in my response, I've been holday'ing. I'll take a look at the `MappedAsDataclass` class tomorrow. If we could remove the need to update the `__repr__` function and always be up to spec, that feels like the right way to go. I'll ping y'all when I have an update in place. 🧐

OpenOversight/tests/test_models.py
michplunkett: Done!

OpenOversight/app/models/database.py

OpenOversight/app/models/database.py

michplunkett:
Looking at the changes needed to implement the `MappedAsDataclass` class, @sea-kelp, I am going to make these updates using two PRs:
1. Update with the changes that @AetherUnbound pointed out.
2. Update SQL Alchemy from `1.4.52` -> `2.x.y` and then add the `MappedAsDataclass` addition.

PR #1150, opened 2025-01-30
Add static `Department` functions

michplunkett:
## Description of Changes
Moved `Department`-related methods to the `Department` object.

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/downloads.py
+++ b/OpenOversight/app/main/downloads.py
@@ -5,7 +5,6 @@ from http import HTTPStatus
 from typing import Any, Callable, Dict, List, TypeVar
 
 from flask import Response, abort
-from sqlalchemy.orm import Query
 
 from OpenOversight.app.models.database import (
     Assignment,
@@ -40,7 +39,7 @@ def check_output(output_str):
 
 
 def make_downloadable_csv(
-    query: Query,
+    query: List,
     department_id: int,
     csv_suffix: str,
     field_names: List[str],
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -14,19 +14,34 @@ from flask_login import UserMixin
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import CheckConstraint, UniqueConstraint, func
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import DeclarativeMeta, declarative_mixin, declared_attr, validates
+from sqlalchemy.orm import (
+    DeclarativeMeta,
+    contains_eager,
+    declarative_mixin,
+    declared_attr,
+    joinedload,
+    validates,
+)
 from sqlalchemy.sql import func as sql_func
 from werkzeug.security import check_password_hash, generate_password_hash
 
 from OpenOversight.app.models.database_cache import (
     DB_CACHE,
+    get_database_cache_entry,
     model_cache_key,
+    put_database_cache_entry,
     remove_database_cache_entries,
 )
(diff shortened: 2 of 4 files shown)


OpenOversight/app/main/downloads.py
```diff
@@ -40,7 +39,7 @@ def check_output(output_str):
 
 
 def make_downloadable_csv(
-    query: Query,
+    query: List,
```
michplunkett, line 42: Corrected typing for this function.

OpenOversight/app/models/database.py
```diff
@@ -620,6 +591,211 @@ class Image(BaseModel, TrackUpdates):
...
+        db.String(100), unique=False, nullable=True
+    )
+
+    __table_args__ = (UniqueConstraint("name", "state", name="departments_name_state"),)
+
+    @property
+    def display_name(self) -> str:
+        return self.name if not self.state else f"[{self.state}] {self.name}"
+
+    @staticmethod
```
michplunkett: Makes it so any other queries on these objects have their cacheing already handled.

OpenOversight/app/models/database.py
```diff
@@ -620,6 +591,211 @@ class Image(BaseModel, TrackUpdates):
...
+    report_number = db.Column(db.String(50), index=True)
+    description = db.Column(db.Text(), nullable=True)
+    address_id = db.Column(
+        db.Integer, db.ForeignKey("locations.id", name="incidents_address_id_fkey")
+    )
+    address = db.relationship(
+        "Location",
+        backref=db.backref("incidents", cascade_backrefs=False),
+        lazy="joined",
+    )
```
michplunkett: Addresses this periodic error: `sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Incident at 0xffff731ba510> is not bound to a Session; lazy load operation of attribute 'address' cannot proceed (Background on this error at: h...`

OpenOversight/tests/test_models.py
```diff
@@ -20,7 +20,14 @@
     User,
 )
 from OpenOversight.app.utils.choices import STATE_CHOICES
-from OpenOversight.tests.conftest import SPRINGFIELD_PD
+from OpenOversight.tests.conftest import AC_DEPT, SPRINGFIELD_PD
+
+
+def test_no_omitted_fields_in_to_dict(mockdata):
```
michplunkett, line 26: Added test to make sure excluded fields were not in the dictionary methods.

b-meson (approved):
This looks great to me but it would be good to have another set of eyes on it before we merge.

abandoned-prototype (approved):
This is some nice refactoring, making some follow up work easier. Thanks!

Is there a specific reason you changed the order of the classes in database.py? Makes it a little harder to see the changes, but I do support changing the order if it serves a purpose.

The one thing I am a little unsure about is if these static functions should maybe live in their own file, possibly in the same directory. They could still be called from the Department object for convenience. So what I mean would be something like
```
class Department(BaseModel, TrackUpdates):
[...]
    @staticmethod
    def get_incidents(department_id: int) -> List[Incident]:
        return department_helper.get_incidents(department_id)
```
I think that would help keeping the `database.py` file smaller and focused on what has to be there.

Just a thought, I am also okay with leaving it as is

michplunkett:
The orders of the objects were switched so that I could do accurate `List[Object]` designations. If you feel like it creates a problem, I could also switch them all over to `List` and leave them at that. Would that be better, @abandoned-prototype?

michplunkett:
The objects have been put back in their original order, and I'll leave the static methods as is. I have no qualms with them being moved at some other point though.

PR #1151, opened 2025-02-04
Remove audit fields from `Description` download

michplunkett:
## Description of Changes
Removed audit fields from the `Description` download function. They leak user information and provide data that isn't particularly useful outside of internal auditing.

## Tests and Linting

### DIFF
--- a/OpenOversight/app/main/downloads.py
+++ b/OpenOversight/app/main/downloads.py
@@ -160,8 +160,5 @@ def descriptions_record_maker(description: Description) -> _Record:
     return {
         "id": description.id,
         "text_contents": description.text_contents,
-        "created_by": description.created_by,
         "officer_id": description.officer_id,
-        "created_at": description.created_at,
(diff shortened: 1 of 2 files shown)


OpenOversight/app/main/downloads.py
```diff
@@ -160,8 +160,5 @@ def descriptions_record_maker(description: Description) -> _Record:
     return {
         "id": description.id,
         "text_contents": description.text_contents,
-        "created_by": description.created_by,
         "officer_id": description.officer_id,
-        "created_at": description.created_at,
-        "last_updated_at": description.last_updated_at,
```
michplunkett, line 166: Unless there's something I'm missing, these fields shouldn't be needed. They aren't included in any other `_record_maker` functions, so my guess is that we do not need to include them here.

abandoned-prototype (approved):
Yes, I agree with not adding these fields to the csv download. Thanks for making this change!

PR #1000, opened 2023-07-31
Delete `create_db.py`

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/675

## Description of Changes
Removed `create_db.py` file, moved the logic to the Alembic file, and created constants for the code around my changes.

## Notes for Deployment
If it works in development, it SHOULD work in production. 

## Tests and linting
 - [x] This branch is up-to-date with the `develop` branch.
 - [x] `pytest` passes on my local development environment.
 - [x] `pre-commit` passes on my local development environment.
 - [x] Ran the `make` default command and there were not any issues.

<details><summary>make dev output</summary>

```console
% make dev
docker-compose build
...
WARN[0000] The "APPROVE_REGISTRATIONS" variable is not set. Defaulting to a blank string. 
[+] Running 2/0
 ✔ Container openoversight-postgres-1  Running                                                                                                                                                             0.0s 
 ✔ Container openoversight-web-1       Running                                                                                                                                                             0.0s 
Postgres is up
## Creating database
docker-compose run --rm web alembic --config=./OpenOversight/migrations/alembic.ini stamp head
...
Postgres is up
## Populate database with test data
docker-compose run --rm web python ./test_data.py -p
...
2023-08-01 21:43:54,796 INFO sqlalchemy.engine.Engine [generated in 0.00006s] {'pk_1': 1}
[*] Completed successfully!
% 
```
</details>

PR #1152, opened 2025-02-10
Revert "Delete `create_db.py` (#1000)"

abandoned-prototype:
## Fixes issue

#1000 unfortunately broke alembic's functionality to create migrations that include new tables (Using `flask db migrate`). The issue is that before alembic has a change to analyze the difference between existing database and schema from database.py the `db.create_all()` command automatically creates the tables making it so that they are not registered for the migration.

I could not come up with a way to reasonably work around this issue, so I think the best course of action is to revert the removal of the create_db.py file.

## Description of Changes

This reverts that breaking commit. It is not a clean revert of #1000, since that PR also included unrelated changes to tests and constants. Those changes were not reverted.

## Notes for Deployment

I have verified the changes in my local environment and the tests are passing, but I have not tested that the changes work as expected in a completely fresh environment (which would use the create_db file)

## Tests and Linting
- [ ] (N/A) Ran `make create_db_diagram` command.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
         description: Only executable files should have shebangs (e.g. '#!/usr/bin/env python')
         entry: "#!/"
         pass_filenames: true
-        exclude: bin|cli|manage.py|app.py|setup.py|docs|test_data.py
+        exclude: bin|cli|docs|test_data.py
         files: \.py$
 
   - repo: https://github.com/ikamensh/flynt/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
         description: Only executable files should have shebangs (e.g. '#!/usr/bin/env python')
         entry: "#!/"
         pass_filenames: true
-        exclude: bin|cli|manage.py|app.py|setup.py|docs|create_db.py|test_data.py
+        exclude: bin|cli|manage.py|app.py|setup.py|docs|test_data.py
         files: \.py$
 
   - repo: https://github.com/ikamensh/flynt/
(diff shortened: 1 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -52,7 +52,7 @@ repos:
         description: Only executable files should have shebangs (e.g. '#!/usr/bin/env python')
         entry: "#!/"
         pass_filenames: true
-        exclude: bin|cli|manage.py|app.py|setup.py|docs|test_data.py
+        exclude: bin|cli|manage.py|app.py|setup.py|docs|create_db.py|test_data.py
```
michplunkett: We can leave `create_db` off of this list, yeah?

OpenOversight/migrations/env.py
```diff
@@ -19,6 +14,12 @@
 logger = logging.getLogger("alembic.env")
 
 
+config.set_main_option(
+    "sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI")
```
michplunkett: Could you change this to: `"sqlalchemy.url", current_app.config.get(KEY_DATABASE_URI)`?

create_db.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python
+from OpenOversight.app import create_app
+from OpenOversight.app.models.database import db
+
+
+app = create_app("development")
```
michplunkett: Can be `KEY_ENV_DEV`.

create_db.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python
```
michplunkett: We don't need the shebang, right?

michplunkett (approved):
If you could make those small shebang and constant related changes, it looks good to go!

.pre-commit-config.yaml
```diff
@@ -52,7 +52,7 @@ repos:
         description: Only executable files should have shebangs (e.g. '#!/usr/bin/env python')
         entry: "#!/"
         pass_filenames: true
-        exclude: bin|cli|manage.py|app.py|setup.py|docs|test_data.py
```
michplunkett, line 55: `manage.py|app.py|setup.py` don't exist in this repository.

PR #1159, opened 2025-02-24
Default to "N/A" if dept state is invalid

sea-kelp:
## Description of Changes
Default to "N/A" if department state is invalid

In [the migration](https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/migrations/versions/2023-07-26-1551_18f43ac4622f_add_state_column_to_departments.py) where we introduced the department state column, we defaulted to empty string for existing departments. We think this may have broken the browse page when trying to load the browse page.

## Notes for Deployment
None!

N/A

## Tests and Linting
Reproduced the internal server error by setting the department state to empty string
```
openoversight-dev=# update departments set state='' where id = 1;                                                                                                                       
```

Checked that the error was no longer displayed once this change was implemented

- [ ] Ran `make create_db_diagram` command.

### DIFF
--- a/OpenOversight/app/filters.py
+++ b/OpenOversight/app/filters.py
@@ -101,6 +101,10 @@ def display_currency(value: float) -> str:
 def get_state_full_name(abbrev: str) -> str:
     if abbrev == "FA":
         return "Federal"
+
+    state = us.states.lookup(abbrev)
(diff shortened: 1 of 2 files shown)


michplunkett (approved):
Noice.

PR #1161, opened 2025-02-27
Fix `make_admin_user` command

### LINKED ISSUE #1160: Fix `make_admin_user` command
The `make_admin_user` command still uses `User.confirmed` which has been removed instead of `User.confirmed_at` and `User.confirmed_by`. 
https://github.com/lucyparsons/OpenOversight/blob/a57b820954d0dcb0c03df6a9399daa30cdc5ee63/OpenOversight/app/commands.py#L64-L72

I fixed this here in our instance:
https://github.com/OrcaCollective/OpenOversight/pull/534/commits/8de5ce595a5f9f69d166f2836d6a7a90ffa382ad#diff-b71efd4759035c21ea60aeab2433d4956db9c176e18e1c6fd8833316edec315d

michplunkett:
## Fixes issue
https://github.com/lucyparsons/OpenOversight/issues/1160

## Description of Changes
Changed `confirmed` field to `confirmed_at` and `confirmed_by` and added a validating test.

## Tests and Linting


OpenOversight/app/commands.py
```diff
@@ -61,11 +61,14 @@ def make_admin_user():
             break
         print("Passwords did not match")
 
+    admin_user = User.query.filter_by(is_administrator=True).first()
```
michplunkett: This does presume that there is at least one admin user in the system.

sea-kelp: If we're using `confirmed_by` for auditing, this technically wouldn't be the user that this new user was confirmed by. Since we don't have the concept of a system user and this field isn't nullable, I set confirmed_by to the user's own id in my change. What do you think?

michplunkett: I think that's a better solution, I'm here for it. I'll make the change tomorrow.

michplunkett: Done!

OpenOversight/app/commands.py
```diff
@@ -65,11 +65,16 @@ def make_admin_user():
         username=username,
         email=email,
         password=password,
-        confirmed=True,
         is_administrator=True,
     )
     db.session.add(u)
     db.session.commit()
+
+    user = User.by_username(username).one_or_none()
```
sea-kelp: If you use `db.session.flush()` here in place of `db.session.commit()`,  I don't think you'llt need to look up the user again

sea-kelp (approved):
Looks good!

OpenOversight/app/commands.py
michplunkett: I'll give it a shot!

OpenOversight/tests/test_functional.py
```diff
@@ -356,6 +356,7 @@ def test_image_classification_and_tagging(browser, server_port, session):
     browser.find_element(By.ID, "last_name").send_keys("Friendly")
     browser.find_element(By.ID, "star_no").send_keys(star_no)
 
+    wait_for_element(browser, By.ID, "submit")
```
michplunkett, line 359: Takes care of a race condition that was causing this test to fail.

Issue #1148, opened 2025-01-16
Support cross-agency identifier field such as for the California POST id


nn35007:
It would be great if OpenOversight would support an additional identifier that allows tracking the same cop across multiple agencies, such as the California POST id. Maybe the identifier should be scoped to a particular state.

The California POST Commission is responsible for tracking certifications across most (but not all) law enforcement agencies in California. It issues a unique identifier, the POST id, to officers and tracks which agency an officer works for and their rank. This POST id is stable across rank changes and even POST member agencies, so if e.g. an officer transfers from LAPD to San Diego PD, or if they transition between full-time and reserve officer, their POST id remains the same. POST also is responsible for certifying and decertifying cops. Over 250,000 California officers across almost 800 agencies have a unique POST id.

California POST ids are strings, apparently of the format `[A-Z][0-9]{2}-[A-Z][0-9]{2}`, so for example B09-M46.

It would be great if users could query by this unique POST id across agencies, so e.g. list all entries for a certain POST id across all agencies (maybe of a certain state).

Many other states have POST agencies or equivalent bodies, but I am not familiar with any unique identifiers they may have or publish. It's not clear to me what identifiers the federal data bases (National Law Enforcement Accountability Database, National Law Enforcement Accountability Database) use, but their data isn't publicly accessible either way.

Ideally this identifier could be used to query the same officer across agencies — so e.g. "query all California POST agencies for the officer with California POST id B09-M46".

I have not dug in enough into the OpenOversight database model ([this file](https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/models/database.py), right?) to make an intelligent suggestion for how to implement this :-)

michplunkett:
@abandoned-prototype and Matt should have reached out to address this.

abandoned-prototype:
I think this is a great suggestion and I am currently working on implementing this.
Right now I am thinking of adding a table for each type of identifier and than have a many to many relationship between that and the officer table via a join table which also holds the identifier value (the actual ID).
A little more work than I originally thought with adding it to officer creation and edit as well as csv uploads probably. Also, I don't think right now we have a cross department search, so this would have to be added too

PR #1164, opened 2025-03-07
Overhaul the `/departments/<department_id>` view

abandoned-prototype:
## Description of Changes

I ran into this when working on #1148

The BrowseForm was basically only used as a data-container, this change is making sure we also use it to filter the incoming data to simplify the code in the view file. I also included other improvements like reducing the number of SQL queries sent.

Functionality wasn't changed (at least that was the goal), so the user experience should remain the same.

## Notes for Deployment
N/A

## Tests and Linting
- [ ] Ran `make create_db_diagram` command.

### DIFF
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -918,16 +918,19 @@ def list_officer(
         error_out=False,
     )
 
-    img_id_to_officer = {}
+    img_id_to_officers: dict[int, list[Officer]] = {}
     for officer in officers.items:
         if officer.face:
             face_image_id = sorted(
                 officer.face, key=lambda x: x.featured, reverse=True
             )[0].img_id
-            img_id_to_officer[face_image_id] = officer
-    images = Image.query.filter(Image.id.in_(img_id_to_officer.keys())).all()
+            if face_image_id not in img_id_to_officers:
+                img_id_to_officers[face_image_id] = []
+            img_id_to_officers[face_image_id].append(officer)
+    images = Image.query.filter(Image.id.in_(img_id_to_officers.keys())).all()
     for image in images:
-        img_id_to_officer[image.id].image = serve_image(image.filepath)
+        for officer in img_id_to_officers[image.id]:
+            officer.image = serve_image(image.filepath)
 
     next_url = url_for_target_null_values_skipped(
         "main.list_officer",
--- a/OpenOversight/app/main/views.py
+++ b/OpenOversight/app/main/views.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from http import HTTPMethod, HTTPStatus
 from traceback import format_exc
 from typing import Optional
+from urllib.parse import urlencode
 
 from flask import (
     Blueprint,
@@ -72,7 +73,6 @@ from OpenOversight.app.models.database import (
     db,
 )
 from OpenOversight.app.utils.auth import ac_or_admin_required, admin_required
-from OpenOversight.app.utils.choices import AGE_CHOICES, GENDER_CHOICES, RACE_CHOICES
 from OpenOversight.app.utils.cloud import crop_image, save_image_to_s3_and_db
 from OpenOversight.app.utils.constants import (
     ENCODING_UTF_8,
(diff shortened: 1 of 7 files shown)


OpenOversight/app/main/views.py
```diff
@@ -866,173 +867,69 @@ def edit_department(department_id: int):
...
-    rank=None,
-    min_age=None,
-    max_age=None,
-    last_name=None,
-    first_name=None,
-    badge=None,
-    unique_internal_identifier=None,
-    unit=None,
-    current_job=None,
-    require_photo: Optional[bool] = None,
```
abandoned-prototype, line 921: flask was never setting these vars, so it is basically just a variable initializer list

OpenOversight/app/main/views.py
```diff
@@ -866,173 +867,69 @@ def edit_department(department_id: int):
...
-            first_name=first_name,
-            badge=badge,
-            unique_internal_identifier=unique_internal_identifier,
-            unit=unit,
-            current_job=current_job,
-            require_photo=require_photo,
-        ),
+        url_for("main.list_officer", department_id=department_id)
+        + "?"
+        + urlencode(request.args.to_dict(flat=False), doseq=True),
```
abandoned-prototype, line 875: This doesn't look the cleanest, but unfortunately flask's `url_for` does not have a keyword to supply the get arguments separately and this works.

OpenOversight/app/utils/general.py
```diff
@@ -159,6 +159,11 @@ def str_is_true(str_) -> bool:
     return bool(strtobool(str_.lower()))
 
 
+def url_for_target_null_values_skipped(target, **kwargs):
```
michplunkett: Could you put types (parameters and returns) on all of these functions?

abandoned-prototype: Done

OpenOversight/app/main/views.py
sea-kelp: This seems to work for me:
```python
url_for("main.list_officer", department_id=department_id, **request.args),
```

Would you mind giving that a try?

michplunkett (approved):

OpenOversight/app/main/views.py
```diff
@@ -866,173 +867,69 @@ def edit_department(department_id: int):
...
-
-    next_url = url_for(
+        if officer.face:
+            face_image_id = sorted(
+                officer.face, key=lambda x: x.featured, reverse=True
+            )[0].img_id
+            img_id_to_officer[face_image_id] = officer
+    images = Image.query.filter(Image.id.in_(img_id_to_officer.keys())).all()
+    for image in images:
+        img_id_to_officer[image.id].image = serve_image(image.filepath)
```
sea-kelp: On my end, I'm seeing that many officers who have images available are showing up as "No Photos" in the browse view. This is happening because multiple officers have the same image assigned to them.

I'm not sure that this happens in practice since we usually use face thumbnails but it might be an edge case want to account for.

OpenOversight/app/templates/list_officer.html
```diff
@@ -91,11 +91,9 @@ <h3 class="panel-title accordion-toggle">Officer Rank, Unit, and Status</h3>
                                 name="rank"
                                 class="select2-multi-search"
                                 multiple="multiple">
-                          {% for rank_key, rank in choices['rank'] %}
```
sea-kelp, line 94: For awareness, a future version of wtforms may break this:
https://github.com/pallets-eco/wtforms/issues/811

OpenOversight/app/main/views.py
abandoned-prototype: This is what I started with, but if someone puts `department_id=` in the url it will cause a server error which I wanted to avoid

OpenOversight/app/main/views.py
abandoned-prototype: Yeah, that shouldn't happen in practice, but I agree with the concern and will make adjustments

OpenOversight/app/templates/list_officer.html
abandoned-prototype: Good call! I will change it so it works for both versions

abandoned-prototype: Made that change, it should now also work in the future versions

OpenOversight/app/main/views.py
abandoned-prototype: Fixed this, based on my tests the issue should be resolved

PR #1166, opened 2025-03-20
Update GitHub Action versions

michplunkett:
## Description of Changes
Increment GitHub Action versions.

## Tests and Linting

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -27,7 +27,7 @@ jobs:
       contents: read
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@4
       - name: Log in to GitHub Docker Registry
         uses: docker/login-action@v2
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_deploy.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       contents: read
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@4
```
michplunkett, line 30: The use of `actions/checkout@4` in the other Actions should validate this change.

OpenOversight/app/utils/general.py
```diff
@@ -101,6 +100,9 @@ def normalize_gender(input_gender: str) -> Union[str, None]:
 
     return normalized_genders.get(input_gender.lower().strip())
 
+def str_to_bool(val: str) -> int:
+    return val.lower() in ("y", "yes", "t", "true", "on", "1")
```
michplunkett, line 105: `distutils` was deprecated in Python 3.10 and removed in Python 3.12. Since we're using Python 3.12 for testing, `distutils.util.strtobool` had to be replaced.

sea-kelp (approved):

PR #1173, opened 2025-04-05
Fix edit user form

### LINKED ISSUE #1172: EditUser form is broken
With [this change](https://github.com/lucyparsons/OpenOversight/commit/00976630a3c8f9898e52444e88a37dbe4e7fd0be#diff-605f590193368d3cdf1ba750d36546d5cbd994cba643db0b0aad4e5f4bde7f08), the edit user form (`/auth/users/:id`) no longer populates the disabled/approved/confirmed checkbox fields and does not allow users to be un-disabled/approved/confirmed.

sea-kelp:
## Fixes issue
Fixes #1172 

## Description of Changes
Fix edit user form

1. Checkboxes to approve/confirm/disable users were not populated correctly
2. Users could not be un-approved/confirmed/disabled once approved/confirmed/disabled

## Notes for Deployment
None!

N/A

## Tests and Linting

### DIFF
--- a/OpenOversight/app/models/database.py
+++ b/OpenOversight/app/models/database.py
@@ -1085,35 +1085,35 @@ class User(UserMixin, BaseModel):
         """Override UserMixin.is_active to prevent disabled users from logging in."""
         return not self.disabled_at
 
-    def approve_user(self, approving_user_id: int):
+    def approve_user(self, approving_user_id: int, approved: bool):
         """Handle approving logic."""
-        if self.approved_at or self.approved_by:
-            return False
+        if approved and not self.approved_by and not self.approved_at:
+            self.approved_at = datetime.now(timezone.utc)
+            self.approved_by = approving_user_id
+            db.session.commit()
+        elif not approved and self.approved_by and self.approved_at:
+            self.approved_at = None
+            self.approved_by = None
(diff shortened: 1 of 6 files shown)


OpenOversight/app/models/database.py
```diff
@@ -1085,35 +1085,35 @@ def is_active(self):
...
-        self.approved_by = approving_user_id
-        db.session.add(self)
-        db.session.commit()
-        return True
-
-    def confirm_user(self, confirming_user_id: int):
+        if approved and not self.approved_by and not self.approved_at:
+            self.approved_at = datetime.now(timezone.utc)
+            self.approved_by = approving_user_id
+            db.session.commit()
```
michplunkett, line 1093: 🔬 Are we likely to run into a scenario where we don't commit at any point? The same goes for the other two functions.

michplunkett (approved):
LGTM, minus that one question!

OpenOversight/app/models/database.py
sea-kelp: It's an expected case: if the user is already approved and the next form submission also has `approved` checked (or not approved and `approved` is not checked), then we don't want to update the existing user value

PR #1174, opened 2025-04-06
Add GitHub Action for schema diagram

michplunkett:
## Description of Changes
Added a GitHub Action so that we don't have to remember to run `make create_db_diagram` and it will let us know if we need to run it.

## Tests and Linting
- [x] Manually tested the `create_db_diagram` command by adding a column to the `users` table and then removing it.


.github/pull_request_template.md
```diff
@@ -26,6 +26,5 @@ dispatcher where…", "Improve our handling of…", etc.
 
 ## Tests and Linting
 - [ ] This branch is up-to-date with the `develop` branch.
-- [ ] Ran `make create_db_diagram` command.
```
michplunkett, line 29: Since we have an Action for this, I don't think we need this line anymore. I could see the benefit of having it more generally tho, so I'm open to putting it back.

requirements.txt
```diff
@@ -19,7 +19,6 @@ dnspython==2.6.1
 docutils~=0.20.1
 dominate~=2.8.0
 email-validator==2.0.0.post2
-Faker==18.13.0
```
michplunkett, line 22: These packages were already being installed in `dev-requirements.txt`.

michplunkett:
Hmmm, may mess with this a bit later. I think I'm gonna need to build a PostgreSQL DB in the GA to get this to work.

michplunkett:
This isn't ready to review yet, plz ignore.

dev-requirements.txt
```diff
@@ -1,6 +1,5 @@
 coveralls==3.3.1
 eralchemy==1.5.0
-faker==18.11.2
```
michplunkett, line 3: This is already included in the requirements file.

requirements.txt
```diff
@@ -58,7 +58,6 @@ requests==2.32.1
 rich~=13.4.2
 s3transfer~=0.6.1
 six~=1.16.0
-sphinx==7.1.2
```
michplunkett, line 61: This is already included in the dev-requirements file.

michplunkett: I'm not sure where I can access the sphinx documentation. If anyone knows where I can, let me know and I'll make sure everything is 💯 before we push.

michplunkett:
Given that @sea-kelp updated the diagram on the last push, I'm curious as to why it had me update it now. Once I updated it, it did work on the GHA. I'm not sure the reason why.

requirements.txt
sea-kelp: I think it should be in the `docs/_build/html` directory
https://github.com/lucyparsons/OpenOversight/blob/784a9decf6962777a25884eaed9b6a73e84b3e96/Makefile#L97-L100

sea-kelp:
Is this an important error message?
```
ERROR:root:either pygraphviz or graphviz should be installed
```
https://github.com/lucyparsons/OpenOversight/actions/runs/14412448640/job/40423165992#step:9:18

database/schema.dot
```diff
@@ -21,70 +21,70 @@
...
...ACE="Helvetica"><TABLE BORDER="0" CELLBORDER="1" CELLPADDING="4" CELLSPACING="0"><TR><TD><B><FONT POINT-SIZE="16">salaries</FONT></B></TD></TR><TR><TD ALIGN="LEFT" PORT="id"><u><FONT>id</FONT></u> <FONT> [INTEGER]</FONT> NOT NULL</TD></TR><TR><TD ALIGN="LEFT" PORT="created_at"><FONT>created_at</FONT> <FONT> [TIMESTAMP]</FONT> NOT NULL</TD></TR><TR><TD ALIGN="LEFT" PORT="created_by"><FONT>created_by</FONT> <FONT> [INTEGER]</FONT></TD></TR><TR><TD ALIGN="LEFT" PORT="is_fiscal_year"><FONT>is_fiscal_year</FONT> <FONT> [BOOLEAN]</FONT> NOT NULL</TD></TR><TR><TD ALIGN="LEFT" PORT="last_updated_at"><FONT>last_updated_at</FONT> <FONT> [TIMESTAMP]</FONT> NOT NULL</TD></TR><TR><TD ALIGN="LEFT" PORT="last_updated_by"><FONT>last_updated_by</FONT> <FONT> [INTEGER]</FONT></TD></TR><TR><TD ALIGN="LEFT" PORT="officer_id"><FONT>officer_id</FONT> <FONT> [INTEGER]</FONT></TD></TR><TR><TD ALIGN="LEFT" PORT="overtime_pay"><FONT>overtime_pay</FONT> <FONT> [DOUBLE_PRECISION]</FONT></TD></TR><TR><TD ALIGN="LEFT" PORT="salary"><FONT>salary</FONT> <FONT> [DOUBLE_PRECISION]</FONT> NOT NULL</TD></TR><TR><TD ALIGN="LEFT" PORT="year"><FONT>year</FONT> <FONT> [INTEGER]</FONT> NOT NULL</TD></TR></TABLE></FONT>>];
```
sea-kelp: The only difference between the old and new schema file seems to be that `DOUBLE PRECISION` is now `DOUBLE_PRECISION`. I think this might have been caused by an old sqlalchemy package version in my env.  I just rebuilt the diagram locally and also got `DOUBLE_PRECISION`.

Would it be possible to generate the schema and diagram within docker (similar to how we run tests in docker) so we don't have to worry about package/OS differences?

michplunkett: Hmmm, that would create less problems. I'm here for it.

requirements.txt
michplunkett: I think another task would be to add validation that the sphinx files are in place. I may do that later.

dev-requirements.txt
```diff
@@ -1,10 +1,10 @@
...
 mock==5.0.2
 nose>=1.3.1
 pandas>=0.23.4
 pre-commit==3.8.0
+pygraphviz==1.14
```
michplunkett: Done, @sea-kelp.

docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.2"
-
```
michplunkett, line 2:
```bash
time="2025-04-13T04:45:32Z" level=warning msg="/home/runner/work/OpenOversight/OpenOversight/docker-compose.yml: `version` is obsolete"
```

dev-requirements.txt
sea-kelp: Sorry, just noticed that the eralchemy README actually mentions that we need to specify extras:
```
$ pip install eralchemy[graphviz]
```
https://pypi.org/project/eralchemy/

docker-compose.yml
```diff
@@ -106,5 +104,5 @@ services:
      OO_SERVICE_EMAIL: "openoversightchi@lucyparsonslabs.com"
      TIMEZONE: "America/Chicago"
    volumes:
-     - ./OpenOversight/:/usr/src/app/OpenOversight/:z
+     - .:/usr/src/app:Z
```
sea-kelp: It might be easier to mount the database/ directory

database/schema.dot
michplunkett: Done: https://github.com/lucyparsons/OpenOversight/pull/1177

docker-compose.yml
michplunkett: I added the whole directory instead of creating a volume statement for the test directory, database directory, and Makefile The thought process is that it's less fragile and we're not really worried about space on this docker container, especially since it is only for the test environment.

dev-requirements.txt
michplunkett: This should be addressed now.

Makefile
```diff
@@ -36,10 +36,10 @@ db_diagram:
 	# Create new dot file showing current version of schema
 	eralchemy -i postgresql://openoversight:terriblepassword@postgres/openoversight-dev -o database/schema.new.dot
 	# Sort new version of schema file
-	sort database/schema.new.dot -o schema.new.dot.sorted
+	LC_ALL=C sort database/schema.new.dot -o schema.new.dot.sorted
 	# Create old schema file if it does not exist and then sort it
 	touch database/schema.dot
-	sort database/schema.dot -o schema.dot.sorted
+	LC_ALL=C sort database/schema.dot -o schema.dot.sorted
```
michplunkett: Makes sorting deterministic by using a byte-wise sort order.

dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
 coveralls==3.3.1
-eralchemy==1.5.0
+eralchemy2==1.4.1
```
michplunkett, line 2: eralchemy is no longer maintained, so I switched to eralchemy2: https://github.com/maurerle/eralchemy2

requirements.txt
```diff
@@ -58,7 +58,7 @@ requests==2.32.1
...
+sphinx==7.0.1
```
michplunkett, line 61: Make it so the version matches what's in `dev-requirements.txt`.

Makefile
```diff
@@ -34,23 +34,32 @@ create_db: start
 .PHONY: db_diagram
 db_diagram:
 	# Create new dot file showing current version of schema
-	eralchemy -i postgresql://openoversight:terriblepassword@postgres/openoversight-dev -o database/schema.new.dot
+	eralchemy2 -i postgresql://openoversight:terriblepassword@postgres/openoversight-dev -o database/schema.new.md
```
michplunkett, line 37: The markdown files were more readable and provided a more consistent result on my local.

michplunkett:
It could be due to test generally running in SQLlite instead of Postgres. I'll look at it later.

michplunkett:
It works! 😭

b-meson (approved):

b-meson:
This looks great, I'm glad we switch to markdown especially if its more consistent.

PR #1181, opened 2025-05-04
Fix officer face aspect ratio

sea-kelp:
## Description of Changes
*Cherry-picked from https://github.com/OrcaCollective/OpenOversight/pull/548*
Fix officer face aspect ratio on the browse page.

The officer face thumbnail currently inherits the column width (`col-md-4 col-12`) and has a set height of 300. This causes images that aren't this particular size to become warped.

This change adds the `object-fit: cover` CSS property to the `officer-face` class, which preserves the thumbnail aspect ratio.
https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

## Notes for Deployment
None!

Before:

After:

## Checks

 - [x] I have rebased my changes on `main`

 - [x] `just lint` passes

 - [x] `just test` passes

### DIFF
--- a/OpenOversight/app/static/css/openoversight.css
+++ b/OpenOversight/app/static/css/openoversight.css
@@ -455,6 +455,7 @@ tr:hover .row-actions {
     border: none;
     height: 300px;
     margin: auto;
+    object-fit: cover;
 }
 
 .officer-face.officer-profile {


michplunkett (approved):
I saw this on Orca and was hoping you'd bring it over. LGTM!

PR #1182, opened 2025-05-04
Fix officer tagging when department not set

sea-kelp:
## Description of Changes
Fix officer tagging when department no set.

Currently, we default to department id 0 in several urls when department id is not set. This causes issues because there is no department 0.

This change fixes this by defaulting to None instead.

**To Reproduce:**
1. Prerequisite: Have at least 1 image uploaded
2. Visit http://localhost:3000/cop_faces/images/1
3. Enter an officer id and click "Add identified face"
4. This will send a post request to http://localhost:3000/cop_faces/departments/0/images/4 which should fail due to foreign key constraint

## Notes for Deployment
None!

N/A

## Tests and Linting

### DIFF
--- a/OpenOversight/app/templates/cop_face.html
+++ b/OpenOversight/app/templates/cop_face.html
@@ -24,7 +24,7 @@
           </div>
         </div>
         <div class="card p-5">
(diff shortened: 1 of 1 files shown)


michplunkett (approved):
Good catch!

PR #1210, opened 2026-03-03
Remove `pygrep` from `pre-commit`

michplunkett:
## Description of Changes
`pygrep` got collected into `ruff` so I removed the extra import in the config. I also updated all of the imported packages in `pre-commit` as well.

## Tests and Linting

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -27,4 +27,3 @@ dispatcher where…", "Improve our handling of…", etc.
 ## Tests and Linting
 - [ ] This branch is up-to-date with the `develop` branch.
 - [ ] `pytest` passes on my local development environment.
-- [ ] `pre-commit` passes on my local development environment.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,7 +56,7 @@ repos:
         types_or: [python,pyi]
         args:
           - --fix
-          - --select=B,C,E,F,W,I
+          - --select=B,C,E,F,W,I,TID
           - --ignore=C901,E203,E402,E501,W391,E261
       - id: ruff-format
         types_or: [python,pyi]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,7 +57,7 @@ repos:
         args:
           - --fix
           - --select=B,C,E,F,W,I,TID
-          - --ignore=C901,E203,E402,E501,W391,E261
+          - --ignore=B905,C901,E203,E402,E501,W391,E261
       - id: ruff-format
         types_or: [python,pyi]
         args:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
       - id: trailing-whitespace
         exclude: ^database/schema.dot$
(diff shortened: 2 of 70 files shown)


.pre-commit-config.yaml
```diff
@@ -65,21 +50,21 @@ repos:
...
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.9
+    rev: v0.15.4
     hooks:
       - id: ruff
         types_or: [python,pyi]
         args:
           - --fix
-          - --select=B,C,E,F,W
+          - --select=B,C,E,F,W,I,TID
```
michplunkett, line 59: `I` - https://docs.astral.sh/ruff/rules/#isort-i
`TID` - https://docs.astral.sh/ruff/rules/#flake8-tidy-imports-tid

.pre-commit-config.yaml
```diff
@@ -16,9 +16,6 @@ repos:
         exclude: ^database/schema.dot$
       - id: check-symlinks
       - id: mixed-line-ending
-      - id: fix-encoding-pragma
-        args:
-          - --remove
```
michplunkett, line 21:
```console
`fix-encoding-pragma` has been removed -- use `pyupgrade` from https://github.com/asottile/pyupgrade
```

docs/conf.py
```diff
@@ -2,7 +2,6 @@
 
 from recommonmark.parser import CommonMarkParser
 
-
```
michplunkett, line 5: > Surround top-level function and class definitions with two blank lines.
> 
> Method definitions inside a class are surrounded by a single blank line.
> 
> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).

Source: https://peps.python.org/pep-0008/#blank-lines

This does not apply to variables, comments, etc.

.pre-commit-config.yaml
```diff
@@ -65,21 +50,21 @@ repos:
...
+    rev: v0.15.4
     hooks:
       - id: ruff
         types_or: [python,pyi]
         args:
           - --fix
-          - --select=B,C,E,F,W
-          - --ignore=C901,E203,E402,E501,W391,E261
+          - --select=B,C,E,F,W,I,TID
+          - --ignore=B905,C901,E203,E402,E501,W391,E261
```
michplunkett, line 60: Used to prevent a change in `csv_imports.py`.

```console
B905 `zip()` without an explicit `strict=` parameter
   --> OpenOversight/app/csv_imports.py:394:33
    |
392 |                 if license_plate_str:
393 |                     parts = license_plate_str.split("_")
394 |                     data = dict(zip(["number", "state"], parts))
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
395 |                     license_plate, _ = get_or_create_license_plate_from_dict(data)
396 |                     license_plates.append(license_plate)
    |
help: Add explicit value for parameter `strict=`
```

Caused by this rule: https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/#zip-without-explicit-strict-b905

.github/pull_request_template.md
```diff
@@ -27,4 +27,3 @@ dispatcher where…", "Improve our handling of…", etc.
 ## Tests and Linting
 - [ ] This branch is up-to-date with the `develop` branch.
 - [ ] `pytest` passes on my local development environment.
-- [ ] `pre-commit` passes on my local development environment.
```
michplunkett, line 30: We have an Action in place that makes this redundant.

b-meson (approved):
Approved but it looks like there are commits to a bunch of files that probably don't need to be touched (the migration scripts) but that doesn't actually appear to be an issue.

michplunkett:
The changes for the migration scripts are an enforcement of a particular formatting rule that I wasn't able to ignore, unfortunately.

